PR #55, opened 2019-08-24
replaced all occurrences of visable with visible

jeffw387:
I fixed a misspelling in the code base ("visable" => "visible"). I tested that the project compiles, and it seems to work as it did before but admittedly I have done very little testing. 

Is this likely to break compatibility with prior themes or something similar?


lex148:
This is great thank you

PR #56, opened 2019-09-29
Avoid zombie child processes

tuncer:
We have two processes that `Command::new` but do not take care of waiting
for child processes. We can either call wait or let the kernel know that
we don't care by ignoring `SIGCHLD`. If we don't do this, any child
process will turn into a zombie that will consume resources and litter
the process tree.

The solution uses `nix::sys::signal` unconditionally, since leftwm is
an X11 wm, and it's reasonable to assume a suitable `nix::sys`
environment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,3 +33,4 @@ chrono = "0.4.10"
 futures = "0.1.29"
 uuid = { version = "0.8.1", features = ["serde", "v4"] }
 nix = "0.15"
+libc = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,3 +32,4 @@ liquid = "0.19.0"
 chrono = "0.4.10"
 futures = "0.1.29"
 uuid = { version = "0.8.1", features = ["serde", "v4"] }
+nix = "0.15"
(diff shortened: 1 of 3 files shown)


tuncer:
Hmm, this breaks `SoftReload`.

tuncer:
Hmm, `leftwm-worker` is a child of `leftwm`, and `SoftReload`/`HardReload` call `::std::process::exit(0)`.

tuncer:
We could restore the default handler after `Nanny::new().autostart()` has finished, but that can cause a race. Regardless, here's an attempt to do that, which fixes `SoftReload`/`HardReload` but turns background processes spawned from `~/.xinitrc` into zombies. So, this isn't the final fix.

tuncer:
Maybe we need to adjust `Command::new` invocations in `child_process.rs`. What do you think?

tuncer:
Ideally, we would wait on each child we spawn in `autostart()` and `leftwm-worker`, and thus avoid re-configuring the signal handler.

tuncer:
Or we could register our own signal handler where, upon detecting `SIGCHLD`, we invoke `waitpid()`.

PR #59, opened 2019-11-22
Add a GridHorizontal layout, which splits the workspace into an even NxN grid (or as close as it can get)

Wallacoloo:
I find this particularly useful when editing many files. It gives each window in the workspace the same width by arranging them into a number of columns, and then splits each column into so many rows.

Here's what it looks like with 5 windows:
leftwm-5-windows

And with 9 windows:
leftwm-9-windows

I also added some documentation & changed the other Layout structs to use a simpler syntax (no braces). I kept those style changes in their own commit though so that it should be easy to not merge them if you don't like it.

------
Thanks for making this window manager! I started using it just a few hours ago and I'm really loving how intuitive the default bindings are and how easy it is to get started with.


lex148:
this is great thanks

PR #70, opened 2020-02-08
Add Fedora installation instructions

tim77:
Hello. I building and updating from time to time LeftWM in Copr for Fedora. You can mention this in README if you found this useful.


lex148:
great! thanks!

PR #79, opened 2020-03-10
Add current layout to leftwm-state output

TatriX:
It can be useful for example if one (like myself) wants to show current layout in the status bar.


lex148:
This is great! thanks for the PR

PR #83, opened 2020-03-12
Refactor

TatriX:
- Removed unused config.rs file
- Reformatted code with `cargo fmt`
- Minor code adjustments


TatriX:
I'll add more commits so please don't merge it just yet.

src/bin/leftwm-worker.rs
```diff
@@ -24,10 +24,10 @@ fn main() {
     }
 
     let result = panic::catch_unwind(|| {
-        let mut manager = Box::new(Manager::default());
```
TatriX: It there a reason why it was boxed?

TatriX:
> //left it to fast for some logging managers. We need to wait to give the logging
> //manager a second to boot
Could you please explain in more details what does it mean?

src/bin/leftwm-worker.rs
lex148: I think this was needed for the layouts way back in version 0.0.1. You should be able to remove it.

lex148:
> I'll add more commits so please don't merge it just yet. Created PR to get your feedback if any.

This is awesome. There is absolutely a lot of code that can be cleaned up.

lex148:
shoot me a message when you think it is ready to go.

TatriX:
I guess I'm finished with this one. I wanted to make layouts restorable as well, but it's a bit more complicated, so I'll make another PR for that.

TatriX:
Ok, done. This should be mergable.
Layouts now restored on reload! Yay!

lex148:
> > //left it to fast for some logging managers. We need to wait to give the logging
> > //manager a second to boot
> 
> Could you please explain in more detail what does it mean?

This is from a bug when using SDDM. It has been a long time since I looked at this one, but from what I can remember LeftWM boots up and requests to be the window manager before SDDM is able to release the locks needed to register to the xwindow events needed for window managers.

TatriX:
I remove `sleep` and it seems to work ok with sddm on my machine.
And having instance reload instead of 2seconds is so sweet for development. Could you check if it works for you without the delay?

PR #87, opened 2020-03-17
Add maximized layout

TatriX:
Bound to `modkey+m` by default.


kbc8090:
Did this never get implemented?

VuiMuich:
Hmm, no indication to me, why this was closed.

PR #94, opened 2020-05-16
feat(layout) center main

uipoet:
I have a super ultrawide monitor (5120x1440) where the splitting down the center has one looking far left or right at the primary window. By centering the primary window splitting the remaining windows left and right a nice balance is experienced.

### DIFF
--- a/src/layouts/mod.rs
+++ b/src/layouts/mod.rs
@@ -9,6 +9,7 @@ pub enum Layout {
     EvenHorizontal,
     EvenVertical,
     Fibonacci,
+    CenterMain,
 }
 
 impl Default for Layout {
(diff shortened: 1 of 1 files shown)


lex148:
This is really nice. I think its my new favorite layout.

uipoet:
Glad you like it. Is there already a way we can target our favorite layouts with key bindings or does that need implementing?

lex148:
No that would need to be added. LeftWM does have a pipe file to receive external commands. This is how most external bars interact with it. I would want to support it through this mechanism. If you add a ticket I will work in this. or if you would like to take a stab check out:
https://github.com/leftwm/leftwm/blob/master/src/handlers/external_command_handler.rs

uipoet:
Cool. Yes, I'll have a look and PR if I can or punt to an issue if not.

Issue #114, opened 2020-08-03
switch between desktops/tags


Valerioooo:
I have used leftwm for the past week and I love it!
But there is a thing that I think is missing: the function to switch between tags with the arrow keys. I think this is a key feature that could be implemented to make this WM even more amazing.

mautamu:
Howdy, I was wanting something similar and wrote b75a3e4 to implement this. My new config.toml file (relevant lines only) is:
`[[keybind]]
command = "FocusNextTag"
modifier = ["modkey"]
key = "Right"

[[keybind]]
command = "FocusPreviousTag"
modifier = ["modkey"]
key = "Left"`

Is this what you were after as well, Valerioooo?

May I submit a PR to make this a feature for everyone?

lex148:
Absolutely I would Love A PR

Valerioooo:
> Is this what you were after as well, Valerioooo?
YESSSS!
This is very funny, I was already making changes to the code in my version to add support for that.
 Thanks man.

PR #115, opened 2020-08-07
Add FocusNextTag FocusPreviousTag, as per leftwm/leftwm#114

mautamu:
Addresses #114


lex148:
This is great thanks!

PR #119, opened 2020-09-06
New Command: master toggle

PVautour:
This pull request adds a command that sends the current active window to the top of the stack. However, if the window is already on top of the stack, it sends it to second position.

By default, this would be mapped to `mod+enter`.


lex148:
very nice improvement. I'm going to be using this one a lot

PR #129, opened 2020-09-25
new Type Tag

lengyijun:
It's cleaner than String.
This commit has not covered all the replacement. 
If necessary, I will replace all later


lex148:
Wow, great thanks for the help cleaning up the code! I will give it a test today and merge

Issue #144, opened 2020-11-19
Is there a way to resize windows with a keybinding?


cjwyett:
DWM has MOD+L/H to make the master window bigger/smaller. Is there something i could do to replicate this in Left?

P.S. Love the project.

lex148:
The closest that currently exist in LeftWM is resizing/moving the window by holding the mod key and pressing the Left/Right mouse buttons to move/resize windows. This feature has gotten a lot of love in the last release so if you aren't on latest you might want to check it out. 

That being said, if there is enough interest in being able to resize/move without the mouse then lets add it.

cjwyett:
@lex148 Movement is easy enough for now via combinations of Mod+Arrow and Mod+Ret. I would just really like to be able to quickly move X (below) for different tasks.   
```
+---------X-------+
|         |       |
|         |       |
|         |       |
|         +-------+
|         |       |
|         |       |
|         |       |
+---------X-------+
```
As [DWM](https://dwm.suckless.org/dynamic_window_management/) put eloquently: 

>  The number and nature of windows you work with changes all the time, and a dynamic window manager lets you adapt to that and always efficiently use your precious screen real estate... you might want to have a patch someone emailed you in view while you look at some code, or edit a configuration file while you look at a man page or website. Dynamic window management lets you mix and match on the fly to always have the windows you need in view, and only the windows you need, without the need to reserve extra space which you may or may not end up using.

KevinNThomas:
I agree that resizing windows with keybindings would be great, it's a function I use all the time on other WMs.

theguy147:
I just implemented this for myself but it's the first time for me to modify rust code so its probably very bad style and not respecting some rust no-go's. as an inspiration I added a PR but somebody who knows what they are doing should probably redo that if others are interested :)

here is the PR: #158

PR #158, opened 2021-01-09
resize windows with keybinding

theguy147:
this is my implementation to resize the main window using keybindings like in many other WMs as presented by @cjwyett in issue #144 (but only for the `main_and_vert_stack` layout for now)

although it does work as intended for me the PR is only meant as an idea as this is my first time modifying rust code and I probably violated many rust and/or leftwm concepts here and i'm not sure how the testing usually works with rust/leftwm

as a result of this code new keybindings can be used to change the size of the main window (for the one layout it was implemented for). The value represents width change as a percentage:

*[config.toml]*:
```
[[keybind]]
command = "IncreaseMainWidth"
value = "5"
modifier = ["modkey"]
key = "l"

[[keybind]]
command = "DecreaseMainWidth"
value = "5"
modifier = ["modkey"]
key = "h"
```

### DIFF
--- a/src/models/workspace.rs
+++ b/src/models/workspace.rs
@@ -16,6 +16,7 @@ pub struct Workspace {
     pub tags: Vec<Tag>,
     pub avoid: Vec<XYHW>,
     pub xyhw: XYHW,
+    pub main_width_percentage: u8,
     xyhw_avoided: XYHW,
 }
 
@@ -45,6 +46,7 @@ impl Workspace {
             layout: Layout::default(),
             tags: vec![],
             avoid: vec![],
+            main_width_percentage: 50,
 
             xyhw: XYHWBuilder {
                 h: bbox.height,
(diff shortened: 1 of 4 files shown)


lex148:
This is great! thanks for your PR

src/models/workspace.rs
```diff
@@ -45,6 +46,7 @@ impl Workspace {
             layout: Layout::default(),
             tags: vec![],
             avoid: vec![],
+            main_width_percentage: 50,
```
theguy147, line 49: the default width should probably not be hardcoded here

theguy147 (commented):
oh, i didnt expect this to get merged. Since I coded this I noticed e.g. that the width of the main window is not set per workspace as I inteded but somehow globally? at least the setting is persistent across workspaces. Maybe because of my use of `manager.focused_workspace` the setting is always used for the focused workspace whichever it might be? Maybe retrieving the index from the focused_workspace to then access it by index is better? not sure - as i said it was the first time ever that i touched rust code...

also i guess the default width value (of 50 percent) shouldnt be hardcoded in `workspace.rs` but rather be set somewhere were it might be more expected otherwise it will be hard to find if the complexity of the codebasis grows.

Issue #52, opened 2019-06-21
Weird floating mode splash screen behavior


nathanleighsays:
Because floating mode windows are present on all tags, when a program with a splash screen loads, if you change tags while it's loading, it opens on the tag you're currently on, not the tag you initially intended to open it to. It would be nice if there were some way for the information of what tag you initially were on to be retained during the loading process, so you don't need to wait for it to load to continue to use the system in order to have it open where you want it, or have to shuffle around programs after the fact.

hertg:
Is slightly related to #427, but not the same.

@AethanFoot I just realized that I lack the permissions to rename issue titles. I am only able to add labels, assign people, or close them. Could you rename this issue to something like `spawn apps on tag where initially launched`.

Also, I may be adding some `severity` or even `complexity` labels based on my judgement here and there. Feel free to change any of those if you believe I misjudged anything.

PR #165, opened 2021-01-30
Spawn windows on parent terminal's workspace

gmorenz:
I made this change for myself to ease development of apps that spawn windows - this way `cargo watch -x run` puts the window on the screen with the terminal that is building it instead of on the screen with my editor(s).

I'm curious whether or not this (or something similar) is of enough interest to get merged. If not that's perfectly understandable, it's a pretty specific feature.

The way it decides whether or not a window is a terminal right now is just checking against the name of a specific terminal... and probably shouldn't be merged. I suspect it's a problem that can't be solved generally, many popular terminal emulators are actually spawning all their windows from the same process - so I don't think we would be able to determine which window is actually responsible for spawning a process. On the other hand, we could at least make the names of the "terminal" processes configurable, and we might even be able to do something like detect processes with a single window and an open tty...

### DIFF
--- a/src/display_servers/xlib_display_server/event_translate.rs
+++ b/src/display_servers/xlib_display_server/event_translate.rs
@@ -30,7 +30,8 @@ impl<'a> From<XEvent<'a>> for Option<DisplayEvent> {
                     Ok(attr) if attr.override_redirect > 0 => None,
                     Ok(_attr) => {
                         let name = xw.get_window_name(event.window);
-                        let mut w = Window::new(handle, name);
+                        let pid = xw.get_cardinal_prop_value(event.window, xw.atoms.NetWMPid);
+                        let mut w = Window::new(handle, name, pid);
                         let trans = xw.get_transient_for(event.window);
 
                         if let Some(hint) = xw.get_hint_sizing_as_xyhw(event.window) {
--- a/src/display_servers/xlib_display_server/mod.rs
+++ b/src/display_servers/xlib_display_server/mod.rs
@@ -201,7 +201,8 @@ impl XlibDisplayServer {
                 };
                 if managed {
                     let name = self.xw.get_window_name(handle);
-                    let w = Window::new(WindowHandle::XlibHandle(handle), name);
+                    let pid = self.xw.get_cardinal_prop_value(handle, self.xw.atoms.NetWMPid);
+                    let w = Window::new(WindowHandle::XlibHandle(handle), name, pid);
                     all.push(w);
                 }
             }),
--- a/src/display_servers/xlib_display_server/xatom.rs
+++ b/src/display_servers/xlib_display_server/xatom.rs
@@ -48,6 +48,8 @@ pub struct XAtom {
     pub NetWMDesktop: xlib::Atom,
     pub NetWMStrutPartial: xlib::Atom, //net version - Reserve Screen Space
     pub NetWMStrut: xlib::Atom,        //old version
+
+    pub NetWMPid: xlib::Atom,
 }
 
 impl XAtom {
@@ -86,6 +88,7 @@ impl XAtom {
             self.NetWMDesktop,
             self.NetWMStrutPartial,
             self.NetWMStrut,
+            self.NetWMPid,
         ]
     }
 
@@ -182,6 +185,9 @@ impl XAtom {
         if atom == self.NetWMStrut {
             return "_NET_WM_STRUT";
         }
+        if atom == self.NetWMPid {
+            return "_NET_WM_PID"
+        }
         "(UNKNOWN)"
     }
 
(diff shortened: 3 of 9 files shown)


lex148:
This seams like a good Idea. I like little niceties like this, they really make the everyday use a lot more enjoyable. I will give it a good testing today. 

I agree with you just looking at the name seams less than ideal. If that is the best we can do, so be it, but lets at least look for alternatives.  

Awesome work, Thanks for the PR!

gmorenz:
So - poking around - it seems like we can reliably discover if a process has spawned a tty by looking running `rg "tty-index" /proc/{pid}/fdinfo`... which seems fairly natural since I'm poking around the proc filesystem anyways (seems to be the only reasonable way to get parent pids). Surprisingly to me `fdinfo` doesn't show bash as having a tty for it's std<in/out/err>, but ok.

I'm thinking the (pseudo-code) logic here should be

```rust
fn find_parent_terminal(pid: Option<u32>) -> Option<Window> {
    let mut pid = pid?;

    // We only care if this is the first window the process spawns
    if has_a_window(pid) {
        return None;
    }

    while let Some(parent) = get_parent(pid) {        
        pid = parent;
        if let Some(window) = find_window(pid) {
            // If it has multiple windows we don't even try
            if find_another_window(pid) {
                return None;
            }

            if is_terminal(pid) /* has a /proc/{pid}/fdinfo/... file containing "tty-index: *" */ {
                return Some(window)
            }

            return None;
       }
    }

    None
}
```

gmorenz:
Also, having used this a bit more sunday, I think I want to make focus only jump to a window if the parent window has focus (or it has no parent window). I'm not sure if that is really the right call for all use cases though.

lex148:
@gmorenz just wanted to check in on you here. What are your thoughts on this PR? are you still planning on playing with it a bit more? You have it marked as `Draft` so i didn't want to override that without discussing with you first. I certainly don't want to leave your fine work waiting on me.

gmorenz:
Hey, I'll come back to this this weekend... probably implement the algorithm suggested above and mark this as not-a-draft. I originally wanted to give you a chance to look at the proposals above (and I'm taking your comment as general agreement with them). This hasn't been high priority since I only use this for recreational programming, hence the time lag.

On the other hand if you want to do something with this sooner feel free, I marked this as a draft because I didn't think this behavior change should be blindly merged (both the behavior in general and the hardcoded terminal part) not because I didn't want you to use the code.

mautamu:
Any updates on this great project? 

I think we could expand the scope (probably in another PR, especially if you don't have much time to work on it, gmorenz). The PR currently has most of the parts necessary to implement #52 and #251. If we add an implement for /proc/[PROC]/exe or maybe /proc/[PROC]/cmdline to get the executable path, a specification in the config.toml file might set a window to float or to be tagged specifically:
```toml
[[window]]
program = "/usr/lib/firefox-developer-edition/firefox"
float = "true" #alternatively, type = "Splash"?
tag = "bravo" # or should we do tag index to speed this up?
```

VuiMuich:
> Any updates on this great project?
> 
> I think we could expand the scope (probably in another PR, especially if you don't have much time to work on it, gmorenz). The PR currently has most of the parts necessary to implement #52 and #251. If we add an implement for /proc/[PROC]/exe or maybe /proc/[PROC]/cmdline to get the executable path, a specification in the config.toml file might set a window to float or to be tagged specifically:
> 
> ```toml
> [[window]]
> program = "/usr/lib/firefox-developer-edition/firefox"
> float = "true" #alternatively, type = "Splash"?
> tag = "bravo" # or should we do tag index to speed this up?
> ```

This would be so great!
Just throwing some random ideas along this:
- we could try to get the window class (like picom does and iirc i3 does this as well) we could use let the user decide to use this or `program`
- we could allow further parameters to be tied to this window(s) like layout, margins, border(-color) etc.
- instead of `float: bool` we could use `force_type: WindowType`
- I think handling tags with the id string is ok and correct in the config. Internally I'd prefer to only use the tag index and use the id srting for i/o. I already started a local branch for this endeavour, but I guess this will be a big one and should go along with the move to tag-centric (which btw. should probably become an actual feature branch eventually)
- tbh I haven't read too much into `toml` specs, does it support variables? Thess config options feature would probably strongly benefit from variables i believe.
- with more and more 'small' config options which invite to tinkering and optimizing configs it would be really nice to make the `soft reload` actually soft aka a 'reload config' that doesn't need to restart leftwm-worker (probably replacing `SoftReload` and `HardReload` commands with `Reload` that takes as value `soft`, `hard` and `config`)


I'll probably pull some of these ideas to issues later, but wanted to dump my thoughts just quickly.


PS: @lex148 it would be really great if you could open the discussion section so we could have these ideas/feature-wishes/random thoughts in a central place and dont't clutter PRs and issues.

VuiMuich:
Just a friendly bump an ping for @gmorenz
Getting some motion back into this great PR would be really awesome!

VuiMuich:
@gmorenz Are you still interested in making this merge ready?
We would really like to get this feature into master.
But if don't want to return to this PR it is ok as well, just let us know and someone else will take this.

PR #181, opened 2021-02-10
main & horizontal stack layout

abhikjain360:
Adds a new layout, similar to the "bstack" layout in this [dwm patch](https://dwm.suckless.org/patches/bottomstack/).

The code is more or less a complete copy to `src/layouts/main_and_vert_stack.rs`, except a few changes here and there, so should work fine.

I have tested on my laptop.


lex148:
This is great! Thanks for the PR. I will give it a test and then merge.

PR #190, opened 2021-02-13
Added "center main balanced" layout

ureeves:
I've added a new layout for personal use and thought I would see if there's some interest in merging it.

The layout separates the screen into three columns, leaving a column in the middle much like the "center main" layout, and distributing the windows through the other columns in a manner similar to the fibonacci layout but "row first".

I've found it particularly useful on my ultra-wide.


lex148:
wow this is great thanks. Sorry it took longer than normal to look at this, I was out of town.

ureeves:
Happy to contribute! =D

Issue #194, opened 2021-02-16
Allow user greater control of margin


AethanFoot:
The issue I have is that the gap between windows is double that around the edge. So I feel giving the user more over the margins and padding, I think bspwm handles this quite well (but maybe a but over what is needed).
For example if margin is set to 8 10, 8 will be the padding for the top and bottom and 10 will be the left and right. 8 10 8 will set top (left right) bottom. And 8 10 10 8 will set top left right bottom. For the result I would want a new variable may need to be added which works differently to margin which defines the gap between windows rather than the padding around every window.

PR #196, opened 2021-02-17
Add more specific margin capabilities

mautamu:
Good evening,

This PR adds functionality to the margins as per #194. I went with the HTML format of [top, right, bottom, left] and [top/bottom, left/right]. To change the margins, edit _~/.config/leftwm/themes/current/theme.toml_ from:
```toml
border_width = 4
margin = 10
```
to:
```toml
border_width = 4
margin = [10, 10]
```

I intended to apply this to borders as well, but it does not appear that this is possible due to an upstream requirement of a singular integer to define borders, but it may be possible.

### DIFF
--- a/src/models/window.rs
+++ b/src/models/window.rs
@@ -4,6 +4,7 @@ use crate::config::ThemeSetting;
 use crate::models::TagId;
 use crate::models::XYHWBuilder;
 use crate::models::XYHW;
+use crate::models::Margins;
 use serde::{Deserialize, Serialize};
 use x11_dl::xlib;
 
@@ -28,7 +29,7 @@ pub struct Window {
     pub type_: WindowType,
     pub tags: Vec<TagId>,
     pub border: i32,
-    pub margin: i32,
+    pub margin: Margins,
     states: Vec<WindowState>,
     pub normal: XYHW,
     pub start_loc: Option<XYHW>,
@@ -48,7 +49,7 @@ impl Window {
             type_: WindowType::Normal,
             tags: Vec::new(),
             border: 1,
-            margin: 10,
+            margin: Margins::Int(10),
             states: vec![],
             normal: XYHWBuilder::default().into(),
             floating: None,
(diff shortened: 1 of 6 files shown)


src/models/window.rs
```diff
@@ -147,9 +148,13 @@ impl Window {
             value = self.normal.w();
         } else if self.floating() && self.floating.is_some() {
             let relative = self.normal + self.floating.unwrap();
-            value = relative.w() - (self.margin * 2) - (self.border * 2);
+            value = relative.w()
+                - (self.margin.clone().left() + self.margin.clone().right())
```
VuiMuich, line 152: @mautamu or @lex148: may I ask, why you apply a margin to a floating window?

mautamu: I’m not sure exactly why it’s applied. Can try running it without and see what happens.

VuiMuich: Might test myself as well. Was just curious if there was some reasoning behind that.

src/models/window.rs
```diff
@@ -205,9 +214,9 @@ impl Window {
         }
         if self.floating() && self.floating.is_some() {
             let relative = self.normal + self.floating.unwrap();
-            relative.y() + self.margin
+            relative.y() + self.margin.clone().bottom()
         } else {
-            self.normal.y() + self.margin
+            self.normal.y() + self.margin.clone().bottom()
```
VuiMuich, line 219: another question if I may: why are you using `.bottom()` here and no `.top()`? Isn't the origin for determining window-positions the top-left screen corner?

mautamu: Now that you mention it, that’s probably right. This might be an artifact from when I was confused and had the HTML top/bottom/ left/right specs backwards.  

Should probably check over and see if the behaviour is what we’re expecting it to be.

VuiMuich: as I am just about to get a `MarginMultiplier` PR ready (just struggling to get the external command working) so I might include this..

mautamu: That sounds great! Thanks for reviewing that. I wonder if it's partially related to #248

VuiMuich: > I wonder if it's partially related to #248

Gonna look into this tomorrow, now it's bedtime. Gn8

Issue #174, opened 2021-02-07
Maximize Window with keybinding


blackCauldron7:
I switched from awesome wm it had a feature to maximize the current focused window with a keybinding. Is it possible to do this in leftwm?

lex148:
Absolutely! We have had a bit of discussion about adding this but haven't decided on exactly what we want yet. 

The two ideas that have been floating around are:
1) a full screen mode like how you can resize windows when holding mod and dragging. 
you would just toggle it on/off per window.
2) add a full screen layout. the layout would only show the main window and it would fill the screen.

What do you think?

blackCauldron7:
I already implemented the second idea you are talking about with the command line utility wmctrl and the keybindings in the config toml.

And i cant understand what you mean by the 1st

And i meant by maximize not fullscreen rather i meant the window will take all the available space on the workspace other than the bar.

And lastly thanks for this amazing application totally love it.I just love rust rewrites.

VuiMuich:
I think option 1 means something like toggle fullscreen when you drag a window to the top-edge of a screen. (Similiar to what Windows7/8/10 does)

@blackCauldron7 would you mind to share your keybind snippet?

blackCauldron7:
@VuiMuich 

Personally i never use a mouse in a tiling wm so i wouldnt want that feature but i think it would have helped me if i was transitioning to a tiling wm and this was my first so i think that would be cool.

### Snippet

```
[[keybind]]
command = "Execute"
value = "wmctrl -r :ACTIVE: -b toggle,fullscreen"
modifier = ["modkey"]
key = "f"
```

VuiMuich:
I agree once one transitioned to tilinWM using a mouse to arrange windows feels awkward :D

Thansk for the share, works great. Obviously this should remein a workaround, since it comes with a variety of little issues, link when putting one of many winows in fullscreen the layout of the remaining windows changes and the are visible through transparent windows.

Haha, when I stumbled on the behaivour of #170 I was not sure if this was an intended feature or a bug.

blackCauldron7:
i agree that keybinding is a hacky workaround that i dont like too much imo there should be something like this:

```
[[keybind]]
command = "ToggleFullscreen"
modifier = ["modkey"]
key = "f"
```

VuiMuich:
also there shoul probably an option if the fullscreen should respect the bar(s) or be truly full screen

blackCauldron7:
@VuiMuich i think there should be two different options like in other wms like

[...]

PR #197, opened 2021-02-17
monocle Layout

VuiMuich:
As proposed in #174 here is my attempt to add a simple Monocle Layout, which sets the primary window to full screen and hides any others. The windows can be cycled through with MoveWindow commands.

In a hacky way it get rid of the margin, but I am not sure how to handle properly to return to the theme margin again in other layouts, yet.

Also it would be possibly nice to have a monocle specific margin, like doule or tripple the theme margin. Maybe have a "monocle-factor" value in the theme.toml?

### DIFF
--- a/src/layouts/main_and_horizontal_stack.rs
+++ b/src/layouts/main_and_horizontal_stack.rs
@@ -1,0 +1,39 @@
+use crate::models::Window;
+use crate::models::Workspace;
+
+/// Layout which splits the workspace into two columns, gives one window all of the left column,
+/// and divides the right column among all the other windows.
+pub fn update(workspace: &Workspace, windows: &mut Vec<&mut &mut Window>) {
+    let window_count = windows.len();
+    if window_count == 0 {
+        return;
+    }
+
+    let height = match window_count {
+        1 => workspace.height() as i32,
+        _ => (workspace.height() as f32 / 100.0 * workspace.main_width()).floor() as i32,
+    };
+
+    //build build the main window.
+    let mut iter = windows.iter_mut();
+    {
+        if let Some(first) = iter.next() {
+            first.set_width(workspace.width());
+            first.set_height(height);
+            first.set_x(workspace.x());
+            first.set_y(workspace.y());
+        }
+    }
+
+    //stack all the others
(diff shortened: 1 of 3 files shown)


AethanFoot:
Not sure if anyone agrees but I feel this layout should respect margins, with a separate feature that allows the user to completely fullscreen any application (not respect margins and hide the bar). Like you can with a browser with pressing F11 and keep the current layout so when you leave fullscreen everything is the same.

AethanFoot:
I have all found this issue, not sure mechanism is not working nicely, but any window added past the initial window covers my monitor starting at +0+0 (not matter which screen im on).
Video:
https://user-images.githubusercontent.com/63980259/108344507-1f2dcc80-71d5-11eb-90fb-a9683981c383.mp4

VuiMuich:
> I have all found this issue, not sure mechanism is not working nicely, but any window added past the initial window covers my monitor starting at +0+0 (not matter which screen im on).

Thanks for the catch. I think I hav an idea, why this is happening. The layout so far does not handle how new windows are created.
New windows are usually just appended to the existing ones but whte monocle layout should put new windows as primary. 
Also (and maybed related) focus switching is kind of inconistent.

Maybe it would be better to make this layout more dependent on the focus state of windows anyway. I try if I can make it work this way.

I agree that a proper fullscreen mode for windows is kind of a must have and should be independent. Though on top of that I think the monocle should have a margin setting that stands out of the usual layouts margins, thus my suggestion with the margin-factor in theme.toml

VuiMuich:
The last commit is somewhat working.
As mentioned I am looking for a way to get the layout dependent on the focus-state, but currently I'm somewhat struggeling to get the focused window returned to my update function.

The Idea is to iterate over all windows of the workspace, and set the size, then match the iter-item with the window that has focus to set its visibility true, all other windows to false.

I am uncertain though if this way the layout update get called automatically if focus changes, or if dedicated action is needed.

Maybe someone with better rust-fu could help me out (esp. on how to properly get the current focuesd window.)

VuiMuich:
with mautamu's helpful PR on my fork I think it is ready to leaf the WIP state for now.

It's still not perfect, but ready for general use I'd say.
How the margins are handled should be in a dedicated PR anyways, imo

Issue #209, opened 2021-02-23
Window focus doesn’t change in some situations


VuiMuich:
When the pointer is above an area that is `never_Focus` the focus is not updated correctly.

In case of switching tags it stays on the focused window of the previous focused tag.

(In case of #197 it creates the new window in the background)

AethanFoot:
I also get this weird behaviour as well, it is almost like the focus is sticky sometimes. If I remember correctly the border will change colour as if the focus has changed but the focus in the app doesn't update, however I'll try and do more testing on it to see if I can find a way to reproduce it.

AethanFoot:
I have noticed when in the monocle layout that when cycling through focus only changes between the top two windows. Also when switching focus the mouse always goes to the far left of my left monitor, instead of the middle of the focused window (this then prevents switching on the main monitor as the cursor moves to the other screen).

lex148:
I see this as well. I am testing a fix right now. should have a PR up later today.

lex148:
This turned out not to be one problem but several little focus oddities. I have found about four today while testing. I have merged the fixes for them. If we continue to see focus bugs let me know and we will keep killing them.

VuiMuich:
absolutely love the pace this project is moving right now :smiley:

AethanFoot:
I am still having focus issues in the monocle layout. The biggest issue I am having is when focus changes when a workspace in this layout, the mouse cursor moves to the far left of my monitor at +0+0 (taking focus with it). Examples of when this happens is when cycling focus, and adding new windows. This may be a bug within the code mautamu added. If it is helpful I can record a clip to better show what happens.

mautamu:
It most likely is. Probably need to add the function that centers the mouse on a window to the lines I put in.

mautamu:
@AethanFoot try removing line 266:
```rust
let act = DisplayAction::MoveMouseOver(handle);```

AethanFoot:
ok, so that helps when cycling through the windows (obviously as moving the mouse is no longer called). However there must be an issue with the handle when on monocle as the issue happens when a new window is spawned which shifts the focus and the mouse to the left of the monitor at +0+0 (but the new window is spawned hidden). 

Aside from the mouse, when cycling through the windows (in monocle) it will cycle going up properly, but will only switch between the original window and the one below when going down.

AethanFoot:
[...]

PR #224, opened 2021-03-01
Temp fix for movement/focus changes in monocle layout

AethanFoot:
Allows for proper window focusing in monocle layout as per #209. When focusing down you can only change between the top two windows but this is a deeper issue in how the vector is reorder.


lex148:
@AethanFoot fine work! I just tested your PR and it looks like it is working much better. You are correct about the window vector. I have been working on a PR that reworks it. In general I'm hopeful focus should be much better very soon. I am considering merging your PR into my branch instead of master just to avoid some of the conflicts I know are going to occur. It looks like we are working right on top of each other.

PR #234, opened 2021-03-05
Update checker and bump liquid to 0.22

mautamu:
Good evening,

This commit updates the checker. Output now appears like as follows:
```bash
~> leftwm-check
Note: Using file /home/mautamu/.config/leftwm/config.toml.bad
    -> Configuration loaded OK
:: Checking keybinds . . .
ERROR: This keybind requres a `string` for keybind Keybind { command: Execute, value: None, modifier: ["modkey", "Shift"], key: "f" }
```
Best,
Mautamu


lex148:
nice work as always @mautamu

PR #235, opened 2021-03-07
Adds left_wider_right_stack and right_wider_left_stack

irmantas:
Made these layouts for personal use 

Adds two new layouts:

* left_wider_right_stack - splits the workspace into two columns. Gives first column 2/3 workspace width on the left side, 1/3 for second column on the right side.
* right_wider_left_stack - splits the workspace into two columns. Gives first column 2/3 workspace width on the right side, 1/3 for second column on left side.


lex148:
@irmantas these are great! I would like to update these two to support the two commands (IncreaseMainWidth & DecreaseMainWidth ). That way they can be resized. Do you want to take this on as part of this Pull Request? If not that is fine. 

If you feel up to it, here is an [example](https://github.com/leftwm/leftwm/blob/master/src/layouts/main_and_vert_stack.rs) layout that supports the two commands already. 

Again thanks for your PR.

irmantas:
@lex148 I would like to try (just started to learn Rust, love it so far). I would update PR in couple of days.

irmantas:
It looks like I cannot get to work (IncreaseMainWidth & DecreaseMainWidth) commands with existing layouts on current master branch. Tested with and without workspaces config. Checked every single layout, but without luck. Any ideas what I am doing wrong?

Here is config https://gist.github.com/irmantas/3034a4ba74e16da5088fe2eecbbd8739

mautamu:
Both require a value (% of screen) to be specified:
```toml
[[keybind]]
command = "IncreaseMainWidth"
value = "5"
key = "h"

[[keybind]]
command = "DecreaseMainWidth"
value = "5"
key = "j"

```

irmantas:
PR updated.
Please can you check if I am going to right direction with this one :)

mautamu:
After a bit of cursory testing, this looks awesome! Thanks mate!

mautamu (approved):

lex148:
@irmantas my apologies, I have been out this last week. fine work!

PR #246, opened 2021-03-18
Added information section on floating windows.  Reformatted Default K…

rebuilt:
…eys section for readability

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,21 +158,32 @@ The settings file to change key bindings and the default mod key can be found at
 ## Default keys
 
 ```bash
-Mod + (1-9) => Switch to a desktop/tag
+Mod + (1-9)         => Switch to a desktop/tag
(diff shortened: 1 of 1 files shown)


VuiMuich:
Just a little nitpick: you could indent the arrows on the keybinds to the same column as the in the floating section for consistent alignment.

rebuilt:
The arrow symbols take up more than a single character of spacing.  This resulted in uneven spacing despite having the same number of spaces in the markdown file.  Reducing the number of spaces got it closer to correct alignment, but you're right, it still looks messy.  I think turning it into a table might fix the alignment.  I'll try that.

VuiMuich:
That's great with the tables! This results in even more mobil friendly display.

mautamu:
Appreciate this! Will cross reference and update mautamu/leftwm_wiki for merging hopefully later tonight.

rebuilt:
Thanks!   This will be my first real pull request.

PR #243, opened 2021-03-17
Add SetLayout Command

nickzana:
As discussed in #174.

PR #253, opened 2021-03-22
Add SetLayout as external command

nickzana:
As requested [here](https://github.com/leftwm/leftwm/pull/243#issuecomment-804003368).

I haven't tested that this super thoroughly, and just went based on grepping the codebase for instances of other external commands, so I've marked it as a draft in case somebody else who has better knowledge of how `ExternalCommand`s work can test and verify the changes.

If something is wrong, just let me know and I'll be happy to fix it.


lex148:
Works great thanks again!

PR #254, opened 2021-03-24
Add separate mouse modifier key

### LINKED ISSUE #215: Allow separate modifier key to be set for mouse events
Allow the user to set a separate modifier key when moving a window with the mouse (and for in the future if resizing a window with the mouse gets added). The reason I suggest this is so I can use I can use a key on my mouse as a modifier (I have one set for ctrl for crouching in games), so you could just use one hand.

AethanFoot:
Adds a separate modifier key for mouse actions to config (default super). I am not 100%, but does this make the mod_key_mask variable in the display server redundant? This closes #215.


lex148:
awesome work!

PR #257, opened 2021-03-27
Fixed Typo `LeftWiderRi(g)htStack` and fix for `SetLayout` not working properly with `LeftWiderRightStack` and `RightWiderLeftStack`

### LINKED ISSUE #256: Why is this layout named LeftWiderRihtStack?
Not a huge deal, but I don't understand why what seems to be a typo to me persists throughout all the code. Shouldn't it be named LeftWiderRightStack? It's quite bothersome and seems like it would just lead to confusing typos.

VuiMuich:
This closes #256

### DIFF
--- a/src/layouts/mod.rs
+++ b/src/layouts/mod.rs
@@ -26,7 +26,7 @@ pub enum Layout {
     CenterMainBalanced,
     Monocle,
     RightWiderLeftStack,
-    LeftWiderRihtStack,
+    LeftWiderRightStack,
 }
 
 pub const LAYOUTS: [Layout; 11] = [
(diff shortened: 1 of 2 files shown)


VuiMuich:
hmm, for some reason `LeftWiderRightStack` and `RightWiderLeftStack` only get updated to, when cycling through layouts, but not with the (iternal as external) `SetLayout` command.

VuiMuich:
this is probably, because these two Layouts set the main_width with `pub fn main_width(&self) -> u8` in side `layouts/mod.rs`
gonna try to fix this. for this PR

PR #258, opened 2021-03-28
Added FloatingToTile command

AethanFoot:
This will snap focused floating to the layout with a command. From a minor bit of testing it seems to work. Let me know anything you think should be changed (including the name, seems abstract but didn't want it to be too long).


VuiMuich:
Wow this looks great!

If you changed `command_handler.rs` Line 194 so it changes `manager.focused-window()` (from the top of my head, but sure if this is the correct function) to floating it could be a `FloatingToggle` command.

AethanFoot:
How about FloatingToTile, as it currently is one directional.

lex148:
Very nice I like a lot! 
@AethanFoot would you mind adding a little quick snip-it of info to the wiki so others will know about this new command. You should have access now. if you don't let me know.

AethanFoot:
@lex148 or @mautamu how would you go about registering a double click within mouse_combo_handler, as I feel it would be natural to have a default binding of mousekey + double Button1 to call the command FloatingToTile.

Also what are peoples thoughts on changing a few things in should_snap: These would be instead of a static number for dist, make it dependent on the size of the workspace (which I feel would help for high dpi monitors). And only checking whether the is close to the top to reduce random accidental snapping when moving the window about (also with the current system whenever I move a window in the stack of my left monitor it immediately snaps to the right monitor)

PR #259, opened 2021-03-29
Add commands (keybind and external) `SetMarginMultiplier`

VuiMuich:
Adds the Command `SetMarginMultiplier` to keybinds and external command pipe

Use this to configure a keybind in config.toml:
```toml
[[keybind]]
command = "SetMarginMultiplier"
# value takes float as multiplier
# use "0.0" for no margins at all
# use "1.0" to reset
value = "2.5"
moddifier = ["modkey"]
key = "m"
```

Use this to send it via comman pip2:
```shell
echo "SetMarginMultiplier 2.5" > $XDG_RUNTIME_DIR/leftwm/command.pipe
```

The margin state of a tag should probably get stored in the /tmp/leftwm.state file on soft-reload but I think this is topic of a bigger commit, since window-order seems to be inconsistent during reloads as well.

I'd like to have some opinions, if it should be also possible to apply this multiplier not to a full workspace as it is currently doing, but also to a single window.
And also if you guys think some Increase-/DecreaseMarginMultiplier would come handy, or are too much of doubling similiar functionallity.

Further possible use of the `window::Window::set_margin_multiplier(f32)` might be to give maximized windows the option to loose the margin, as long as they are maximized.

PR for Wiki will follow after review and merge.


VuiMuich:
Oh and btw. I snuck in two little changes as discussed with mautamu in #196

AethanFoot:
Looks good, I feel the increase/decrease is less useful than the multiplier, as I feel once you have set the margins you don't change them about too much unless you want to go into zen mode and remove, or reddit mode and make them unnecessarily big.
This has also brought up another thought of should leftwm-check test the variable type? Also should the user be able to enter a negative without a warning (currently negated with a .abs)?

mautamu:
I think it would be helpful for leftwm-check to be able to test variable type as well. I’ve been pondering how best to verify the values of other commands as well, such as seeing if exec would actually complete. Haven’t come up with an eloquent solution yet unfortunately.

Personally I think that a warn!() might be fine if it’s negative, although I’m curious what others think of it. Could maybe make it a debug!() or even a trace!().

AethanFoot:
I suppose as it is a debugger, it doesn't need to be too graceful. As we have a limited number of types, we could set traits which would return whether the variable of it that type, or try to parse it into the variable and check for errors. However I am used to languages where this is more relevant, so these might not be appropriate for "proper" use.

lex148:
@VuiMuich looking good!. let me know when you get the new window fix and I will merge

VuiMuich:
Thanks @AethanFoot for adding multiplier to state.

I have a new approach in 1444b1e, where I try to use workspace.update_windows(), the benefit is this gets applied to newly spaned windows automatically, the downside is that I need to find a way to filter for windows of the focused tag/workspace (and I want if possible a nicer way then what I copy-paste-hacked in the initial commit). Right now it gets applied to all_windows.

AethanFoot:
I think there is an example of getting the windows of the focused workspace in the MoveWindowUp, which create a test function which tests the tags and then uses the vec extract helper to get the relevent windows.

VuiMuich:
This was where I took my snippet in the first place.
What I kind of dislike with this helper, that it rips the windows from the Manager and you have to append them after manipulating, which feels ok, when you actually work on their positions in the vector itself. But for simply handing them a value and triggering an update this seems to 'aggressive' IMO, so I started to look into a smoother solution.

VuiMuich:
Ok, applying the `margin_multiplier` in `workspaces::Workspace.update_windows(Vec<&Window>)` appears to be "ein Holzweg" ('barking up the wrong tree' just dosn't translate well..)

And it looks like I need to setup a proper debugging, to find, why my new Idea to call it from `window_handlers::created(&mut Manager, Window)` is not doing what I expected. But tbh. I have no idea, how hook in the debugger on a reload.

mautamu:
Do you have systemd @VuiMuich? If so just compile with flag —features=journald. Stuff gets dumped to journalctl (journalctl -xe).

There’s also the file logging (see src/bin/leftwm-worker) but I haven’t really used it so can’t say much about it.

VuiMuich:
> Do you have systemd @VuiMuich? If so just compile with flag —features=journald. Stuff gets dumped to journalctl (journalctl -xe).
> 
> There’s also the file logging (see src/bin/leftwm-worker) but I haven’t really used it so can’t say much about it.

Unfortunatelly both are empty (at least in regard of leftwm).
So I was more hoping for breakpoints and inspecting variables.
But I guess I need to learn how to drop information to the log then.

AethanFoot:
To get info in journal log add a line like
```
log::info!("Some text to identify: {:?}" value_you_want_to_check);
```

VuiMuich:
Ok, I feel stupid now :facepalm: it took me way too long to realise that simple reloads er not enough to get journal-logs started...

Basically margins get applied now to new windows as well, but since the margin is related to the workspace, this feels very much like like misbehaving.
So I guess margin multipliers really should live in tags.

lex148:
I agree this should live with the tag.

VuiMuich:
I just found a workaround.
It its not pretty, but practical: for new windows or `move-to-tag` operations it will copy the margin multiplier from any existing window or will default to '1.0'.

With this latest commit it should be fully functional.
I will investigate how to properly store it in the tag, though I think I tried this and had problems to get the Arc<TagModel> type mutable or something.

@lex148 If you think the workaround is ok for merge I'd like to do one cleanup commit, then it's ready to merge.

lex148:
@VuiMuich I have been testing your PR this morning and am having some issues. I added the keybind

```
[[keybind]]
command = "SetMarginMultiplier"
value = "2.5"
modifier = ["modkey"]
key = "m"
```

When I press mod+m the worker process is crashing. Can you please check this on your end. It very well could be me doing something dumb :(  

Great work by the way. Can't wait to see this one in actions :)

VuiMuich:
@lex148 sure, gonna check later this evening.

Have you tried other values?
Maybe one of the latest refactor commits breaks one of my workarounds. 
Have you any specifics on the crashing? I had some situations (especially with 'new window' and 'move' and 'snap', when the target was empty) I got a crash with a log like this: `leftwm-worker crashed with error: Any.`

This gives me the idea that I haven't tested what happens if one tries to apply the multiplier to an empty workspace/tag. I have  a feeling this is what breaks it, when it cannot find any focused window.

VuiMuich:
@lex148 can you reproduce the crash still after the fixed merge-conflicts?

I just tried here and everything works ok on workspaces with windows with this keybind:
```toml
[[keybind]]
command = "SetMarginMultiplier"
value = "2.5"
key = "m"
```
But as expected I can produce a crash, when I try to set the multiplier on a tag that previously had a window.
A 'virgin workspace' doesn't crash though.
And the error in journalctl is `leftwm-worker [<pid>]: Completed with error: Any` btw.
I'll look into a fix right away..

VuiMuich:
It was this completely unnecessary two lines, that caused the crash:
```rust
let act = DisplayAction::MoveMouseOver(manager.focused_window().unwrap().handle);
manager.actions.push_back(act);
```
I think I put them there, together with reappending the extracted `Vec<Window>` to the manager and forgot them..

VuiMuich:
Oops, I think I still have some cases where I can crash the leftwm-worker

VuiMuich:
Hmm, after fixing clippy lints I couldn't crash it any more.
Could someone else try?
Steps to reproduce:
 - go to empty tag
 - create window
 - destroy window
 - execute `SetMarginMultiplier` -> sometimes crashed here directly
 - create window -> sometimes crashed here

VuiMuich:
So far i wasn't able to produce any further issues with this PR, has anybody else been observing something?

From my side I'd call it ready for merge now.

AethanFoot:
Just did a small bit of testing and it seems to be ok (been upgrading my pc so haven't been able to do a longer testing session, sorry). Great work.

PR #262, opened 2021-04-01
Update to focus system to watch for missed EnterEvents

lex148:
Every so often xlib fails to send us Window EnterNotify events. This means we don't get notified that the cursor has moved over a window. That means focus is left in the previous window not the new window we just moused over. This Update its to address this issue.

### DIFF
--- a/src/bin/leftwm-worker.rs
+++ b/src/bin/leftwm-worker.rs
@@ -64,6 +64,11 @@ where
     xdg::BaseDirectories::with_prefix("leftwm")?.place_runtime_file(path)
 }
 
+async fn timeout(mills: u64) {
+    use tokio::time::{sleep, Duration};
+    sleep(Duration::from_millis(mills)).await;
+}
+
 async fn event_loop(
     manager: &mut Manager,
     display_server: &mut XlibDisplayServer,
@@ -92,6 +97,13 @@ async fn event_loop(
         tokio::select! {
             _ = display_server.wait_readable(), if event_buffer.is_empty() => {
                 event_buffer.append(&mut get_events(display_server));
+                continue;
+            }
+            //Once in a blue moon we miss the focus event,
+            //This is to double check that we know which window is currently focused
+            _ = timeout(100), if event_buffer.is_empty() => {
+                let mut focus_event = display_server.verify_focused_window();
+                event_buffer.append(&mut focus_event);
                 continue;
             }
             Some(cmd) = command_pipe.read_command(), if event_buffer.is_empty() => {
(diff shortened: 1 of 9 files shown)


lex148:
Hey everyone I have been working to cleanup and fix the focus logic all week. I think It is significantly better, but would really appreciate help testing / reviewing . This just touches so many things, I would really like a couple eyes on this one. Let me know what you think

src/bin/leftwm-worker.rs
```diff
@@ -91,6 +96,13 @@ async fn event_loop(
                 event_buffer.append(&mut get_events(display_server));
                 continue;
             }
+            //Once in a blue moon we miss the focus event,
+            //This is to double check that we know which window is currently focused
+            _ = timeout(500), if event_buffer.is_empty() => {
```
mautamu, line 101: There does appear to be a bit of delay in focusing the new window: (sorry it's blurry, forgot to turn the compositor off!)
Peek 2021-03-31 20-20

I wonder if 500ms is the right amount of time? I know I probably wouldn't notice, but might miss a keystroke or something.

mautamu (approved):
This is great! Thank you for fixing this!
One of the common ways I produced the issue in the past appears to be resolved:

- Start typing in FireFox
- Switch to a different tab 
- Launch new program
- Return to old tab prior to new program launching
- Move mouse over FF after new program launches
- Stuck typing on old window

^ That route appears to be fixed in this patch.

The one thing is that slight delay, not sure how we could make that smoother.

AethanFoot:
Immediately I can tell this is better, with just switching between terminals and dialogs. Will test this for the rest of the day, Awesome Work!

src/bin/leftwm-worker.rs
AethanFoot: Yeah I feel 500ms is too high for an interactive system, for my testing I have lowered this to 250ms to see if it still works completely and will lower it as far as it can on my setup (ideally 100 or below)

AethanFoot: Would something like this work to find out the optimum value
```
            _ = timeout(25), if event_buffer.is_empty() => {
                for n in 0..50 {
                    let mut focus_event = display_server.verify_focused_window();
                    if !focus_event.is_empty() {
                        event_buffer.append(&mut focus_event);
                        if n > 0 {
                            log::info!("Leftwm n: {:?}", n);
                        }
                        break;
                    }
                    timeout(5);
                }
                continue;
            }
```

AethanFoot: @mautamu can you check where n gets to in your use cases, as with my testing (where issues occurred before) seem to be fixed by the other changes outside the delay so n doesn't go above 0 with even 0 delay.

AethanFoot (approved):
This seems to have completely fixed any issues I was seeing. Great work!!!!

lex148:
Ok I have turned it down to `100ms` and you are correct it does feel much better. I am reluctant to go much farther but if it's needed its needed. I don't want to just hammer the CPU all the time checking focus.

AethanFoot:
Yeah awesome, that makes sense. 100ms should be ok for most, and is on the edge of perceived time. Maybe this should be set as a requirement that any action involving user input should take at most 100ms, for future reference.

VuiMuich:
I guess something broke with this commit. My two conky windows (using a modifierd `soothe-theme` with conky adopted from `dracula-theme` and therefore the `ibhagwan picom`, but don't think this actually relates) appear on top of everything.
With resetting `a2d994811166a3cdfdd98b39696c8bfba226b3e7` I get them behave correct again.

AethanFoot:
I assume that is because of this:
```
let unmanged: Vec<WindowHandle> = self
                    .xw
                    .get_all_windows()
                    .unwrap_or_default()
                    .iter()
                    .filter(|&x| *x != self.root)
                    .map(|x| WindowHandle::XlibHandle(*x))
                    .filter(|h| !wins.contains(&h))
                    .collect();
                let all: Vec<WindowHandle> = unmanged.iter().chain(wins.iter()).cloned().collect();
                self.xw.restack(all);
```
Assuming conky is an unmanaged window, meaning this behaviour needs changing.

VuiMuich:
well generally I think it is the correct idea to have unmanaged windows on top.
(btw: typo `unmanged`)

So maybe we need to make conky register as `dock` maybe?

AethanFoot:
Ummm yeah, not sure if making it a dock would create more weirder behaviour. With the conky I made it appears to be managed, with a windows type of Desktop. So it may be better to create special case for window type Desktop (it also has a state of below which could be used).

VuiMuich:
Oh right, `Desktop` window type makes even more sense.

I'll check my conky windows later today.

VuiMuich:
Hmm, there is probably going on something else on my system, since I just noticed that they dont have the "Conky" window-class as configured.
Also setting window type in conky to `desktop` makes it behave weird. Leftwm gets confused when calculating the desktop area and window sizes.
My conky windows are set as type `override`.
`xwinfo` looks quite different for me then yours:
```
xwininfo: Please select the window about which you
          would like information by clicking the
          mouse in that window.

xwininfo: Window id: 0x800002 (has no name)

  Absolute upper-left X:  5
  Absolute upper-left Y:  1135
  Relative upper-left X:  5
  Relative upper-left Y:  1135
  Width: 621
  Height: 300
  Depth: 32
  Visual: 0x6f
  Visual Class: TrueColor
  Border width: 0
  Class: InputOutput
  Colormap: 0x800001 (not installed)
  Bit Gravity State: ForgetGravity
  Window Gravity State: NorthWestGravity
  Backing Store State: Always
  Save Under State: no
  Map State: IsViewable
  Override Redirect State: yes
  Corners:  +5+1135  -1934+1135  -1934-5  +5-5
  -geometry 621x300+5-5
```

AethanFoot:
So ye, in my conky I have set (whole config https://github.com/AethanFoot/dotfiles/blob/master/.conkyrc):
```
own_window_type = 'desktop',
own_window_hints = 'undecorated,below,sticky,skip_taskbar,skip_pager',
```
So this is easily changeable, currently as Desktop type has no special code it gets treated like a Normal window (gets tiled). Setting it to dock seems to work (Don't know why thought it wouldn't).

VuiMuich:
Oh right, desktop being tiled explains the behavior I saw perfectly.
So desktop type should probably default to `must-float` and `below`, right?

AethanFoot:
As type dock seems to work (with the hints), maybe we should treat type Desktop like Dock (or mostly the same). Then maybe make a note in the wiki or something about needed settings for conky.

VuiMuich:
I think I found one of the expected quirks when using dock: only one of the two conky windows spawns. 
Gotta investigate more if this is really the cause, but it seams plausible..

PR #265, opened 2021-04-01
Fixes an issue from #263 where leftwm-state printed blank output. Cre…

mautamu:
Howdy, 

this PR fixes a degradation in #263 where leftwm-state reported blank lines. Thanks @bmclean2 for finding this issue!

Best,
Mautamu

### DIFF
--- a/src/bin/leftwm-state.rs
+++ b/src/bin/leftwm-state.rs
@@ -120,10 +120,15 @@ fn template_handler(
 
     let mut output = template.render(&globals).unwrap();
     output = str::replace(&output, "\r", "");
-    if !newline {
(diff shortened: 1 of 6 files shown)


carrascomj:
I had just started to write an issue about this. It solves it for me on arch, thanks for the quick fix!

mautamu:
Appears Azure is having some issues. @mautamu whenever Azure is back up and I’ll submit another minor patch so this will compile.

mautamu:
> I had just started to write an issue about this. It solves it for me on arch, thanks for the quick fix!

Haha! Glad I could get to it quickly then.

PR #267, opened 2021-04-02
Manually create more consistent behaviour in reorder_vec and added method cycle_vec to provide better behaviour to monocle and similar layouts

AethanFoot:
Manually controlling the behave around the ends of the list allows for the behaviour to be the same and switch items in the middle of the list. No 100% if this is the best code (let me know and I will update it). This further breaks the window cycling in monocle layout, thinking about creating a custom method to handling the reordering differently where the focus is always on the head and doesn't follow the window through.

### DIFF
--- a/src/bin/leftwm-state.rs
+++ b/src/bin/leftwm-state.rs
@@ -120,10 +120,15 @@ fn template_handler(
 
     let mut output = template.render(&globals).unwrap();
     output = str::replace(&output, "\r", "");
-    if !newline {
+    // We use newline rather than !newline to avoid negative logic,
+    // but note the difference between print! and println!. Trying to skip println!
+    // will result in theme degradation, as in #263.
+    if newline {
+        print!("{}", output);
+    } else {
         output = str::replace(&output, "\n", "");
+        println!("{}", output);
     }
-    print!("{}", output);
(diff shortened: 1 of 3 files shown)


AethanFoot:
Anyone more rusty than me, let me know if there is a better way of doing the reorder_vec logic. Monocle Layout now works as desired with cycle_vec.

VuiMuich:
AFAIK is a expression like
`temp_list.iter_mut().for_each(|index, item| { … })`
more efficient then a `for … in …` but I ain't a rust expert (nor educated programmer) neither.

AethanFoot:
Just did a test for curiosity, printing numbers in the range 0..999999. for_each did it in 96s, the for loop did it in ... 96s 🤣. So at least in this basic test speed seems to be a non-factor. Then I feel the for loop is more readable and you immediately can recognise that there is a loop there (when looking over the code afresh).

VuiMuich:
Hmm maybe they are more efficient, when you use them with additional conditions, like `.filter()` or `.find()`. 
Thanks for the test though.

lex148:
@AethanFoot same story on this one. Tested and ready to go. Its ready to be merged, just have to resolve conflicts first.

Issue #137, opened 2020-10-25
Feature request deck layout


sillen102:
Hi.

I really, really like this project. I think you are doing an amazing job. I really like many of the features in LeftWM but there are a couple of things that at least I would need in order to be able to use this WM full time. One of those being the deck layout. It's a layout I use a lot in dwm.

This is how it works:
The screen is split in two between the master and the stack. When you open a new window it's placed on the stack, just like in lwm default layout. But in the deck layout the windows in the stack are placed on top of each other and not split. Like a deck of cards, hence the name. Basically it allows you to always have a two split no matter how many windows you have open and then you can switch focus between the windows in the stack, switch a window from the stack with the master etc.

It would also have been neat to implement the _NET_WM_STATE_HIDDEN flag for windows not in focus so that if you are using a terminal with transparency the windows behind it wouldn't be seen through if you set up opacity rules in picom.

Deck :
```
+--------------+-------------+
|              |             |
|              |             |
|              |             |
|       M      |      S1     |
|              |             |
|              |             |
|              |             |
+--------------+-------------+
```

lex148:
I like this idea. Should be simple enough. I will see what I can do.

grv07:
I am also waiting for deck layout feature.
Any update will be grt :)

VuiMuich:
@lex148 are you working on this layout? If not, I might give this a shot, this looks like a mix of _main + stack_ and _monocle_.

(also might try to generalise the the focus fix for monocle)

hertg:
@VuiMuich  
Was just browsing the open issues, i think this can be closed, as PR #269 was merged?

VuiMuich:
Hey, thanks for this heads up.
I guess you are right, this issue should be closed by #269.
Since I can't close issues, as I am simply a normal contributor, we would nee to ping @sillen102 or @lex148 (theoretically @mautamu should be able to close issues as well, right?) to close this.

PR #268, opened 2021-04-03
Even more clippy lints

mautamu:
Howdy,

To avoid a repeat of #263/#265, I'm trying these in smaller batches by lint category. This PR performs the Panic! documentation lint ~~(partially~~ complete ~~)~~, the needless_pass_by_value, and a few other minor lints. It also accidentally includes a reverted #165 as I was testing that and didn't realize what was up.

Best,
Mautamu

PR #269, opened 2021-04-03
Deck layout & Opinions Requested

AethanFoot:
As requested in #137. This creates a main and a window stack on the right side. Right now when moving windows on stack they are cycled like with monocle (with the cycle_vec addition), but the cycling doesn't switch out the main window. To switch out the main window you have to use the MoveWindowTop command. Focus currently just swap between the top two windows (this code is quite bloated for what it is as currently it is the same as normal focus but I just strip out and use the top two windows).

Is this how we want this to behave? (As is it works but feels like the code is bloated)


mautamu:
Indeed, it does feel a touch bloated. Clippy makes some recommendations for how to fix this (am actually working on that right now haha). The general pattern is to include only that which is absolutely necessary in an `if let` rather than a match:
```rust
              let mut to_reorder =
                  helpers::vec_extract(&mut manager.windows, for_active_workspace);
              let is_handle = |x: &Window| -> bool { x.handle == handle };
              helpers::reorder_vec(&mut to_reorder, is_handle, -1);
              manager.windows.append(&mut to_reorder);
              let mut act = DisplayAction::MoveMouseOver(handle);
              if let Some(crate::layouts::Layout::Monocle) = layout {
                    // For Monocle we want to also move windows up/down
                    //Not the best solution but results
                    // in desired behaviour
                    let new_handle = match helpers::relative_find(&to_reorder, is_handle, 1) {
                        Some(h) => h.handle.clone(),
                        None => {
                            return false;
                        }
                    };
                    helpers::cycle_vec(&mut to_reorder, -1);
                    manager.windows.append(&mut to_reorder);
                    act = DisplayAction::MoveMouseOver(new_handle);
                }
                else if let Some(crate::layouts::Layout::MainAndDeck) = layout {
                    //Don't cycle main window
                    let main = to_reorder.remove(0);
                    let new_handle = match helpers::relative_find(&to_reorder, is_handle, 1) {
                        Some(h) => h.handle.clone(),
                        None => {
                            return false;
                        }
                    };
                    helpers::cycle_vec(&mut to_reorder, -1);
                    to_reorder.insert(0, main);
                    manager.windows.append(&mut to_reorder);
                    act = DisplayAction::MoveMouseOver(new_handle);
                }
                manager.actions.push_back(act);
                true
```
(Not thoroughly checked, just comparing to my working tree, will push tree up soon with diffs if you want to see more). 

IMHO it's still pretty bloated, but I think we might be able to clean it up some more. Might also factor out some of it to another function so that we follow DRYS.

AethanFoot:
After these bits got merged I was going to go through the command_handler to move repeated code to functions (especially around the focus/movement commands). Shall I try and start cleaning it now or shall we start after the current features are merged and then start cleaning to avoid conflicts as much as possible.

mautamu:
I'll push my work up in probably thirty minutes or so, and then we can prolly start working on the function refactor once we know the match refactor doesn't make a mess.

mautamu:
I pushed to https://github.com/mautamu/leftwm/tree/refactor-command-handler. 

I also noticed a regression regarding floating windows that I'm working on now. I think this might also impact #268 so I'll mark that as a draft for now.

AethanFoot:
I have discovered an issue where upon a return false in at least MoveWindowUp/Down will crash leftwm. Should we maybe just continue out of the if statement and append the windows back without any changes?

AethanFoot:
There is also the issue when new invisible windows are created the focus doesn't know what to do and moves the mouse to the center left of the left-most screen.

VuiMuich:
> There is also the issue when new invisible windows are created the focus doesn't know what to do and moves the mouse to the center left of the left-most screen.

In the window_handler the `focus_handler::focus_window(manager, &window.handle);`
could be restricted by an `if window.visible()`.

AethanFoot:
Ye I think I tried restricting to if window.visible(), but the layout isn't applied at that point which means the window isn't yet invisible, so it has no effect.

AethanFoot:
To elaborate on this the window get created here:
```
else => {
        event_buffer.drain(..).for_each(|event| needs_update = handler.process(manager, event) || needs_update)
}
```
Which then says an update is needed, which incurs various things to be set including the layout:
```
if needs_update {
            match &manager.mode {
                Mode::Normal => {
                    let windows: Vec<&Window> = manager.windows.iter().collect();
                    let focused = manager.focused_window();
                    display_server.update_windows(windows, focused);
                    let workspaces: Vec<&Workspace> = manager.workspaces.iter().collect();
                    let focused = manager.focused_workspace();
                    display_server.update_workspaces(workspaces, focused);
                }
                //when (resizing / moving) only deal with the single window
                Mode::ResizingWindow(h) | Mode::MovingWindow(h) => {
                    let focused = manager.focused_window();
                    let windows: Vec<&Window> = (&manager.windows)
                        .iter()
                        .filter(|w| &w.handle == h)
                        .collect();
                    display_server.update_windows(windows, focused);
                }
            }
        }
```
So the only thing to do would be to allow the layout to alter focus, like you suggested before. Will look into this.

mautamu:
Oh cool! Glad those bugs got spotted, and sounds like a solid approach to fixing them! 🙂

VuiMuich:
> To elaborate on this the window get created here:
> 
> ```
> else => {
>         event_buffer.drain(..).for_each(|event| needs_update = handler.process(manager, event) || needs_update)
> }
> ```
> 
> Which then says an update is needed, which incurs various things to be set including the layout:
> 
> ```
> if needs_update {
>             match &manager.mode {
>                 Mode::Normal => {
>                     let windows: Vec<&Window> = manager.windows.iter().collect();
>                     let focused = manager.focused_window();
>                     display_server.update_windows(windows, focused);
>                     let workspaces: Vec<&Workspace> = manager.workspaces.iter().collect();
>                     let focused = manager.focused_workspace();
>                     display_server.update_workspaces(workspaces, focused);
>                 }
>                 //when (resizing / moving) only deal with the single window
>                 Mode::ResizingWindow(h) | Mode::MovingWindow(h) => {
>                     let focused = manager.focused_window();
>                     let windows: Vec<&Window> = (&manager.windows)
>                         .iter()
>                         .filter(|w| &w.handle == h)
>                         .collect();
>                     display_server.update_windows(windows, focused);
>                 }
>             }
>         }
> ```
> 
> So the only thing to do would be to allow the layout to alter focus, like you suggested before, or do your other suggestion with placing new windows on the top of the stack in these layouts. Will look into this.

When I added the `apply margin multiplier on window creation` I followd the `window_handler::process.created(...)` backwards to a call from `DisplayEvent::WindowCreate` ind `DisplayEventHandler::process(...)`

Maybe there are several attempts to set the focus on the new window?

BTW: when I tried to get  logging set up I didn't see any changes unless I restarted `leftm-bin` (unfortunatelly I didn't try hard-reset or `pkill leftwm-worker` so I can't tell if that would've worked as well). At the same time my statement I added started to work as well. It's a bit odd, since most changes work right after compiliation+soft-reload, but this didn't.

AethanFoot:
Quick update, I think I have found another part of the problem when AddedWindow is called down the line in xwrap the function move_cursor_to_window is called.

AethanFoot:
For now set it so if a new window is invisible (set if they are in monocle or deck) they don't take focus or cursor. Will look at this later or tomorrow to consider spawning new windows in these layout on top of their respective stacks + fix the crash issue.

VuiMuich:
I am wondering, what keeps the windows from spawning on top, as `window_handler::process.created(…)` calls `manager.sort_windows()` wich actually should put the new window on top.
Or is this just referring to the `memory Stack` not the `stack of windows`?
Then maybe we could call `move_to_top()` depending on the layout before calling the focus_handler.

AethanFoot:
Well looking at the code to new windows a put on the end of the array (or the top of the stack depending on how you look at and treat the vector). Then sort_windows doesn't look any deeper than just what type the window is, so the new window will still be below the other windows in its workspace (actually be last in the level3 list). So to get around this I just ripped out the windows on the same workspace to make sure that the new window is before them (or second when it comes to the deck layout).

lex148:
@AethanFoot great job with this one. Sorry I wasn't paying attention when doing PRs this morning. I merged the refactor one first and now your PR has conflicts. I have tested it and It all looks good. I will be happy to hit the merge button when the conflicts are resolved. If you want help fixing them just let me know

AethanFoot:
Thanks, they should be easy enough to fix, just doing some minor cleanup in command handler to reduce repeated code. When I push through the commits it should be good to merge.

AethanFoot:
This should be done, this also closes #267 as it includes that code. Let me know if you guys discover any bugs. 🙂

VuiMuich:
This also closes #137, right?

mautamu:
> This also closes #137, right?

We might also want to implement _NET_WM_STATE_HIDDEN as per #137 before closing it.

VuiMuich:
Oops, missed that one sorry.

Do I get this right, that it would do basically the same as we are enforcing in `deck` and `monocle` during `layout::update(...)` but also telling EWMH about it, right?

VuiMuich:
@AethanFoot just a small possibility for a little buff if you have an idea for a quick fix: when the `main` has the focus it would be nice if `window-up/down` still would cycle through the deck.

PR #272, opened 2021-04-04
Refactor large functions into smaller ones

lex148:
@mautamu I appreciate your help cleaning up the code base. I thought I would help out too. I don't want to step on your toes with this project so if you want to take a different route that's cool with me. Here is a start of a refactor of the two big switch blocks. I have taken the content of each branch and pulled them out into a function. That is all. There should be no logical changes in this PR. 

My thinking here is its much easier to work with smaller functions. I could organize them into different files or refactor the contents individually. this is just a step one...


mautamu:
Thank you so much! This is exactly what I think we were aiming for in the #269 discussion, for the same reasons you describe.🙂

PR #274, opened 2021-04-05
Bump LeftWM to v0.2.8, update signature to resolve all non-cast/non-mut-mut pedantic lints

mautamu:
Howdy,

This PR will do the following:
* Update LeftWM/master to v0.2.8
* Update all dependencies to their current published version
* Update README to clear up some confusion 
* Update CHANGELOG to reflect all major changes since 0.2.6
* Update function signature to solve a Clippy lint

Rationale:
We have accumulated quite a number of substantial changes since 0.2.6 thanks to several people (thank you all, these changes have been wonderful)! :) We have six new layouts, a new external command (maybe two), a new binary, and three new configuration options. We also have a fair amount of new documentation available for https://docs.rs/. 

A few things to consider:
If #259 closes before this, we probably should add the attached patch to the CHANGELOG:  
[259.patch.txt](https://github.com/leftwm/leftwm/files/6261114/259.patch.txt). Also might be helpful to give #259 a day or two of testing prior to releasing 0.2.7. 

We'll also likely want to update the PKGBUILDs to have leftwm-check as well:
```
install -Dm755 leftwm leftwm-worker leftwm-state leftwm-check -t "$pkgdir"/usr/bin
```

Finally, we might work toward de-cluttering the git repositories. It might be helpful to delete stale branches as part of the larger "spring cleaning" effort we've been on of late. 

Best,
Mautamu


AethanFoot:
Another point I was thinking of mentioning , in terms of decluttering, was to go through the issues and closing resolved ones, labelling them where possible to make it easier to go through them, and maybe adding an indicator of urgency (and maybe difficulty). This is obviously tedious but would make it easier for users to find an issue that will help them, and make it easier for someone to find an issue they could try and solve.

lex148:
Everyone really has done an amazing job, and you are correct @mautamu its time for a release. 
good work team!

VuiMuich:
I believe there is a minor thing in the CHANGLOG:
Shouldn't be line 36 be like:
```
- FloatingToTile command that allows a floating window to be tiled again in config.toml
```

Issue #177, opened 2021-02-08
Support for polybar occupied indicator.


aTosser:
I may just be very bad at configuring polybar, but nothing I do gets the tag occupied indicator to work. (an indicator to show if your have a window open in a tag at a glance) I assume that means it is not supported by leftwm. If that is the case if that could be added to a planner for later support.

lex148:
sure, can you add a copy of the section of your polybar config you want us to support? That way who ever picks up the ticket knows exactly what you are talking about.

carrascomj:
Also, take a look at https://github.com/leftwm/leftwm/pull/146 and the liquid template being used. The workspace module for polybar I used was [this one](https://github.com/b4skyx/leftwm-soothe/blob/986a1afede9daa77ac05e90af96ed22a0b8a3fb9/theme/polybar.config#L70).

VuiMuich:
Found this issue to late, after I stumbled around quite a lot to figure out how to get my template the way I like.
The mention of PR-146 is helpfull. I think I'm gonna work on a wiki page about the template.liquid..

Here is template, which hides empty tags:
```liquid
{% for tag in workspace.tags %}
{% if tag.mine %}
%{A1:$SCRIPTPATH/change_to_tag {{workspace.index}} {{tag.index}}:}
%{F#FFFFFF}%{B#455A64} | {{tag.name}} | %{B-}%{F-}
%{A}
{% elsif tag.busy %}
%{A1:$SCRIPTPATH/change_to_tag {{workspace.index}} {{tag.index}}:}
%{F#90A4AE}  {{tag.name}}  %{F-}
%{A}
{% else tag.visible  %}
%{A1:$SCRIPTPATH/change_to_tag {{workspace.index}} {{tag.index}}:}
{{""}}
%{A}
{% endif %}
{% endfor %}
%{c}
{{ window_title }}
```

By the way can anyone explain to me, what the different between these two states pulled from the original template is (besides the different polybar-colors ofc), I fail figure it out smh:
```liquid
...
{% elsif tag.visible  %}
%{A1:$SCRIPTPATH/change_to_tag {{workspace.index}} {{tag.index}}:}
%{B#282A36}%{F#D19A66}   {{tag.name}}   %{B-}%{F-}
%{A}
{% else tag.visible  %}
%{A1:$SCRIPTPATH/change_to_tag {{workspace.index}} {{tag.index}}:}
%{F#546e7a}   {{tag.name}}   %{F-}
...
```

[...]

PR #279, opened 2021-04-11
Docs: add busy indicator to basic themes

carrascomj:
Related to #177 
### Description

Since this feature needs some documentation, a good way is to just show it to newcomers. 

__1__ is visible (other monitor), __2__ is busy (with windows, but not visible) and __3__ is owned tag:

* Polybar
polybar

* XMobar
xmobar

* Lemonbar
lemon

I tried to respect the style of the theme, but any change is welcome!

### Implementation
Simply modified the `template.liquid` files for each basic theme.


mautamu:
Good work! I think this will be a nice thing to add to the basic themes.

PR #283, opened 2021-04-13
cycle the deck when main has focus

AethanFoot:
As requested in #269, the deck will cycle when main has focus.
Thanks.


mautamu:
Thank you, this is great!

VuiMuich:
Great, thanks for this quick fix!

PR #284, opened 2021-04-13
minor clippy fixes

AethanFoot:
Some minor clippy fixes caused by recent pr's.


VuiMuich:
did you get them with this clippy command?
`cargo clippy -- -W clippy::pedantic -A clippy::must_use_candidate -A clippy::cast_precision_loss -A clippy::cast_possible_truncation -A clippy::cast_possible_wrap -A clippy::cast_sign_loss -A clippy::mut_mut`

Just wondering, why they slipped on my check for the PR..

AethanFoot:
Yeah, it is slightly cluttered output because of all the casts, which is annoying.

mautamu:
Are the casts not turning off with all the -A flags? If not we might allow them in the crate route.

AethanFoot:
Ah with that command it does, I had a different one in my history that I used. That is much better.

PR #285, opened 2021-04-13
Add recent clippy lints to the default 

mautamu:
Howdy, 

As discussed in #284, the clippy lints can be cumbersome to type out as a command. This PR adds the clippy lints used in recent PRs to the root of the crate so that they execute on `cargo clippy`. 

The disadvantage to this method is that CI may flag new clippy pedantic lints causing builds to fail. This is not quite unlike now, where new "all" clippy lints have the same effect. The advantage is that we might end up with more correct code more often. In future PRs we should watch for #[allow()]s and make sure to add a comment describing the reason why something needs allowed. It might also be helpful to periodically review the things that have been allowed to check whether we can think of a better implementation.

If anyone would like to modify which lints are globally allowed/warned, feel free! These are just the one's we've been using of late.

Best,
Mautamu


VuiMuich:
Cool. Do I get this right, that with this commit a simple `cargo clippy` that passes will also pass the PR-check?

I was just about to set me up some simple 'cargo-preflite' script to quicker check clippy and fmt before committing, with this it would merely be an alias for `cargo clippy & cargo fmt -- --check`

mautamu:
I think you still need `--all-targets --all-features`, but it will check `-W clippy::pedantic -A ....` so the latter won't be necessary. Please correct me if I'm misunderstanding. If I remember correctly, there is a way to tell Clippy (in maybe clippy.toml?) to automatically use `--all-targets --all-features` but I can't recall what it is off the top of my head.

On the other note, here's what I've been doing in my .bashrc for the preflight script (add +stable after cargo to use the stable version):
```bash
# Example use (adds all of src/ and then the commit message):
# checkleft src/ "This is the commit message"
checkleft() {
	cargo fmt
        cargo clippy --all-targets --all-features
       	git add -A "$1"
       	git status
       	git commit -S -m "$2"
}
```

It's a bit crude, and will keep going even if something fails.  Could rewrite to use if:
```bash
# Example use (adds all of src/ and then the commit message):
# checkleft src/ "This is the commit message"
#UNCHECKED
checkleft() {
	if cargo fmt ; then
            if cargo clippy --all-targets --all-features ; then
       	        if git add -A "$1" ; then
       	            if git status ; then
       	                git commit -S -m "$2"
                    else
                        echo "Status failed"
                     fi
                 else 
                     echo "Add failed"
                fi
            else 
                echo "Clippy failed"
            fi
         else
                echo "FMT failed"
          fi        
}
```

VuiMuich:
Oh, this is way more sophisticated then my humble plan. Definitely gonna check the out.

PR #290, opened 2021-04-14
Refactor external command parsing

### LINKED ISSUE #289: Question about parse_command
Is there a reason that an if/else statement is used here. I feel it would be much more readable (and performant? as there is not a constant call to starts_with) as a match statement with the first word is the string compared. The first word would be extracted by splitting the string (with whatever variation of split), then using iter.next. Then any command with more than one word can have the iterator passed ready for the function.
Does this sound reasonably, if so I'll start going through it.
Thanks.

mautamu:
Howdy,

This PR adds code I had stashed, closes #289. Should probably double check that all commands were added and that this works (I don't usually use external commands).

Best,
Mautamu

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -739,9 +739,9 @@ dependencies = [
 
 [[package]]
 name = "redox_syscall"
-version = "0.2.5"
+version = "0.2.6"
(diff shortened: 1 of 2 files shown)


AethanFoot:
Awesome, that is much better!

PR #293, opened 2021-04-15
Flip layout

### LINKED ISSUE #273: Feature Idea: `flip layout`
Add the possibility to flip asymmetric layouts.

For some it might even make sense to flip not only horizontally but even vertically.

This could make sense both as config option, as well as command.

Horizontal Flip:
`MainAndVertStack`, `MainAndDeck`, `LeftWiderRightStack`, `CenterMain` and `CenterMainBalanced`, `Fibonacci`

Vertical Flip:
 `Fibonacci` and `MainHorizontalStack`

Both:
 `Fibonacci`

AethanFoot:
Adds command for flipping vertically and horizontally (external coming after the command parsing change is merged). This functionality is implemented on the layouts side (I tried my best to update some of them, they could be implemented much better). The flip is saved per tag. Closes #273.
Thanks.


src/command.rs
```diff
@@ -24,6 +24,8 @@ pub enum Command {
     NextLayout,
     PreviousLayout,
     SetLayout,
+    FlipTagHorizontal,
```
mautamu, line 27: If we're gonna use shorthand (FlipTagVert), we might want to do so consistently.

mautamu (commented):
This is awesome!!! So excited for this! Good work.

Just the one note about commands.

src/command.rs
AethanFoot: I'll change horizontal to horz. Should I update all of the horizontals and verticals to their short hands (most the verticals are in the layout names)

mautamu: Hmm, I don't see much other shorthand in the code to inform an opinion, and don't have an existing opinion on it myself. Sounds like a lot of work for not too much gain.

VuiMuich: I believe I remember stumbling over this inconsistency once in another place and was thinking I need to check later which one would be the more common used, but I don't remember where that was..

VuiMuich:
Aethan seems to be on a spree lately :smile: it's amazing. Great work!

lex148:
@AethanFoot first let me say great work. This is really cool and a nice feature to add. I do have a couple thoughts about it that I think we should talk through before merging since this adds a new command. 

1) I'm not sure this logic should be tied to the Tag. I think It might be better to tie to the layout. I use one horizontal screen and one vertical one. When you tie this to the tag, the flip affect follows the tags around to the other monitor. This is less ideal for me at least because I want to flip on only one screen/layout. What does everyone else think here? 

2) Can you think of any instance where you would want to flip vertical AND flip horizontal? My thinking here is would one single generic command be better than two specific ones? What do you think about `MirrorTag` or `MirrorLayout` if we put this logic on layouts. If we have a single command then the same key combo works on many layouts which would be a plus. 

3) If you really want to go with two commands here, lets use full names. No abbreviations. I don't care if you abbreviate in the code base, but commands are a different story. they are user facing.

Again awesome work, I am being a bit picky here because we are adding new commands. New commands are a bit like working in permanent ink, once we add them users expect them to be there and expect their behavior to stay mostly the same, even if that behavior is "broken"

AethanFoot:
> I'm not sure this logic should be tied to the Tag. I think It might be better to tie to the layout. I use one horizontal screen and one vertical one. When you tie this to the tag, the flip affect follows the tags around to the other monitor. This is less ideal for me at least because I want to flip on only one screen/layout. What does everyone else think here?

I may not be correct on this, but if the were to be tied to the layout would it not have memory (e.g. if you switch from that tag this change is removed). This would then be frustrating as each time you return to the tag it would be different, I suppose this is the other side to your frustration. Then the other place would be the workspace which would go against the tag-centric flow we are going for in 0.3.0. I can see both sides of annoyance, but feel allowing the user the ability to setup each tag in their own way would be ideal (which would be layout (flipped or not), applications, etc.). 

> Can you think of any instance where you would want to flip vertical AND flip horizontal? My thinking here is would one single generic command be better than two specific ones? What do you think about MirrorTag or MirrorLayout if we put this logic on layouts. If we have a single command then the same key combo works on many layouts which would be a plus.

One instance has already appeared with the fibonacci layout, which allows you to flip in both directions. I suppose there could a general command which changes both (for most layouts won't matter), but we would still want both to be toggleable for layouts like fibonacci.  

> If you really want to go with two commands here, lets use full names. No abbreviations. I don't care if you abbreviate in the code base, but commands are a different story. they are user facing.

I agree with this, I think the abbreviations got stuck in my head where the main layout I use is the main_and_vert_stack.

I don't mind nitpicking as it often helps improving my ability, and the little details are most important. I'll look at updating things later once @mautamu or @VuiMuich have given their opinions.
Thanks.

AethanFoot:
I also moved the theme loading to a function in external_command_handler as clippy was warning of line count in the function.

VuiMuich:
I fully agree, that a single `FlipLayout` command is way more userfriendly to configure.
And since right now the fibonacci layout is the single corner case where both flip axes are useful w might allow a corner-case handling, by cycling through all four layout projections with a single command. Also flipping layouts is probably not something you will be doing all the time, but only every now and then, when you change the setup of this tag.

> 1. I use one horizontal screen and one vertical one. When you tie this to the tag, the flip affect follows the tags around to the other monitor. This is less ideal for me at least because I want to flip on only one screen/layout.

What do you think about handling 'special behaviour' with optional parameters in the value field?
We could have `value = "w:0"` or `value = "tags:[1,2,8,9]"` to restrict the flip-command to only these workspaces/tags. Or we could  follow the scheme of `leftwm-state` to allow even more complex configurations, like for example a tripple-workspace setup, where the `main` is always close to the centre screen.
This might be tricky to parse and to document clearly, so it's definitely something for the advanced user. But it would be pretty cool imo and I believe quite unique among other WMs.

> I agree with this, I think the abbreviations got stuck in my head where the main layout I use is the main_and_vert_stack.

I believe this was also the one I had in mind, when we discussed abbreviations earlier. Maybe we could give `leftwm-check` the ability to error kind of like the compiler does with variables/functions that are not found, but a similiar one is present (gonna look up the Error-Number later). This way we could not only sneak in slight name changes and let the user know we broke his config :P but also better handle small typos when reporting broken configs..

AethanFoot:
Ok, just thinking quickly, the idea for a single command which will cycle through the possible permutations of the booleans. This vector of possible permutations can be accessed through the layout with each layout defining the possible array (most of these would return the same with horizontal flipping). Then the index in the array could be stored in the tag, which could be incremented by the command (hope this makes some sense). 
With the restriction I'll put thought into this tomorrow (we could store which workspace this change was made but the would be more things stored tag).
Thanks.

AethanFoot:
I have reduced it to one command, RotateTag (or would RotateLayout or something else be better), this will check would permutations of flippingness is possible for the layout and cycle through them. To fix your issue I could reset the rotation index and the bools in could be reset within gototag (this could be an option if the user wants this behaviour). Any other things you think I should change let me know (or any changes I can make to help with merging to your safer code branch).
Thanks.

PR #294, opened 2021-04-17
Safer code

lex148:
The goal of this branch is to:
- remove all unwrap calls ( If they can't be removed, convert to expect ) 
- remove unsafe accessing of Vec [index] convert to .get(index) so rust will bounds check


src/utils/helpers.rs
```diff
@@ -59,18 +53,16 @@ where
...
     let len = list.len() as i32;
-    let (index, item) = match list.iter().enumerate().find(|&x| test(x.1)) {
-        Some(x) => (x.0, x.1.clone()),
-        None => {
-            return;
-        }
-    };
+
+    let index = list.iter().enumerate().find(|&x| test(x.1))?.0;
+    let item = list.get(0)?.clone();
```
AethanFoot, line 64: doesn't this need to be index not 0?

AethanFoot (commented):
This is great, most definitely taking some mental notes on this.

src/utils/helpers.rs
lex148: good catch thanks!

lex148:
Ok, This is ready for others to look at if you want. 
A couple notes about this PR:
- I didn't touch unwraps in tests. This seams appropriate. If a test panics you most likely want to know
- My intent was to not change functionally anywhere with the one exception of the mutability cleanup in ExternalCommandHandler

PR #295, opened 2021-04-18
Fix crashes in some corner cases

VuiMuich:
When calling `move_window_change` with just a single window in `to_reorder` the worker crashes.


AethanFoot:
Does this crash happen when cycle_vec or reorder_vec is called, as in reorder_vec I can see a possible crash as the rotating would be applied to an empty slice. The best fix would be to go straight to the cause and add a check to the top of reorder_vec (also maybe cycle_vec to stop any unneeded work).

VuiMuich:
I believe it happens in both `reorder_vec` and `cycle_vec` as it not only occurs with `monocle` or `main_and_deck` but also with `center_main_balanced`

I agree if it is one of the helpers crashing it would be better to fix the helper, I'll check that.

Also: just noticed a crash with `main_and_deck` when creating a window on an empty tag, maybe related?

AethanFoot:
No, my edge case testing is real bad atm darn, that crash would be in window_handler with inserting the new window at index 0. To fix that you could add an or check whether to_reorder is len == 0 with the check if it is monocle (inserting the new window at zero. The snippet would be:
```
        let mut to_reorder = helpers::vec_extract(&mut manager.windows, for_active_workspace);
        if Layout::Monocle == layout || to_reorder.len() == 0 {
            to_reorder.insert(0, window.clone());
        } else {
            to_reorder.insert(1, window.clone());
        }
        manager.windows.append(&mut to_reorder);
```

VuiMuich:
From what I can tell so far it is probably `rotate` which panics, ant this would mean that `change` is bigger then `list.len()` if I read the doc of `rotate` correct.

Thanks, I figured it would probably be in `window_handler` but wanted to push this fix first and then got distracted by a weird focus behaviour when a tag is mixed with tiled and floating windows.

AethanFoot:
Ye that makes sense, so I would put a check at the top of reorder_vec whether len < 2, and a check whether the shift is less then length for cycle_vec (we could subtract len from the shift until shift is less then length, but this more then likely not going to be needed).

AethanFoot:
Just to add, I would also keep the check in common_vars as this will reduce unnecessary work.

VuiMuich:
> Just to add, I would also keep the check in common_vars as this will reduce unnecessary work.

lol, yes, it looks like fixing this in the helpers will be quite the rabbit hole, as it looks like `relative_find` might be affected as well..
I'll just leave some comments that these functions don't like lists with `len ==1`

Issue #296, opened 2021-04-18
Several odd behaviours with focus/move and floating windows (some of them)


VuiMuich:
When a workspace contains both floating and tiled windows I observed some weird behaviours:
- with `main_and_deck` layout the focus is restricted to switching between the first two items of the windows list, so when one of these is a floating window this breaks
- when a floating window covers the area where the mouse-pointer will be put by `move_mouse_over` it happens that the floating window catches the attempt to change the focus to the underlying tiled window
- when moving windows up/down in `main_and_deck`  the focus stays on main but should probably move to the deck, right?
- With the layout `center_main_balanced` the order of the focus moving feels unnatural: it hops left-right all the time, where it probably should go through all windows on the left, then all on the right, maybe go to the center-one when switching side

Some ideas for fixes:
- keep floating windows always at the end of the list
- ignore floating windows in `move_window` functions, maybe with a check in `common_vars`
- if we agree on a new behaviour for `center_main_balanced` it would be probably a good chance to refactor the layout-specific logic of `move` and `focus` functions with match statements opposed to the if-else-chains they are right now.

jwnpoh:
Hello, if I may contribute an opinion, the `main and deck` layout has turned out to be one of the useful layouts. 

Use case: I'm working on something on main (Window 0), but need multiple references that I've put in the deck. So for example, I am working on Window 0 and have Window 1 next to it to reference. Occasionally I need to cross-reference to Window 2, so I cycle the deck to Window 2 and immediately continue typing with focus having remained on Window 0.

VuiMuich:
Hi, your opinion is very welcome, and I fully agree, that `man_and_deck` is probably the most useful layout I have ever used in any WM. I currently only use `main_and_deck`, `monocle`, `center_main_balanced` and `left_wider_right_stack` with the last two sharing probably less then 5% screentime.

I also experienced your use case and find it helpful, but believe it is not a predictable and intuitive behaviour, so I wanted to gather some opinions on that.
Lex was quite right, as he stated in another discussion, that if a behaviour is introduced (even if it's 'broken') it'll be hard to change away from it.

As a counter thought on your conclusion: with the explicit focus move needed this breaks another workflow that might be equally useful. For example you have on the `main` your editor for code and on the `deck` a bunch of windows with references some of them you regularly search in some other you regularly use to copy snippets from, a debug console and a terminal for compilation. So you see it might be equally valid to move the focus to the deck or keep it on main while cycling the deck.

[...]

PR #299, opened 2021-04-19
create 'correct' focusing in center_main_balanced

AethanFoot:
As suggested in #296.

### DIFF
--- a/src/handlers/command_handler.rs
+++ b/src/handlers/command_handler.rs
@@ -329,6 +329,26 @@ fn focus_window_change(
             let act = DisplayAction::MoveMouseOver(new_focused.handle);
             manager.actions.push_back(act);
         }
+    } else if let Some(crate::layouts::Layout::CenterMainBalanced) = layout {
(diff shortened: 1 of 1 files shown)


lex148:
@AethanFoot I'm having a bit of trouble with this one. I am unsure what you are doing and why. What problem are you solving here? I like to give everything a good test before merging, but I'm unsure what I should be testing.

AethanFoot:
@VuiMuich suggested in #296, to have the focus cycle first down the left then the right to prevent the cursor  and focus hopping side to side. I wasnt sure myself, but I thought I would implement to see if it felt more natural. 
I suppose the pull request is more to see if more people felt it is better or unnecessary. (I should have added a description but didnt have much time yesterday).

lex148:
@AethanFoot I haven't forgotten about this PR. I would like to try and do a little refactoring to make it clearer as to what is going on here. I have no problem with the change, I just would like to lower the bar a bit if anyone needs to come behind you.

PR #305, opened 2021-04-26
Cleanup in mostly handlers

AethanFoot:
This is mainly a cleanup and refactor to make the code more consistent and readable. The only logical changes are around increase/decrease_main_width where I generalised the function called and the logic to the tag.
Even though I didn't really change any logic, this should still be properly tested in case I have missed something.
Thanks.


lex148:
This looks great! I didn't see anything when looking over the code. Because it touches so many things I am going to run it all day tomorrow and make sure I don't see any regressions.

PR #306, opened 2021-04-28
When focusing closest window, check that it is still on that workspace

### LINKED ISSUE #297: Unexpected behaviour when switching tags with multiple monitor
Hello, I've recently encountered unexpected behaviour where LeftWM randomly switches focus tto another monitor/workspace by itself. Here's the scenario:

I have monitor 2 running some process on Tag 9.

I'm working on monitor 1 on Tag 1. I need to find something that I forgot which tag it's on, so I cycle through the tags one at a time - Tag 2, Tag 3, etc. Somewhere along the way, randomly, say for example when I switch to Tag 5, LeftWM changes the tag on monitor 2 and Tag 5 gets diplayed on monitor 2. Interestingly the mouse cursor (which I arbitrarily use to find where my focus is when I lose focus) remains on monitor 1 when this happens.

Any advice if I'm doing something wrong here? Thanks!

Edit: I think this might be related to #275 somehow. When this behaviour happens my polybar seems to display similarly erroneous output, except that I don't necessarily have Brave or Firefox open on those affected tags.

AethanFoot:
Should close #297. This prevents focus changing to different workspaces when an empty tag is focused.
Thanks.

### DIFF
--- a/src/handlers/command_handler.rs
+++ b/src/handlers/command_handler.rs
@@ -263,9 +263,6 @@ fn move_window_change(
 ) -> Option<bool> {
     let is_handle = |x: &Window| -> bool { x.handle == handle };
     if let Some(Layout::Monocle) = layout {
-        // For Monocle we want to also move windows up/down
-        // Not the best solution but results
-        // in desired behaviour
         handle = helpers::relative_find(&to_reorder, is_handle, -val)?.handle;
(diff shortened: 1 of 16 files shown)


lex148:
I have tested this one and it appears ready to go. If you fix the merge conflict I will merge

PR #311, opened 2021-05-03
Leftwm-Command

AethanFoot:
This is basically a wrapper for sending external commands into commands.pipe. It allows for simpler syntax in doing so and allows you to chain commands (where commands with args must be surrounded in quotes).
Thanks.


VuiMuich:
This is absolutely awesome and very helpful!

PR #312, opened 2021-05-03
Don't update other windows when in fullscreen

AethanFoot:
This removes unneeded work and also prevents an issue for monocle and main_and_deck where windows would be set visible behind a fullscreen window.
Thanks.

### DIFF
--- a/src/models/workspace.rs
+++ b/src/models/workspace.rs
@@ -132,29 +132,34 @@ impl Workspace {
     }
 
     pub fn update_windows(&self, windows: &mut Vec<Window>) {
-        //mark all windows for this workspace as visible
-        let mut all_mine: Vec<&mut Window> = windows
+        if let Some(w) = windows
             .iter_mut()
-            .filter(|w| self.is_displaying(w))
-            .collect();
-        match all_mine.iter_mut().find(|w| w.is_fullscreen()) {
-            Some(w) => w.set_visible(true),
-            None => all_mine.iter_mut().for_each(|w| w.set_visible(true)),
+            .find(|w| self.is_displaying(w) && w.is_fullscreen())
+        {
+            w.set_visible(true);
+        } else {
+            //Don't bother updating the other windows
+            //mark all windows for this workspace as visible
+            let mut all_mine: Vec<&mut Window> = windows
+                .iter_mut()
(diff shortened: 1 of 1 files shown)


VuiMuich:
This looks very interesting.

I'm gonna test this later, as this also might fix a little glitch that on reloads windows get resized for a moment, which sometimes could lead oddly horizontally "scrolled" vim buffers.

VuiMuich:
I actually don't have much time tonight to go into corner-case testing but after a brief initial test the glitch seems to be gone.

If anyone else want's to have a look into it: most notably it appeared to me, when I have a split in vim, after reload the left buffer view was smaller then the right one and when the cursor was at the end of a line the "horizontal scroll" was all to the right.

AethanFoot:
I can reproduce the issue where the right pane scrolls to the far right when the cursor is at an end of a line. However I feel like this is an issue with vim/neovim, as you can recreate this if you swap the window to a tag with another window (causing the terminal to squash). Also, atleast on my system, the reason for the slight resizes is that leftwm lauches faster than polybar which is unavoidable.

VuiMuich:
I agree that this is an issue, where vom fails to reproduce the panel size, when size of the terminal has changed and then changed back. 
My misconception with this PR was that I thought it might apply during reload, but naturally the visibility state of the window doesn't change it is just copied from `old`. 

And yes polybar being slow on reloads is quite annoying.

mautamu:
I know at one point we'd discussed having three types of reload:

- Settings (?): Reload theme.toml and config.toml, then redraw windows
- Soft: Reload theme.toml, reload config.toml, restart leftwm-worker maintaining state, and run up/down
- Hard: Restart leftwm-worker while discarding state

Would that be a valid solution to the polybar issue?

VuiMuich:
I believe we are safe to consider this not-opened issue closed :D

But you are right mautamu, I mentioned this idea of a pure config-reload and I am about to give it a try. It's just that the last two weeks were quite full at work, so I had little time to spend in code and decided to use this to finally finish up with rustlings-exercises (still 4 left now).
Also I forked SilverSoldier/rlock into VuiMuich/leftlock with the idea to more and more "rustify" this code as a further learning project.
So if anyone wants to take this immediately feel free.

PR #315, opened 2021-05-05
Initial named scratchpads: Some questions 

AethanFoot:
An initial implementation of named scratchpads. Right now I have 2 questions:
1.  What would be the best way of preventing the "bolted on" tag being displayed by leftwm-state/the bars, and should we allow the user, if they explicitly add it to there config, show the NSP tag.
2. How should we handle multiple windows with the scratchpad name, which happens if the user launches it their selves or when an execute command. Currently the new window gets toggled and the old window (which can be still interacted with) no longer gets toggled by the toggle command until the new window is closed.
Also any other points or alterations you guys can think of.
Thanks.


lex148:
I think if they add the tag, they are asking to display it and that is fine. 

As for (2), I think you need to go a slightly different approach. You are associating the window that is created when the scratchpad is open and the scratchpad its self based on the window name. This is problematic, worse the user has to know and set the window name in the config. 

I think a much better approach would be to store the PID of the application that boots from the execute command. You could then use that to lookup the window when the scratchpad is toggled.

AethanFoot:
Personally I feel that allowing the user to set the name is more standard, this is how others wm implement this. With most applications having a -t option to change the name (this then allows the user to easily start the app outside of leftwm). From your suggestion, I think what I can do is leave the name and "translate" that to the handle of the window. My thought is to change scatchpads to a hashmap (storing the names as keys), and on window creation set the value for the name to the handle of the window. If another window is opened with the correct name, we first checked if there is already an associated window (can either remove the handle on window close or check if the handle already associated is still open).
Does this make sense?

AethanFoot:
This should be a working solution, if anyone feels that something should be different (or is broken) let me know.
Thanks.

AethanFoot:
@lex148 I have implemented scratchpads using your pid method, so there can be a better comparison between. Does this align we your thought process? I do agree that this pid method is more robust.

AethanFoot:
Apt update seems broken :/

lex148:
@AethanFoot sorry I was traveling all last week. I will take a look at this one when I get back to the office Monday

lex148:
@AethanFoot I have been testing and this looks ready to go. If you resolve the conflict I will merge

PR #319, opened 2021-05-06
Force border opacity to 100%

### LINKED ISSUE #313: Some programs have transparent borders on light wallpapers
Hi,
I wanted to create a theme with a light wallpaper and I'm struggling with the issue that I can't see which `alacritty` is currently focused. It looks like some strange behavior is applied to borders, like some sort of opacity.

Take a look at this 'Dracula Rounded' theme with 2 alacritty instances opened:
image

And with wallpaper set to something light:
image
The focus remain unchanged.
I've created a wallpaper using light gray from top and dark gray from bottom, here is the result:
image

The strange thing is that most programs are not affected. Here is a gif presenting this:
Top row - affected: `alacritty Thunar Zathura`
Bottom row - not affected:  `gvim pcmanfm gimp` 
borders

Screen of terminal emulators as most of them are affected, the left one is focused (border should be solid green). It looks like only Xterm is handled correctly.
image

AethanFoot:
As mentioned in #313. Should also close #313.
Thanks.

### DIFF
--- a/src/display_servers/xlib_display_server/xwrap.rs
+++ b/src/display_servers/xlib_display_server/xwrap.rs
@@ -79,7 +79,7 @@ impl Default for XWrap {
 impl XWrap {
     /// # Panics
     ///
-    /// Can panic if unable to contact xorg.  
+    /// Can panic if unable to contact xorg.
     #[must_use]
     pub fn new() -> XWrap {
         const SERVER: mio::Token = mio::Token(0);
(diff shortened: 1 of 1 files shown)


VuiMuich:
Can confirm fix. You are awesome Aethan! :rocket:

Zwo1in:
I also confirm that this fixes, thanks

PR #321, opened 2021-05-06
Create CONTRIBUTING.md

AethanFoot:
A basic guide for contributing as more and more people are wanting to contribute. As a lot of these users are new to rust, should we give links to useful guides and documentation to rust?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,3 +53,41 @@ Similar to documentation rustfmt along with clippy are used for general style gu
 cargo clippy --release
 ```
 Again these are checked with CI, but it is always easier to check them before creating a pull request.
+
+### Setting up your Envriroment
+
+This section has some tips and tricks provided by contributors to help you get on your way in contributing.
+
+Tip 1. To make your life easier once your done coding, adding this function
+```
+checkleft() {
+        if cargo fmt ; then
+          if cargo test; then
+            if cargo clippy --all-targets --all-features ; then
+                if git add -A $1 ; then
+                    if git status ; then
+                        git commit -S -m "$2"
+                    else
+                        echo "Status failed"
+                     fi
+                 else
+                     echo "Add failed"
+                fi
+            else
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,55 @@
...
+```
+Or many text editors/IDE's have a supported plugin that will auto-format your code, e.g. for vim there is [rust.vim](https://github.com/rust-lang/rust.vim).
+
+### Style
+
+Similar to documentation rustfmt along with clippy are used for general style guidance. Clippy can be run using:
+```
+cargo clippy --release
+```
+Again these are checked with CI, but it is always easier to check them before creating a pull request.
```
mautamu, line 55: Could maybe add:
"""
To make it easier for yourself, you could add a function in your `~/.bashrc` called checkleft() that will do all of this for you (assumes you have a gpg-key, remove `-S` flag from `git commit` if not):
```bash
checkleft() {
        if cargo fmt ; then
          if cargo test; then
            if cargo clippy --all-targets --all-features ; then
                if git add -A $1 ; then
                    if git status ; then
                        git commit -S -m "$2"
                    else
                        echo "Status failed"
                     fi
                 else
                     echo "Add failed"
                fi
            else
                echo "Clippy failed"
            fi
          else
            echo "Test failed"
          fi
         else
                echo "FMT failed"
          fi
}
```

Once you add it, refresh bash (reopen your terminal) and then you can type:
```bash
# Example, adds all of src/, Cargo.toml, and Cargo.lock to be committed.
# You can pass anything that would go to git add -A 
checkleft 'src/ Cargo.toml Cargo.lock' 'Commit message'
```
"""

mautamu (commented):
This sounds like a good thing to add, good job!

CONTRIBUTING.md
lex148: I would like to move developer level tips and tricks to the wiki and link to it. I don't want to overload the main page

PR #292, opened 2021-04-14
[CONFIG BREAKING DRAFT] Begin implementing #201, #260, #270, #329

mautamu:
Howdy,

This PR can serve as a tracking issue for #201, #260, #270, and related configuration issues. As mentioned in #270, we will likely want to increment LeftWM to 0.3.0, as the changes posed in the linked issues will likely require some changes to theme and user configurations. While 15edd1a does not require such changes, once fully implemented those changes will change the overall flow of LeftWM such that changes are warranted. Here is the new theme.toml specification, with slightly modifications from #201 

```toml
border_width = 4 # define these up here for backwards compatibility with leftwm < 0.2.7
margin = [10, 10]
default_border_color = "#1E272B"
floating_border_color = "#1E272B"
focused_border_color = "#78824B"

name = "mytheme" # The name of the theme, MUST be set
description = "A basic theme" # A basic description, such as "A brown and orange theme inspired by sunflowers"
repository = "https://github.com/mautamu/theme" # The git repository for the theme. 
# Optional but requires commit if included. 
#Commit may be "*", but "*" may be rejected by maintainers for security purposes.
commit = "*" # pin at a specific version as a "release"
version = "1.0.0" # SEMVER-compatible version for the THEME
leftwm_versions = ">2.6.7" # SEMVER-compatible leftwm versions
dependencies = ["polybar", "dunst"]  # Used to check if required packages are installed in the system, internally calls is_in_path(dependency[i]) 

# task.program_name.up sets up the program. The following section will spawn `polybar mybar`
[task.polybar.up]
command = "polybar" # if command is not set, will default to program_name
# NOTE: do not include anything other than /path/to/program_name or program_name
# Everything else MUST go in arguments or else your task will not be called.
args = ["mybar"]

# defaults to pkill program name if no down script is provided.
[task.polybar.down]
command = "killall"
args = ["-q", "polybar"]

# if program_name in task.program_name.{up,down} does not match any dependency listed, a command value must be provided.
# paths are relative to the current directory of theme.toml file.
[task.notification.up]
command = "/usr/bin/dunst"
args = ["--config", "configs/dunstrc"]

[task.polybar.new_window]
packages = {pacman = ["polybar"], apt =["polybar"], dnf = ["polybar"] }

[task.polybar.install]
packages = {pacman = ["polybar"], apt =["polybar"], dnf = ["polybar"] }

[task.action_name.post_install]
command = "killall"
args = ["-q", "polybar"]

[task.action_name.pre_install]
[...]

PR #329, opened 2021-05-10
Add margins and gutters to workspace

### LINKED ISSUE #194: Allow user greater control of margin

### LINKED ISSUE #270: Opinions requested: `inner margins`
During testing for #259 I once again noticed that we have the margins between windows adding up.

I believe a more correct behaviour would be, if it is just the single margin between windows, but wanted to get your opinions, what you would desire to be correct.

And if we want to change this, would it be preferred to be hard-coded or configurable?

### LINKED ISSUE #327: EWW bar compatability
When using EWW to draw a bar, windows do not leave a gap at the top to compensate, and EWW just draws over windows. Is there a way to create a space at the top of the workspace for EWW to exist? Using the `margin` in the `theme.toml` file does not work since even using the vector format causes every window to have a large gap at the top, instead of the whole workspace just having the gap.

I've included a screenshot of what happens now.

image

AethanFoot:
Adds margins to workspaces, this closes #327 as the user could now add a gap their selves at the top of the workspace if the automatic system doesnt work. This also closes #194 and closes #270, as if you add half the margin you want to workspace and the other half to window, the gap will be consistent. E.g. if you want a gap of 8 consistently, if you set window_margin to 4 and workspace_margin to 4, all the gaps will be 8.
The reason the theme config breaks is because margin has changed to window_margin to be more clear, but this isn't necessary if we don't want to break the config.
Thanks.

### DIFF
--- a/src/config/theme_setting.rs
+++ b/src/config/theme_setting.rs
@@ -11,6 +11,7 @@ pub struct ThemeSetting {
     pub border_width: i32,
     pub window_margin: Margins,
     pub workspace_margin: Margins,
+    pub gutter: i32,
     pub default_border_color: String,
     pub floating_border_color: String,
     pub focused_border_color: String,
@@ -37,6 +38,7 @@ impl Default for ThemeSetting {
             border_width: 1,
             window_margin: Margins::Int(10),
             workspace_margin: Margins::Int(10),
+            gutter: 0,
             default_border_color: "#000000".to_owned(),
             floating_border_color: "#000000".to_owned(),
             focused_border_color: "#FF0000".to_owned(),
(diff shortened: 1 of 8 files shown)


AethanFoot:
Added a gutter to actually provide a fix for #327.
Thanks

AethanFoot:
Ye i'd say if there is a need for these changes now, I can change window_margin back to margin so theme.toml doesn't break. Then change the name with the other breaking changes.

mautamu:
Will add that to the pile in #292

lex148:
Sounds great. @AethanFoot let me know when you have made the change I will give it a merge

AethanFoot:
Switched window_margin back to margin.

PR #333, opened 2021-05-14
`leftwm-state` option to use template string instead of template file

VuiMuich:
Add option for leftwm-state to use template liquid with a string literal instead of a `template.liquid` file.
This allows for very simple output for testing and scripting without the need to use a tempalte-file.

For now I did just a lazy initial implementation to get some opinions, without handling the strings nicely with quotes.
So for now only very rudimentary use like
```sh
leftwm-state -s {{ window_title }} -w 0
leftwm-state -w 0 -s {{ workspace.tags }} -q
```
is possible. But if the idea is welcome, I'll dive into supporting full inline liquid strings

### DIFF
--- a/src/bin/leftwm-state.rs
+++ b/src/bin/leftwm-state.rs
@@ -22,6 +22,14 @@ async fn main() -> Result<()> {
                 .takes_value(true),
         )
         .arg(
+            Arg::with_name("string")
+                .short("s")
+                .long("string")
+                .value_name("STRING")
+                .help("Use a liquid template string literal to use for the output")
+                .takes_value(true),
+        )
+        .arg(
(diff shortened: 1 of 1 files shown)


VuiMuich:
Oops, is the 'remove file' option on the files changed actually removing files from the tree? I thought it wuld just remove them from the diff..
I guess I butchered this PR..

VuiMuich:
Uhm, I guess my more complex tests I did were not breaking from missing quotes on the string, which appear to get parsed away correctly anyways.
Instead it was only incorrect `liquid` syntax. We probably could need quite a bit more documentation on this feature. Will put this on my todo..

But I believe this makes this PR 'review ready'.

Issue #351, opened 2021-05-21
Not able to get window swallowing working


steveonlinux:
Has anyone been able to work in swallowing functionality into this wm? I've tried using devour and other wm agnostic solutions that seem to be broken here.

VuiMuich:
I am not sure if this wouldn't need at least #165 to be implemented.

AethanFoot:
This is because devour uses XUnmapWindow which is not a send_event, and currently we don't handle these events. 
Heres a fix to devour using XWithdrawWindow (which is a send_event):
```C
/*
 * devour
 * X11 window swallower
 */

#include <X11/Xlib.h>
#include <stdlib.h>
#include <string.h>

#define UNSAFE_CHARS "`\"'()[]& "

void run_command(char **argv) {
  char arg_char;
  char *arg;
  char cmd[1024] = {0};

  while ((arg = *++argv)) {
    while ((arg_char = *arg++)) {
      if (strchr(UNSAFE_CHARS, arg_char))
        strcat(cmd, "\\");
      strncat(cmd, &arg_char, 1);
    }
    strcat(cmd, " ");
  }
  system(cmd);
}

int main(int argc, char **argv) {
  int rev;
  Window win;
  Display *dis = XOpenDisplay(0);
  int snum = DefaultScreen(dis);

  XGetInputFocus(dis, &win, &rev);
  XWithdrawWindow(dis, win, snum);
  XFlush(dis);
  run_command(argv);
  XMapWindow(dis, win);
  XCloseDisplay(dis);

  (void)argc;
  return 0;
}
```

steveonlinux:
> This is because devour uses XUnmapWindow which is not a send_event, and currently we don't handle these events.
> Heres a fix to devour using XWithdrawWindow (which is a send_event):
> 
> ```c
> /*
>  * devour
>  * X11 window swallower
>  */
> 
> #include <X11/Xlib.h>
> #include <stdlib.h>
> #include <string.h>
> 
> #define UNSAFE_CHARS "`\"'()[]& "
> 
> void run_command(char **argv) {
>   char arg_char;
>   char *arg;
>   char cmd[1024] = {0};
> 
>   while ((arg = *++argv)) {
>     while ((arg_char = *arg++)) {
>       if (strchr(UNSAFE_CHARS, arg_char))
>         strcat(cmd, "\\");
>       strncat(cmd, &arg_char, 1);
>     }
>     strcat(cmd, " ");
>   }
>   system(cmd);
[...]

PR #361, opened 2021-05-27
Hide bars when entering fullscreen

### LINKED ISSUE #360: Xmobar does not hide in fullscreen
Hi,
I use terminator with transparency set to 80%. If I press F11 to maximize the terminal I still can see Xmobar in the background

2021-05-26-201436_1024x600_scrot

Regards
Markus

AethanFoot:
This allows us to hide bars aswell when entering fullscreen and closes #360.
Another issue I would like to ask about comes from not handling non send event unmap events (if that makes sense), related #351. For the devour script there was an easy fix, but for larger projects, such as polybar, which use XUnmapWindow to hide their window can't easily be modified to work. Is there a reason not to handle these events, and if there is should we partially handle them, as for the example of polybar when telling it to hide it will disappear (nothing updates as we don't handle it) and then won't reappear when unhiding as the manager already knows about it.
Many Thanks.

### DIFF
--- a/src/models/window.rs
+++ b/src/models/window.rs
@@ -91,7 +91,7 @@ impl Window {
     #[must_use]
     pub fn visible(&self) -> bool {
         self.visible
-            || self.type_ == WindowType::Dock
+            // || self.type_ == WindowType::Dock
             || self.type_ == WindowType::Menu
             || self.type_ == WindowType::Splash
             || self.type_ == WindowType::Dialog
--- a/src/utils/window_updater.rs
+++ b/src/utils/window_updater.rs
@@ -1,30 +1,50 @@
 use crate::models::Manager;
-use crate::models::Window;
 
 /*
  * step over all the windows for each workspace and updates all the things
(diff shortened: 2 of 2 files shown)


lex148:
So close with this one. I have one small issue while testing.

I have a three monitor setup
```
|----|          |----|
|    ||--------||    |
| 1  ||    3   ||  2 |
|----||--------||----|
```

if I full screen a window on monitor 3 the bar on monitor 3 AND 2 are hidden.

AethanFoot:
Hmmm, strange. The only way I think that could happen is if the strut x,y for monitor 2s bar is wrong, or there is an inclusion error with contains point. I will do some more testing tomorrow, if it is an inclusion issue I can check if the midpoint of the bar is within the workspace instead.
Thanks.

AethanFoot:
@lex148 can you check whether this helped. 
Thanks.

lex148:
> @lex148 can you check whether this helped.
> Thanks.

works great!  awesome work

PR #363, opened 2021-05-29
Global up script

### LINKED ISSUE #362: [Feature] Global up script
Currently the only way to do setup or run custom programs in the background is to use the `~/.config/leftwm/theme/current/up` script. This is really inconvenient if you want to be able to switch between themes - you have to edit each theme's `up` script to do what you need.
I propose a global up script (`~/.config/leftwm/up`) which can remain the same while you switch between themes.

mfdorst:
This implements the change proposed in #362.

Closes #362.

### DIFF
--- a/src/bin/leftwm-worker.rs
+++ b/src/bin/leftwm-worker.rs
@@ -165,8 +165,15 @@ async fn event_loop(
             }
         }
 
-        //after the very first loop boot the theme. we need the unix socket to already exist
+        //after the very first loop run the 'up' scripts (global and theme). we need the unix
+        //socket to already exist.
         after_first_loop.call_once(|| {
+            match Nanny::new().run_global_up_script() {
+                Ok(child) => {
+                    child.map(|child| manager.children.insert(child));
+                }
+                Err(err) => log::error!("Global up script faild: {}", err),
+            }
             match Nanny::new().boot_current_theme() {
                 Ok(child) => {
                     child.map(|child| manager.children.insert(child));
--- a/src/bin/leftwm-worker.rs
+++ b/src/bin/leftwm-worker.rs
@@ -175,13 +175,13 @@ async fn event_loop(
         //after the very first loop run the 'up' scripts (global and theme). we need the unix
         //socket to already exist.
         after_first_loop.call_once(|| {
-            match Nanny::new().run_global_up_script() {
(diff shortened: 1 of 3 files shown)


mautamu:
Tagging #292 as this would need to be considered therein.

lex148:
So I don't have a problem with this if people want this feature, but why not just put what you want in `.config/autostart`

LeftWM respects the behavior of the autostart folder and boots everything in it only once at boot time.

AethanFoot:
I feel it has its place, this is a common point I see often brought up (mostly as they dont know about autostart directory). However I feel for newer users it may be easier to setup through the wm, instead of having to additionally search how to use the autostart directory or .xprofile, etc.
If we feel this isnt entirely needed then we should add a bit to the wiki on these various methods of autostarting which are wm independent

lex148:
I have no problem with adding this as well. I would just encourage people to use a more standard way if they can.

VuiMuich:
Maybe we should add a 'getting started' section in the wiki where we help with common pitfalls when getting new into leftwm.

I'll propose something in the next days.

mfdorst:
I looked into autostart but what I found seemed to indicate that it was part of Gnome. I see now on further inspection that it's not. Perhaps we could add a link to the autostart documentation somewhere in the README? I found this link ([arch wiki](https://wiki.archlinux.org/title/XDG_Autostart)), idk if there's a better resource for it.

mfdorst:
Just a final two cents, I think that this has its own place even for people who are aware of and use `autostart`, because there may be things people want to run _only_ for `leftwm`, but not when using some other WM or DE. This will serve in that case.

rigoletto-freebsd:
I think a middle ground and simpler solution would be something like i3: [exec and exec_always](https://i3wm.org/docs/userguide.html#_automatically_starting_applications_on_i3_startup). :-)

mautamu:
> I think a middle ground and simpler solution would be something like i3: [exec and exec_always](https://i3wm.org/docs/userguide.html#_automatically_starting_applications_on_i3_startup). :-)

I'm trying to think of the best way to implement this. Perhaps one alternative is with #292 to have `Task`s in config.toml as well as theme.toml and add hooks for .up (theme), .down (theme), .start (leftwm), .end (leftwm), and .install (leftwm-theme).

PR #367, opened 2021-05-30
Add subcommands to leftwm

mfdorst:
I'm not sure why, but I find it really irritating to type `leftwm-theme` rather than `leftwm theme`. So I implemented four subcommands for `leftwm`:
+ `leftwm check`
+ `leftwm command`
+ `leftwm state`
+ `leftwm theme`
If you run any of these, it will simply run the corresponding program e.g. `leftwm-theme` rather than attempting to start the window manager. If you call it with no arguments, it will function as normal and attempt to start the window manager.

Further arguments will be passed to the corresponding program, so `leftwm theme update` will run `leftwm-theme update`.

### DIFF
--- a/src/bin/leftwm.rs
+++ b/src/bin/leftwm.rs
@@ -109,6 +109,10 @@ fn execute_subcommand(args: Vec<String>) {
 ///
 /// If `--help` or `--version` flags are not passed, this will do nothing.
 ///
+/// # Arguments
+///
+/// + `args` - The command line arguments leftwm was called with.
+///
 /// # Exits
 ///
 /// Exits early if `--help` or `--version` flags are passed.
--- a/src/bin/leftwm.rs
+++ b/src/bin/leftwm.rs
@@ -1,6 +1,6 @@
 use leftwm::child_process::{self, Nanny};
 use std::env;
-use std::process::Command;
+use std::process::{exit, Command};
 use std::sync::{
     atomic::{AtomicBool, Ordering},
     Arc,
@@ -68,7 +68,8 @@ fn main() {
 ///
 /// If a valid subcommand is supplied, executes that subcommand, passing `args` to the program.
 /// Valid subcommands are `check`, `command`, `state` and `theme`.
-/// Prints an error to STDERR and returns `false` if an invalid subcommand is supplied.
+/// Prints an error to STDERR and exits non-zero if an invalid subcommand is supplied, or there is
+/// some error while executing the subprocess.
 ///
 /// # Arguments
 ///
@@ -77,15 +78,23 @@ fn main() {
 /// # Panics
 ///
 /// Panics if `args` has length < 2.
-fn execute_subcommand(args: Vec<String>) -> bool {
-    if ["check", "command", "state", "theme"]
-        .iter()
-        .any(|x| x == &args[1])
-    {
-        println!("Running leftwm-{} with args {:?}", &args[1], &args[2..]);
-        true
+///
+/// # Exits
+///
+/// Exits 1 if the first argument is not a valid subcommand.
+/// Exits 2 if the first argument is a valid subcommand, but the associated program failed to run.
+fn execute_subcommand(args: Vec<String>) {
+    let subcommands = ["check", "command", "state", "theme"];
+    // If the second argument is a valid subcommand
+    if subcommands.iter().any(|x| x == &args[1]) {
+        // Run the command
+        let cmd = format!("leftwm-{}", &args[1]);
+        if let Err(e) = Command::new(&cmd).args(&args[2..]).spawn() {
+            eprintln!("Failed to execute {}. {}", cmd, e);
+            exit(2);
+        }
     } else {
         eprintln!("Invalid command '{}'.", &args[1]);
-        false
+        exit(1);
     }
 }
--- a/src/bin/leftwm.rs
+++ b/src/bin/leftwm.rs
@@ -112,7 +112,7 @@ fn execute_subcommand(args: Vec<String>) {
 ///
 /// Exits early if --help or --version flags are passed.
 fn handle_help_or_version_flags(args: &[String]) {
-    App::new("leftwm")
+    App::new("LeftWM")
         .author("Lex Childs <lex.childs@gmail.com>")
         .about("A window manager for adventurers.")
         .long_about(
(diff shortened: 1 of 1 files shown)


mfdorst:
Note: as written this does not allow you to run arbitrary `leftwm-<something>`s the way that git and cargo do. It could easily be made to do that, but I wasn't sure if that was desirable or not. I think perhaps it is, but I'll wait to hear some feedback before changing it.

mfdorst:
Also note: this does not use `clap` - it processes the arguments manually. I started with `clap`, but found that this was much simpler since we're only looking at the first argument and then forwarding the rest. The downside is that you don't get the nice usage messages that `clap` provides.

VuiMuich:
I like this. I remember when getting started with leftwm I got puzzled a bunch of times, that I couldn't use any subcommands or flags with the `legtwm` command itself. 

So while you are already at this, would you mind to add a `help` and `version` subcomands, which also can be executed with `leftwm -h` and `leftwm -v`?
These would feel very natural and might help newcomers quite a bunch.

mfdorst:
@VuiMuich As I mentioned, I'm handling arguments manually here. I'm not sure what the best way is to add things like `-h`/`--help` and `-v`/`--version`. I'd much rather use `clap` for that rather than write the code manually myself, but I'm not sure if I can do what I'm doing and still delegate the help and version menus to `clap`.
For now I'm going to leave this as is, since I don't know the best way forward. Please let me know if you have a suggestion.

mfdorst:
To be clear I could add those menus, but I'd rather avoid writing the code manually if I can, because it will be tedious and also it will need to be maintained as more options become available, which sounds irritating.

mfdorst:
Actually I take back what I just said, I think I have an idea for this, but it may have to wait for tomorrow.

src/bin/leftwm.rs
```diff
@@ -51,3 +64,38 @@ fn main() {
...
+/// Panics if `args` has length < 2.
+///
+/// # Exits
+///
+/// Exits 1 if the first argument is not a valid subcommand.
+/// Exits 2 if the first argument is a valid subcommand, but the associated program failed to run.
+fn execute_subcommand(args: Vec<String>) {
+    let subcommands = ["check", "command", "state", "theme"];
+    // If the second argument is a valid subcommand
+    if subcommands.iter().any(|x| x == &args[1]) {
```
mautamu, line 90: Might want to do lower case check in case someone types for example leftwm Command? Not sure what proper shell behaviour is.

VuiMuich:
I believe you could transform your code into a `subcommand_handler` and then match over `args[1]` to call this or print help or version info

src/bin/leftwm.rs
mfdorst: I think if someone types `leftwm Command` it will print `Invalid command 'Command'`. I think that is acceptable behavior.

mfdorst: If I type `leftwm-Command` I get `zsh: command not found: leftwm-Command` which makes perfect sense.

mfdorst:
> I believe you could transform your code into a `subcommand_handler` and then match over `args[1]` to call this or print help or version info

I don't see any reference to `subcommand_handler` in the `clap` docs.

My idea for now is that I write out the `clap::App` the way you normally would, but just don't call `get_matches_for()` unless the first argument is one of `-h`, `--help`, `-v` or `--version`. So basically you're just treating `clap` as a way to generate the help and version text. It's a _little_ bit manual, but I think it's better than writing out the help text by hand.

VuiMuich:
What I meant was, that you refactor manual handling into a `subcommand_handler` (you might want to have a look at the `template_handler` in `leftwm-state.rs`) and have a `match` calling either the `subcommand_handler` or 'print help' or 'print version'.

A bit hard to provide a snippet as I am only on mobile today.

mfdorst:
Okay, that worked out pretty well. Here's the current help text:
```
leftwm 0.2.8
Lex Childs <lex.childs@gmail.com>
Starts LeftWM if no arguments are supplied. If a subcommand is given, executes the the corresponding leftwm program,
e.g. 'leftwm theme' will execute 'leftwm-theme', if it is installed.

USAGE:
    leftwm [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -v, --version    Prints version information

SUBCOMMANDS:
    check      Check syntax of the configuration file
    command    Send external commands to LeftWM
    state      Print the current state of LeftWM
    theme      Manage LeftWM themes
```

mfdorst:
> What I meant was, that you refactor manual handling into a `subcommand_handler` (you might want to have a look at the `template_handler` in `leftwm-state.rs`) and have a `match` calling either the `subcommand_handler` or 'print help' or 'print version'.

Ah okay. Well, check out what I've written and let me know if you think it would be better to do it manually. Personally I like that `clap` colorizes the output, but that's a minor thing.

src/bin/leftwm.rs
mautamu: Fair enough. That makes sense, was more so interested in how others handle sub commands, but they do the same.

VuiMuich:
Wow this looks nice 👍🏼 
Still just on mobile so I can't actually test but I'm very much looking forward to!

VuiMuich:
Sorry to report this, but for me `leftwm state` is not working properly.
When I use it without further arguments, it somewhat gets semi-disowned: it returns to the promt, but keeps printing its output. It cannot be cancelled via CTRL-C, but the procces has to be killed (e.g via htop)
When providing arguments (even a simple `-q`) returns 
```
error: Found argument '-q' which wasn't expected, or isn't valid in this context
```
Is this the subcommand not passing or parsing the arguments correctly, or is it that clap get confused by the subcommand using arguments?

mfdorst:
@VuiMuich I'm working on a fix. It's because I didn't wait for the process to finish, so it continues to run in the background after `leftwm` quits.

Regarding the arguments, I'm not sure why it's not just forwarding them through. I'll have to figure out how to get clap to accept arbitrary arguments and not try to validate them if they come after a subcommand. Alternately I could just skip the clap part if there is at least one argument, but I like that clap gives you suggestions if you type something wrong, like if you type `leftwm status` it will recommend that you might have meant `leftwm state`, which is pretty cool.

VuiMuich:
Great, thanks!
I agree that the output of clap is really helpful and great. Very much in the spirit of rustc errors in a way..

PR #373, opened 2021-05-31
Be more efficient when updating bar tags

### LINKED ISSUE #366: Single window in full-screen leaves tag "empty"
**Describe the bug**
When a tag holds a single window and this window is put in full-screen it renders the tag (in `leftwm-state`) as empty (or `"busy":false`)

**To Reproduce**
It is easy to observe in any status bar but thie following command show this also very clear:
```shell
$ leftwm-state -w 0 -q -s '{% for tag in workspace.tags %}{{ tag.index}}: busy:{{ tag.busy }} {% endfor %}'
```
Output:
```shell
0: busy:true 1: busy:false 2: busy:false 3: busy:false 4: busy:false 5: busy:false 6: busy:false 7: busy:false 8: busy:false
```
**Expected behavior**
As the window is still just displayed on a specific tag it should keep this tag  as `"busy":true`

AethanFoot:
Rework of #361. I felt that the previous attempt unnecessarily updated the tags for the docks to often taking unneeded processor time (not matter how little). This now only updates the tags of the windows with a strut when a change occurs (the strut changes or we move to a different tag). I am not 100% we need the code in state, I guess it catches windows that don't get relaunched on reload.
Thanks.

Now fixes #366.

### DIFF
--- a/src/handlers/command_handler.rs
+++ b/src/handlers/command_handler.rs
@@ -214,32 +214,13 @@ fn swap_tags(manager: &mut Manager) -> Option<bool> {
     if manager.workspaces.len() >= 2 && manager.focus_manager.workspace_history.len() >= 2 {
         let hist_a = *manager.focus_manager.workspace_history.get(0)?;
         let hist_b = *manager.focus_manager.workspace_history.get(1)?;
-        //Update dock tags
-        let ws_a = manager.workspaces.get(hist_a)?;
-        let ws_b = manager.workspaces.get(hist_b)?;
-        let mut changed = vec![];
-        manager
-            .windows
-            .iter_mut()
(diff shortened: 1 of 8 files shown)


VuiMuich:
I like this.
The omitted code in `window_updater.rs` looks a bit like it was meant to solve a (possibly only a potential) corner case, maybe?

BTW: while this is almost related, could you check if this fixes #366 or maybe include a fix? I can't really test today, maybe tomorrow in the evening..

lex148:
awesome work, but I am seeing a little weirdness while testing. 

If you `mod+w` to swap windows around and then toggle full screen, The wrong bar is hidden. The workspace that previously had the window hides the bar not the workspace that now has the window.

AethanFoot:
This should now work with swap_tags.
 Thanks.

AethanFoot:
Ok this should be ready to merge.
Thanks!

PR #378, opened 2021-06-05
Fix subcommands continuing in background, fix clap errors showing for valid arguments to subcommands

mfdorst:
As pointed out by @VuiMuich [here], subcommands had some weird behaviours.

1. `leftwm state` would give you your prompt back immediately and then run in the background (vomiting all over your terminal whenever window manager actions occurred) rather than staying alive and allowing you to stop it with `Ctrl+C`. This was because we weren't `.wait()`ing for the command to finish before exiting.
2. `leftwm state -q` and other such commands were being considered invalid by `clap`. This is because clap does not know about the arguments to the external subcommands, so it thinks they are all invalid.

`1` was an easy fix, we just needed to `.wait()` on the subcommand.

`2` does not have an obvious fix - I was unable to find a way to get `clap` to accept arbitrary arguments to subcommands as valid. The workaround I came up with is to simply skip the `clap` part if a valid subcommand is passed.

While I was at it I made two changes:
1. List the subcommands in one place so that you don't need to update multiple functions to add a new subcommand.
2. Change the short `--version` flag from `-v` to `-V`. I personally prefer `-v` because it's more common, but since our other tools already use `-V` and one of them already has a `-v` option that does something else, it's better to be consistent.

[here]: https://github.com/leftwm/leftwm/pull/367#issuecomment-852329220

### DIFF
--- a/src/bin/leftwm.rs
+++ b/src/bin/leftwm.rs
@@ -5,6 +5,7 @@
 
 use clap::{crate_version, App, AppSettings, SubCommand};
 use leftwm::child_process::{self, Nanny};
+use std::collections::BTreeMap;
 use std::env;
 use std::process::{exit, Command};
 use std::sync::{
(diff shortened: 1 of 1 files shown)


mautamu:
>2. Change the short --version flag from -v to -V. I personally prefer -v because it's more common, but since our other tools already use -V and one of them already has a -v option that does something else, it's better to be consistent.

Is there a generally accepted default for verbosity that's not -v? If so could maybe use that.

mfdorst:
@mautamu not that I know of. Most programs I know of that have a `--verbose` flag use `-v`.

mfdorst:
@mautamu `cargo`'s `--version` flag is `-V`. I assume that's why `clap` defaults to `-V`.

mautamu:
Ah, that's where the heritage is from. Yeah, `cargo build -v` is for verbose, `cargo -V` is version.

Issue #384, opened 2021-06-10
leftwm-command help should list available command


lex148:
It would be really nice if running the help for leftwm-command list the commands that you could run

VuiMuich:
I already started a local branch on this, but if AethanRocket want's to beat me to sending a PR I'll be happy to trash my minuscule progress i had on this 😬

PR #385, opened 2021-06-10
`leftwm-command` add `list` option

VuiMuich:
I just had a spontaneous little time slot so this is just very clumsily printing a literal string with a list of the available external commands.
But so far I don't think we could auto-extract the list of external-commands somewhere, right? The closest to automating this would be to process the Command enum from command.rs but the wouldn't allow to add info about the arguments some commands take.
Probably the 'correctest' way would be to add doc-comments to each command which could be processed.

BTW: is there a special reason, why we let `Reload` and `MoveWindowToTop` run into a `{}`?

### Usage:
`leftwm-command -l` or `leftwm-command --list` will print a list of the available external commands.


VuiMuich:
Contrary to #384 I did not want to put the list directly in the help message because personaly I prefere if the first layer of help is kept tight but provides options to get mor information, so I put the list behind the `-l/--list` flag

lex148:
I agree with putting it behind the flag. the main thing is  that it is available.

PR #393, opened 2021-06-16
Fix #392

### LINKED ISSUE #392: Focus suddenly switch to master.
Hello,

I've noticed these days if the focus is not on the master window, after some time[1] it suddenly switch to the master window. The first few times I've noticed this behavior, I thought I may had did something without realizing but now I'm sure this is happening by its own, specially after it switched while I was writing an e-mail. ^^

FreeBSD, LeftWM: e55355851ef6afd40d94414347cacc01197ab2e3.

`focus_behaviour = "Driven"`

[1] I couldn't figure out exactly.

Thank you!

AethanFoot:
This fixes #392, however doesn't address further issues brought up yet.
Thanks.

### DIFF
--- a/src/bin/leftwm-worker.rs
+++ b/src/bin/leftwm-worker.rs
@@ -1,6 +1,6 @@
 use leftwm::{
     child_process::{self, Nanny},
-    models::{FocusManager, Tag},
+    models::{FocusBehaviour, FocusManager, Tag},
 };
 
 use crate::models::TagModel;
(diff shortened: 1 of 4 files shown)


AethanFoot:
When hiding a scratchpad the last focused window or the first window on the workspace will be focused. When bringing a scratchpad to tag it will be focused no matter the mode you are in.

Side note I have discovered a small lock when in ClickTo, which can be broken by pressing any key. To create this lock consistently I had to open/close a tab on brave when first clicking on it (to also give it focus which calls replay click) then next clicking on another window. The time between clicking the next window to give it focus didn't seem to matter but the problem happen more often the quicker you clicked on the next window after closing/opening a tab. The lock is that the other window won't get focus and can't interact with anything with the mouse, however as soon a you press a key on the keyboard everything updates to where you clicked or scrolled in that time.
I not really sure what is causing this, I could only get it to happen when either opening a tab or closing a tab on my browser.
Thanks.

PR #399, opened 2021-06-23
using hjkl instead of arrow keys

mjbn:
hey guys, i think using vim like keybindings (hjkl) as default keybindings is better cause thay are more easy to access than arrow keys and most of the users at least have some experince with vim.
and one more thing.
most of the wm users will set mod+Return as the key for launching terminal cause you need to launch terminal many times a day, it has less key and its easy to access.
if  we change this keybindings thats less problem for people to at least give a try to leftwm.
they dont have to Memorize new keybinding to be able to launch a terminal or even use the config.toml to change the keybindings maybe.
for example i use mod+1..9 to change focus between desktops, some wm use somethings else but with leftwm i dont have to change the keybindings.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -191,7 +191,7 @@ impl Default for Config {
             Keybind {
                 command: Command::Execute,
                 value: Some(default_terminal().to_owned()),
-                modifier: vec!["modkey".to_owned(), "Shift".to_owned()],
+                modifier: vec!["modkey".to_owned()],
                 key: "Return".to_owned(),
             },
             // Mod + Shift + q => kill focused window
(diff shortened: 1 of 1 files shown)


VuiMuich:
Welcome Adventurer and thank you for your Contribution Request!

Personally I also have vim-like keybinds for navigation and have tried to switch `launch terminal` with `swap with main` but noticed I am in fact swapping more often then I open new windows actually (using the layout `MainAndDeck` 90% of the time) so went back to the default.
I believe the defaults we currently have are quite sane:
Though chances are high that new users are familiar with vim bindings you can't be sure, so arrow-keys for navigation are just the safe fallback. Similarly having `Mod+Shift+Enter` the default for launch a terminal is as far as I can tell the more common default.

Getting used to do changes to `config.toml` is at some point almost a necessity for any leftwm adventurer (as in general is the case with any WM) and editing the keybinds is not too difficult either, so I don't think this is something we should fear would drive new users away too much.

This is just my opinion, though. Lets see what @lex148 thinks about this.

If this won't get merged I hope you'll find other opportunities to contribute to leftwm in the future!
glhv

Phate6660:
I for one (as a random user) would be against this.
I'm on the opposite side.
In fact, I use Vim but despise it's keybindings.
If LeftWM had Vim keybinding by default when I first found it, I probably would've passed it up completely.

mautamu:
I will admit that knowing to use “Down” and “Up” as the proper key names (instead of, for example, “down arrow” or “down”) is not as clear as it could be to new users. Where the current default at least gives some examples with “Down” and “Up,” this is not the case with this new default. 

If this is merged, I would suggest either adding another default with “Down” or perhaps considering a way to make non-standard keys like “Down” or “eacute” easier to find within the documentation. (Also tagging #292).

mjbn:
@mautamu yeah i think we should add some examples for non-standard keys in the docs, for example i wanted to change the dmenu keybinding to alt+space but i couldnt find anything about it in the docs.
Well, after some try i found out that i should write it as
Mod1 + Space (with capital letter)

VuiMuich:
Maybe we should replace the link at the end of the [keybind paragraph](https://github.com/leftwm/leftwm/wiki/Config#keybind) in the Wiki with a proper page instead of just linking to `xkeysym_lookup.rs`.
And change the layout a bit so it pops a bit more as relevant information.

lex148:
I like the Idea as I use vim everyday, but I really want to keep the barriers to entry as low as possible. 

What about this. It isn't very obvious that you can bind multiple key combinations to the same command. It would be good to have an example of this in the default config. Why not just add vim keys and keep arrows too. 

The one down side I see is we are claiming more key combos that other people might want for different command. then again they can always be changed. 

What does everyone think about this?

mjbn:
@lex148 it sounds good to me, if you want to use hjkl or arrow keys for something else you can delete the other one

VuiMuich:
> if you want to use hjkl or arrow keys for something else you can delete the other one

Or just change the `command` value

lex148:
Sounds like we have a plan, @mjbn just give me a ping and I will merge when you have updated the PR

PR #414, opened 2021-07-04
Fix #413

### LINKED ISSUE #413: No focus when a window is moved to an empty tag.
At least when using the `Driven` focus behavior, when a window is moved to an empty tag it will not be focused, and can't be focused unless creating a new window.

1. move a window to an empty tag
2. switch to that tag
3. the windows is not focused (the focus is on the previous tag)
4. create a new window as the only way to have focus on this tag.

 - OS/Distro: FreeBSD 13R
 - Output of `leftwm-check`
 
`:: LeftWM version: 0.2.8
:: LeftWM git hash: 10-eol-90394-geb4f5f9ed525-modified
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK (has XDG_RUNTIME_DIR)`

PS. the unfocused problem issue when LeftWM is reloaded is back.

AethanFoot:
Fixes #413, however the is a weird bug which I have no idea what causes it. For some reason when closing windows the focus moves to the second window on the tag, no matter which window you close (even though it has the same code as to move the window to a different tag). Will have another look tomorrow but if anyone can spot the issue before then it would be much appreciated.
Thanks!

### DIFF
--- a/src/display_servers/xlib_display_server/xwrap.rs
+++ b/src/display_servers/xlib_display_server/xwrap.rs
@@ -778,19 +778,21 @@ impl XWrap {
     // TODO: Verify that Error is unreachable or specify conditions that may result
     // in an error.
     pub fn move_cursor_to_point(&self, point: (i32, i32)) -> Result<(), XlibError> {
-        let none: c_int = 0;
-        unsafe {
-            (self.xlib.XWarpPointer)(
-                self.display,
-                none as c_ulong,
-                self.get_default_root(),
-                none,
-                none,
-                none as u32,
-                none as u32,
-                point.0,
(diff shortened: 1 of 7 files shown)


VuiMuich:
I believe when closing the `main` window the focus moves to the new main, so at least that part seams to work correctly.

AethanFoot:
Moved the repeated code to a function in window_handler, also the "bug" was me putting for_active_workspace somewhere is_handle was suppose to be 😅.
Thanks.

VuiMuich:
great!
LGTM

PR #421, opened 2021-07-25
focus_handler: cheaper `clone` operation in `move_focus_to_point`

vrmiguel:
We can clone just `WindowHandle` instead of the whole `Window`


lex148:
Awesome work! thanks for the PR

PR #428, opened 2021-08-04
Unfocus last window when switching to an empty tag

Syudagye:
Hey,

This pull request fixes a problem i had.
When switching to an empty tag from a tag where a window had focus, the window on the previous tag would still have the focus.
This can be very annoying since you could still perform actions on this window while it was not visible.

I fixed this issue by telling LeftWM to focus an imaginary window with an arbitrary handle value, which seems to work fine for now.
I had to adapt some code for this to work on the first focus of a tag.

### DIFF
--- a/src/display_servers/xlib_display_server/xwrap.rs
+++ b/src/display_servers/xlib_display_server/xwrap.rs
@@ -23,6 +23,7 @@ use crate::models::XyhwChange;
 use crate::utils::xkeysym_lookup::ModMask;
 use crate::DisplayEvent;
 use crate::{config::ThemeSetting, models::FocusBehaviour};
+use std::alloc::handle_alloc_error;
 use std::ffi::CString;
 use std::os::raw::{c_char, c_int, c_long, c_uchar, c_uint, c_ulong};
 use std::ptr;
@@ -1253,18 +1254,25 @@ impl XWrap {
     }
 
     pub fn unfocus(&self) {
-        let windows = self.get_windows_for_root(self.get_default_root());
-        if let Ok(windows) = windows {
-            let dock = windows.iter().find(|w| {
-                self.get_atom_prop_value(**w, self.atoms.NetWMWindowType)
-                    == Some(self.atoms.NetWMWindowTypeDock)
-            });
-            if let Some(dock) = dock {
-                self.window_take_focus(&Window::new(WindowHandle::XlibHandle(*dock), None, None));
-                return;
-            }
+        let handle = self.get_default_root();
+        unsafe {
+            (self.xlib.XSetInputFocus)(
+                self.display,
+                handle,
+                xlib::RevertToNone,
+                xlib::CurrentTime,
+            );
+            (self.xlib.XChangeProperty)(
+                self.display,
+                self.get_default_root(),
+                self.atoms.NetActiveWindow,
+                xlib::XA_WINDOW,
+                32,
+                xlib::PropModeReplace,
+                vec![c_ulong::MAX].as_ptr().cast::<u8>(),
+                1,
+            );
         }
-        self.window_take_focus(&Window::new(self.get_default_root_handle(), None, None));
     }
 
     pub fn kill_window(&self, h: &WindowHandle) {
--- a/src/display_servers/xlib_display_server/xwrap.rs
+++ b/src/display_servers/xlib_display_server/xwrap.rs
@@ -23,7 +23,6 @@ use crate::models::XyhwChange;
 use crate::utils::xkeysym_lookup::ModMask;
 use crate::DisplayEvent;
 use crate::{config::ThemeSetting, models::FocusBehaviour};
-use std::alloc::handle_alloc_error;
 use std::ffi::CString;
 use std::os::raw::{c_char, c_int, c_long, c_uchar, c_uint, c_ulong};
 use std::ptr;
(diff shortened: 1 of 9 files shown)


AethanFoot:
Looks good, one note would be to change u64::MAX to c_ulong::MAX (to help 32 bit compatability). In terms of unfocusing we could do this on the xlib side, whether there is a built in binding to do so or we give focus to the root window on the screen.
Thanks.

Syudagye:
I experimented a bit with focusing the root window, and it it works, i can unfocus the last window correctly. But it somehow breaks my polybar, and after some time, i can't change tag anymore. So i don't think it's the right way to go (or I just didn't do it properly).
Also i could not found any binding to directly unfocus in xlib

lex148:
Syudagye, This is great thanks for the PR. You are awesome for taking on bugs like this!

I am having a bit of trouble testing, and am wondering if clearing focus is working differently across versions of XOrg. @AethanFoot @VuiMuich is clearing focus working for either of you?

VuiMuich:
I'll be happy to give it a try. Although I have to admit I'm a bit occupied IRL, but will have a look at it with the next chance I'm back on my machine.

VuiMuich:
Well, the opportunity came quicker then expected.
Unfortunatelly though, it doesn't work at all on my build with:
```
leftwm check
:: LeftWM version: 0.2.9
:: LeftWM git hash: 6da68cf
```
My test:
- go to `Tag 5`
- open a terminal window
- open `htop`
- quit `htop`
- go to `Tag 6`
- press `Up-Arrow` and `Enter`
- go back to `Tag 5` and see `htop` running

@Syudagye: could you please push your attempt with focusing on the root window? I really would like to try this as well as I believe as well this is the cleaner and 'more correct' way of doing it.
Regarding you issues with polybar: is its window type "dock" by default? Can't remember, as I moved away from it since a couple of months.

Maybe focussing on the root window could be a 'sane fallback' and the first attempt would be to find a window of type `dock` to focus at?

AethanFoot:
Now I'm back at my PC I have found and seems to work using an external script, is to use XSetInputFocus. Set the display as normal, set the window to focus to 0 (same as None in C), set the revert to RevetToNone, and time to CurrentTime. This should allow you to construct a function which unfocuses all windows. After this we would either need to act on the FocusIn/FocusOut events or, the easier option, make sure we update the window history with None so that if we try to grab the focused window we aren't given the window we just unfocused (This is probably an issue with the current implementation).

Heres the main chunk of the script (written in C):
```
int main(int argc, char **argv) {
  Display *dis = XOpenDisplay(0);

  XSetInputFocus(dis, 0, RevertToNone, CurrentTime);
  XFlush(dis);
  XCloseDisplay(dis);

  (void)argc;
  return 0;
}
```
If something didn't make sense or anything else let me know.
Thanks.

Syudagye:
After a bit of testing it turns out it was not working for me with the current method (I don't know how i thought it was), but actually works with the root window method. 
@VuiMuich [here is my implentation of this   ##method.](https://github.com/Syudagye/leftwm/tree/unfocus_when_switching_to_empty_tag-root_window_method)
Unfortunately, as i said, it seems to freeze certain parts of my polybar (the `internal/xworkspaces` section, displaying tags, and the window title section, which pulls infos from `leftwm-state` ([here](https://github.com/Syudagye/changed-sunset/blob/master/polybar.ini) is the config for it if you want)), so i'll need to investigate on this problem.

@AethanFoot I tried your method [here](https://github.com/Syudagye/leftwm/tree/unfocus_when_switching_to_empty_tag-handle_zero_method) by chaging the handle from `c_ulong::MAX` to `0`, but it leads to wierd behaviours: when swithcing between tag where there is a window everything is fine, but when the tag is empty i can use any of my keybindings.

> Maybe focussing on the root window could be a 'sane fallback' and the first attempt would be to find a window of type dock to focus at?

Thank you for this suggestion, i'll try this now.

> [...] the easier option, make sure we update the window history with None so that if we try to grab the focused window we aren't given the window we just unfocused (This is probably an issue with the current implementation).

This sound like a cleaner way of dealing with window history, i'll work in this, thanks !

Syudagye:
Focusing the dock looks like the best idea, i'll just make the fall back for when there is no dock window to focus the root window.

VuiMuich:
The last patch is working for me and so far I don't see any issues.

Syudagye:
> Aha, so maybe `leftwm-state` breaks when the root window has focus? Possibly this generates a invalid combination of `mine`, `visible`, `focused` and `busy` parameters.

Well no, `leftwm-state` work fine, it's just polybar who freezes for some reason.
I also can't interact with the bar with my mouse (at least for the workspace part) when it has focus, for some reason. It's maybe a problem with polybar itself, i will further investigate this later.

> The last patch is working for me and so far I don't see any issues.

Noice :D ! i'll push a better implementation in a moment.

src/display_servers/xlib_display_server/xwrap.rs
```diff
@@ -1252,6 +1252,21 @@ impl XWrap {
...
+                    == Some(self.atoms.NetWMWindowTypeDock)
+            });
+            if let Some(dock) = dock {
+                self.window_take_focus(&Window::new(WindowHandle::XlibHandle(*dock), None, None));
+                return;
+            }
+        }
+        self.window_take_focus(&Window::new(self.get_default_root_handle(), None, None));
+    }
+
```
AethanFoot, line 1269: I say rather than using window_take_focus to use ```self.send_xevent_atom(handle, self.atoms.WMTakeFocus);```,
as for the dock most of the function is skipped (plus we don't care whether the user clicks it) and maybe the root window? (idk if the root window does anything if you ask to take focus, and if it does it might be less buggy then us forcing the focus 🤷‍♂️).

Syudagye:
Just looked at this after a night and found a better way of doing this. The best method was in fact to use the two binding in the `window_take_focus` and just modify them a little to give the input focus to the root, and change the root window property for the currently focused window to something ridiculous like `c_ulong::MAX` (for `leftwm-state` to work properly).

I also changed `window_history` to use `MaybeWindowHandle`, as suggested by @AethanFoot.

Syudagye:
Ah, and it solved the polybar problem somehow :shrug:

VuiMuich:
Looks good and is working fine for me.

mautamu:
> Ah, and it solved the polybar problem somehow 🤷

Glad to hear it! Good job 👏

PR #430, opened 2021-08-07
Added ToggleFloating and TileToFloating commands

CaioVieiraF:
also added options to config file for aways_float and default width
and height

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -11,6 +11,8 @@ pub enum Command {
     ToggleFullScreen,
     GotoTag,
     FloatingToTile,
+    TileToFloating,
+    ToggleFloating,
     MoveWindowUp,
     MoveWindowDown,
     MoveWindowTop,
--- a/src/config/theme_setting.rs
+++ b/src/config/theme_setting.rs
@@ -10,6 +10,9 @@ use std::path::Path;
 #[serde(default)]
 pub struct ThemeSetting {
     pub border_width: i32,
+    pub default_width: i32,
+    pub default_height: i32,
+    pub always_float: bool,
     pub margin: Margins,
     pub workspace_margin: Margins,
     pub gutter: Option<Vec<Gutter>>,
@@ -45,6 +48,9 @@ impl Default for ThemeSetting {
     fn default() -> Self {
         ThemeSetting {
             border_width: 1,
+            default_width: 800,
+            default_height: 600,
+            always_float: false,
             margin: Margins::Int(10),
             workspace_margin: Margins::Int(10),
             gutter: None,
@@ -74,6 +80,9 @@ mod tests {
     fn deserialize_custom_theme_config() {
         let config = r#"
 border_width = 0
+default_width = 400
+default_height = 400
+always_float = true
 margin = 5
 workspace_margin = 5
 default_border_color = '#222222'
@@ -91,6 +100,9 @@ value = 0
             config,
             ThemeSetting {
                 border_width: 0,
+                default_width: 400,
+                default_height: 400,
+                always_float: true,
                 margin: Margins::Int(5),
                 workspace_margin: Margins::Int(5),
                 gutter: Some(vec![Gutter {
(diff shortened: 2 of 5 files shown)


lex148:
@CaioVieiraF fine work here. This looks to be a really nice feature that would be great to have in LeftWM. I have one small request and one testing comment. 

Would you please at your two commands to the ExternalCommands and leftwm-command this way they can be ran even if you don't have keybindings setup. 

I think there is something a little wanky going on with the command TileToFloating. Sometimes it floats all the windows in the tag, Sometimes it only floats the one under the cursor. I personally like only the one under the cursor, but its your PR. 

Again great job. I can't wait to get this one merged. I see you have marked the PR as draft. I will wait until you think it is ready to go and give it another look

stevensonmt:
It looks like elsewhere in the window_handler module anytime there is a call to `window.set_floating` there is then a call to set `window.normal` and then to run `window.set_floating_exact`. Is this why the original window dimensions are not preserved in this branch when toggling to float?

CaioVieiraF:
> @CaioVieiraF fine work here. This looks to be a really nice feature that would be great to have in LeftWM. I have one small request and one testing comment.
> 
> Would you please at your two commands to the ExternalCommands and leftwm-command this way they can be ran even if you don't have keybindings setup.
> 
> I think there is something a little wanky going on with the command TileToFloating. Sometimes it floats all the windows in the tag, Sometimes it only floats the one under the cursor. I personally like only the one under the cursor, but its your PR.
> 
> Again great job. I can't wait to get this one merged. I see you have marked the PR as draft. I will wait until you think it is ready to go and give it another look

Thanks for the tips! I will try it out.



> It looks like elsewhere in the window_handler module anytime there is a call to `window.set_floating` there is then a call to set `window.normal` and then to run `window.set_floating_exact`. Is this why the original window dimensions are not preserved in this branch when toggling to float?

I just tested it and that was the missing piece! Now I'm just going to get every thing ready for the PR

stevensonmt:
If those steps are required for `set_floating` to work as expected, should it be refactored such that setting `window.normal` and `set_floating_exact` are called from `set_floating` itself? Is there a situation in which you would want `set_floating` without the subsequent `set_floating_exact`?

VuiMuich:
@CaioVieiraF you might want to consider adding these commands to the`external_command_handler.rs` as well. And if so please add them to the list in the help text of `leftwm-command.rs`

VuiMuich:
Hey, thanks for still returning!

May I request two things:
- the new fields you introduced to `theme.toml` would be great to be `Option<>` so we don't break all currently existing themes
- would you mind to get those merge conflicts ironed out?

AethanFoot:
theme_settings was combined into config. So they should accessible through manager.state.config.

VuiMuich:
Now that I looked at this again, I'd also reason, that applying the default-sizes should actually be done either in `window_handler` similar to the `snap_to_workspace` method. From there you should also be able to accsess the manager and the config

AethanFoot:
Nice! Thats working much better.

CaioVieiraF:
> @CaioVieiraF What do you use for your editor, as you should be able to set it up to auto fmt on save.

I use neovim, usually I manually run cargo fmt for that

AethanFoot:
If you use a plugin manager for nvim, I use [rust.vim](https://github.com/rust-lang/rust.vim). Then the settings:
```
let g:rustfmt_autosave = 1
let g:rustfmt_emit_files = 1
let g:rustfmt_fail_silently = 0
```
Then you never need worry about fmt again.

CaioVieiraF:
> If you use a plugin manager for nvim, I use [rust.vim](https://github.com/rust-lang/rust.vim). Then the settings:
> ```
> let g:rustfmt_autosave = 1
> let g:rustfmt_emit_files = 1
> let g:rustfmt_fail_silently = 0
> ```
> Then you never need worry about fmt again.

Nice, thanks for the tip!

PR #432, opened 2021-08-11
Margins per workspace

stevensonmt:
This works but there is one problem I have not solved. Take a `config.toml` that defines 3 workspaces, A, B, and C. A.id = 1, B.id is undefined, and C.id = 2. Checking for uniqueness at the config level would pass but at the screen creation level would lead to a bug with both B and C having id = 2. If you check for uniqueness and want to avoid a crash you have to have a strategy for resolving the ID collision. What seems to be the best way to handle this?


VuiMuich:
> This works but there is one problem I have not solved. Take a `config.toml` that defines 3 workspaces, A, B, and C. A.id = 1, B.id is undefined, and C.id = 2. Checking for uniqueness at the config level would pass but at the screen creation level would lead to a bug with both B and C having id = 2. If you check for uniqueness and want to avoid a crash you have to have a strategy for resolving the ID collision. What seems to be the best way to handle this?

Is there really a use case where it is practicable to leave B undefined in the config? If not I'd say this should error and fallback to default. Accordingly `leftwm check` should error as well, maybe even give a helpful message (guess I got spoiled by `eww`'s error messages, sorry).

stevensonmt:
I agree that it is not unreasonable to expect that defining an ID for any workspace means you should define it for all, but to make it backwards compatible the ID field needs to be an option type. This makes it hard to enforce defining an ID for each entry.
I also agree with falling back to a default, but what should that be? Ideally I would think you want to respect both the assigned IDs and the order of definition.

stevensonmt:
@VuiMuich do you mind looking at the most recent commit? I think I've implemented the fallback to default config correctly.

stevensonmt:
I will fix the clips and fmt issues. I hadn't realized that my vim plus ale configuration had stopped working correctly. I've fixed it now.

VuiMuich:
> @VuiMuich do you mind looking at the most recent commit? I think I've implemented the fallback to default config correctly.

I have not tested it yet, but by reading on mobile this looks good to me.

The comment in the diff is just a tiny nitpick, technically this should be ok as is.

Btw: @mautamu has posted a neat script for submitting, that checks `clippy` and `fmt` before pushing, you can find it in the [wiki: Contributing Tips and Tricks](https://github.com/leftwm/leftwm/wiki/Contributing-to-Leftwm---Tips-and-Tricks)

PR #435, opened 2021-08-12
fix #433

### LINKED ISSUE #433: [bug] scratchpad breaks special window move behaviour of layout MainAndDeck
**Describe the bug**
When a scratchpad is open while the layout `MainAndDeck` is used the special focus behaviour that `MoveWindowUp/Down` moves windows through the deck breaks. Instead the Scratchpad is considered as the `Main` window so the actual one gets moved away into the deck

**To Reproduce**
see above

**Expected behavior**
Idealy floating windows would be ignored or considered "additional main windows" so `MoveWindowUp/Down` still just cycles the deck

**Environment:**
 - OS/Distro: Manjaro
 - Output of `leftwm-check`
```
leftwm check
:: LeftWM version: 0.2.9
:: LeftWM git hash: 9906a0c
```

AethanFoot:
Fixes #433.
Thanks.


VuiMuich:
Thanks Aethan:rocket:

PR #438, opened 2021-08-13
Refactor tag to remove needless arc

### LINKED ISSUE #437: Bug: Focus Next and Previous Tag command takes you to a tag which shouldn't exist
I have set up keybindings for "FocusNextTag" and "FocusPreviousTag" and I have a total of 9 tags. Moving between tags works as intended using the keybindings, however, when I go to the previous tag from Tag 1, it takes me to a hidden tag which shouldn't exist. Same thing happens when I go to the next tag from Tag 9. It's as if there's a hidden Tag 10 which is only accessible through this the focus tag command.

**Steps to reproduce the behavior:**
1. Set up keybindings for "FocusNextTag" and "FocusPreviousTag" or execute `leftwm-command "FocusNextTag" / "FocusPreviousTag" ` from the terminal
2. Try going to the previous tag from your first tag or next from your last tag using the commands from the first step
3. You'll end up in a new tag which works but shouldn't exist

**Expected behavior**
When "FocusPreviousTag" is executed from the user's first tag, it should take the user to the last tag (Tag 9, in my case). 

When "FocusNextTag" is executed from the user's last tag, it should take the user to the first tag (Tag 1, in my case).

AethanFoot:
Initial refactor of tag, there will probably other bits around this that I will change. This fixes #437 in this, with the solution being to add a hidden flag to a tag (which will be on by default for the scratchpad tag) as it may be useful to have a general way to hide a tag rather then a specific method for just the scratchpad tag.
Thanks!

### DIFF
--- a/src/bin/leftwm-worker.rs
+++ b/src/bin/leftwm-worker.rs
@@ -3,9 +3,8 @@ use leftwm::{
     models::{FocusBehaviour, FocusManager, Tag},
 };
 
-use crate::models::TagModel;
 use leftwm::{
-    config, external_command_handler, models, CommandPipe, DisplayEvent, DisplayEventHandler,
+    config, external_command_handler, CommandPipe, DisplayEvent, DisplayEventHandler,
     DisplayServer, Manager, Mode, StateSocket, Window, Workspace, XlibDisplayServer,
 };
 use std::panic;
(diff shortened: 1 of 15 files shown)


VuiMuich:
This is great!
I believe it might also help to simplify implementing the possibility for having different layouts per tag.

AethanFoot:
This now fixes #437, in the future we will want to decide whether a user should be able to access hidden tag with a direct goto_tag and whether the user can move windows to a hidden tag with move_to_tag (currently they can with both, I have just made so they can't accidentally show any hidden tags).
Thanks.

PR #440, opened 2021-08-15
Partial liquid templates resolve #439

### LINKED ISSUE #439: Path for partial liquid  templates
I have been unable to find any documentation on liquid that would help me understand the directory structure needed to use partial templates within liquid templates. I found one shopify specific thread that suggested templates need to be in a `templates` directory with a sibling `snippets` directory for partial templates. This does not seem to work for my polybar config.
```
my_theme/
               |_polybar.config
               |_templates/
               |          |_main_template.liquid
               |
               |_snippets/
                         |_partial.liquid
```
I have tried different syntax in main_template.liquid, such as: `{% include 'partial' %}`, `{% render 'partial' %}`, `{% snippet assign partial = 'partial' %}`, and replacing single quotes with double quotes. These return errors of `partial does not exist` for `include` or `unable to parse template ... Unknown tag` for the `render` and `snippet assign` version. Any suggestions?

stevensonmt:
This will add support for liquid templates that make use of partial templates. It assumes partial template filenames begin with an underscore and are in the same directory as the template passed to `leftwm-state -t <template>`. As implemented the full path is registered with liquid, which I believe means the partial template would have to be called with the full path. It would be simple enough to process the path to register just the file name. It could then be called with `{% include 'filename' %}` as opposed to `{% include '/path/to/file/name' %}`. I'm not sure which is preferable.

There is also a question about whether `leftwm-check` should include a check for templates.


stevensonmt:
The last commit message is in error. The test passes but it only does so if there is a release binary and the tests are run with the `--release` flag. This is apparently a limitation of the `assert_cmd` crate. I will refactor it to use `escargot` per the recommendation of the `assert_cmd` crate author.

stevensonmt:
I think this is good. The test is quite slow, taking about 4 seconds on my machine. Not sure what can be done to speed that up as it's probably due to the creation of the temp files.

stevensonmt:
@VuiMuich the last commit might have some useful ideas for leftwm-check using the assert_cmd crate.

lex148:
@stevensonmt great work. This looks like an excellent addition to LeftWM. Do you know what could be going on with the build here? I want to try to get it building with the CI server before we merge.

stevensonmt:
The test I wrote runs the leftwm-state binary. It works fine locally but seems to not work in CI.

stevensonmt:
I don't know how to re-write that test to avoid this issue with the CI build. Any suggestions?

lex148:
This is a hard one, and there isn't really a correct answer. 

How about this, the function get_partials has three jobs it finds files in a directory, finds the files that start with an underscore, and converts them into partials. Finding all files, and converting into partials is really interfacing with external stuff. That makes it really hard to test. Really the only thing you are worried about here is the underscore filter. If you split this into multiple functions you could test just that part and call it done.

stevensonmt:
I'm struggling to come up with a way to split it up that doesn't end up with something trivial like:
```
let partial = OsStr::new("_partial.liquid");
assert!(partial_check(partial))
```
and the partial_check definition is just
```
fn partial_check(file: OsStr) -> bool {
    let file_str = file.to_str().unwrap_or(" ");
    file_str.starts_with('_') && file_str.ends_with(".liquid")
}
```

stevensonmt:
@lex148 I think the last commit addresses the issue with the test. I apologize for the extra builds on the CI that are triggered by my micro-commits. I forgot to remove the unnecessary dependencies from `Cargo.toml` and then tried to fix my error of not ignoring the sweep.timestamp file initially, which caused an issue with conflicting gitignore files. Should be good to go now.

stevensonmt:
Once this is merged and part of a release I can update the config wiki. Just making this comment to remind myself.

PR #445, opened 2021-08-18
Add theme checks to `leftwm-check`

VuiMuich:
This adds checks for the essential theme files `up` and `down` scripts (including check for executable permissions) and `theme.toml`.

It is quite clumsy I'll freely admit and broken in some parts (doesn't handle missing files gracefully), but basically it works - I guess.
Also clippy (pedantic) has an opinion on the `let tf = match BaseDirectories::with_prefix(path_current_theme.clone().unwrap())?.find_config_file(&file_name)` suggesting an `if let` here, but I struggle to keep `tf` in scope then...

I guess my Rust got a bit rusty :lol:


VuiMuich:
Argh, all ths commits in the PR from a previous merge, sorry..

stevensonmt:
I'm looking at the xdg crate and having a hard time understanding how to use the `BaseDirectories::find_config_files` method. It seems like the path to search should be `self` and if you knew a specific filename you'd use `find_config_file`. But if you could use the iterator that `find_config_files` returns, you could then `match` on the `Item` like this pseudocode
```
match file {
    "up" => check_permissions(file),
    "down" => check_permissions(file),
    "theme.toml" => check_toml_settings(file)
}
```

VuiMuich:
Yeah, that was kind of my assumption as well.
Maybe I just got an idea from your pseudo code:
I believe the iterator holds the absolute paths of the found config files so we could try to match on a `ends_with()`. 

On a side note: What I also find very irritating is that as you said `find_config_file` should only take `self`, but when called like that it complains about missing arguments.

VuiMuich:
Thanks @stevensonmt your refactor PR on my fork was a great help!

Error Messages are still all over the place style wise. But other than that I believe this is almost merge ready.

mautamu:
I’ll try to test it later tonight :). 

> Error Messages are still all over the place style wise


Perhaps it’s about time for a nice error/warn/trace macro. I know we’d discussed reducing dependencies in another issue/PR so probably should just implement our own.

PR #446, opened 2021-08-18
Update `EWW` example theme to new `yuck` config language

VuiMuich:
elkowar/eww merged their config rework branch into master and thereby switched from xml as config language to their onw created simple exprassion based `yuck` config language.

draft because something in the eww-bar example is broken right now (or it is just my setup? my handwriten bars are working fine though)
as soon as the example runs well will open for review, just though somebody might be interested in it already.


VuiMuich:
ahh crap butchered it again.. will fix tomorrow...

Issue #447, opened 2021-08-19
[Question] Is there any way to manage layout on per tag basis?


tabetai:
When I switch between layouts they change for all tags. Can I somehow change layouts for tags independently?

VuiMuich:
Currently it's not but it kind of is a long desired feature of mine as well.

I don't see if I can attempt an implementation anytime soon. But if no one will, at some point I might try this eventually.

AethanFoot:
Hopefully I can start this on the weekend. To start off with there wont be any setup in the config, but I will open a pr later for a greater conversation on what and how we handle everything.
Thanks.

stevensonmt:
I like this idea. I would suggest users should be able to toggle "apply layout change to all tags" as a bool. If someone is expecting that behavior it would be very tedious for them to have to apply the layout to each tag individually.

VuiMuich:
> I like this idea. I would suggest users should be able to toggle "apply layout change to all tags" as a bool. If someone is expecting that behavior it would be very tedious for them to have to apply the layout to each tag individually.

This might actually be also a good command. 

So the config would have an `apply_layouts_to_all_tags` bool with default `true` (to keeping existing behavior persistent) and we get a new command `ToggleApplyLayoutsToAllTags` (name is a bit awkward though) that toggles the above bool and when switching from `false` to `true` applies the layout of the current tag to all others?

PR #356, opened 2021-05-24
Focus manager

AethanFoot:
This is a start on #326. So far the basic model is set up with all previous functionality, I have the click replay working for click to focus (currently all processed clicks replay which will change). I need to implement the things from #350, and try and clean up areas where focus comes up. I will also remove moving the pointer when focus isn't sloppy. 
If anyone has any opinions please share. I will be continuing this tomorrow hopefully.
Thanks.

Issue #260, opened 2021-03-30
Remember layout by tag


jwnpoh:
Is there a way to get LeftWM to remember a different layout for different tags? For eg., I would like tag 1 to be in Monocle and tag 2 to be MainAndVertStack.

It seems like LeftWM's default behaviour is to remember layout by workspace, which is useful for multi-monitor setups but for a single screen, maybe remembering layout by tag is more useful?

Thanks!

VuiMuich:
Currently there is now way to do this.
And I agree, that at least on single-workspace setups tag-centered behaviour would be more intuitive.

I have this on my personal whish-list as well, together with a possibility to assign default workspaces for specific programs, like "browsers" on tag 2, mocp and vlc on tag 6 etc.

AethanFoot:
Should we set this up as a discussion, as personally I feel that per tag layouts feels more natural where you could set a default layout for a tag. However this would be a fundamental change, in the extreme case it could be an option to chose between them, or at least make it default to tag centric with one workspace (however this may become confusing and bloated).

VuiMuich:
> Should we set this up as a discussion, as personally I feel that per tag layouts feels more natural where you could set a default layout for a tag. However this would be a fundamental change, in the extreme case it could be an option to chose between them, or at least make it default to tag centric with one workspace (however this may become confusing and bloated).

+1 on having this in discussions. 
I feel there are several topics, that would live better in the discussion category and not in issues or comments on PRs. 
Especially for some general design questions like this.

mautamu:
If we are to add this to config.toml, one question that comes to mind is whether leftwm should save state or not? E.g. I start leftwm with a tag named Alpha in Monocle layout:
```toml
[[tag]]
name="Alpha"
layout="Monocle"
```
And then change layouts to, for example, Fibonacci. Should config.toml track this change?
```toml
[[tag]]
name="Alpha"
layout="Fibonacci"
```

Another couple of ideas that I've had: perhaps a `leftwm-config` that replaces `leftwm-check`. It would allow us to avoid issues like #188 by providing an interface to capture keypress names as LeftWM knows them (e.g. Modkey + Shift + eacute) and bind them to commands that LeftWM knows. Example excerpt:
```
mau@mau-pc ~> leftwm-config keys
==> 5 Keybinding found:
    [1] Modkey + Shift + X           -> HardReload
    [2] Modkey + Shift + R           -> SoftReload
    [3] Modkey + e                       -> Execute
[...]

PR #448, opened 2021-08-19
Layout Per Tag

AethanFoot:
Tagging #260, #447, #292. This is just an initial commit, and to open a bit of a discussion. I agree with the notes in #447, where there should be an option to allow the user to choose to have per tag or per workspace layouts with which could be toggled via a command. Other issues we should address is the management of layouts, with this added complexity it becomes harder for a new dev to find their way through the code, e.g. the fact that the layout vec is stored in config, manager, workspace, and now tag. I feel it would be helpful to add a layout manager similar to the [focus manager](#356), which holds the list, which mode the user is in, and common functions around changing the layout. 

Another option we could add would be choosing the mode per workspace, say tags sent to workspace one will be forced to the same layout but on workspace two will use the tags preferred layout.

Many Thanks!

### DIFF
--- a/src/bin/leftwm-worker.rs
+++ b/src/bin/leftwm-worker.rs
@@ -37,7 +37,7 @@ fn main() {
         let mut tags: Vec<Tag> = config
             .get_list_of_tags()
             .iter()
-            .map(|s| Tag::new(s))
+            .map(|s| Tag::new(s, config.layouts.clone()))
             .collect();
         tags.push(Tag {
             id: "NSP".to_owned(),
--- a/src/handlers/screen_create_handler.rs
+++ b/src/handlers/screen_create_handler.rs
@@ -30,7 +30,7 @@ pub fn process(manager: &mut Manager, screen: Screen) -> bool {
     //make sure are enough tags for this new screen
     if manager.tags.len() <= tag_index {
         let id = (tag_index + 1).to_string();
-        manager.tags.push(Tag::new(&id));
+        manager.tags.push(Tag::new(&id, manager.layouts.clone()));
     }
     let next_tag = manager.tags[tag_index].clone();
     focus_handler::focus_workspace(manager, &workspace);
--- a/src/layouts/mod.rs
+++ b/src/layouts/mod.rs
@@ -48,6 +48,12 @@ pub const LAYOUTS: [Layout; 12] = [
     Layout::LeftWiderRightStack,
 ];
 
+impl Default for Layout {
+    fn default() -> Self {
+        Layout::MainAndVertStack
+    }
+}
+
 // This is tedious, but simple and effective.
 impl Layout {
     pub fn new(layouts: &[Layout]) -> Self {
--- a/src/models/tag.rs
+++ b/src/models/tag.rs
@@ -1,12 +1,13 @@
-// We allow this so that we can be verbose and indicate that
-// this is `TagModel` and not `WindowModel` or anything else.
-#![allow(clippy::module_name_repetitions)]
 use serde::{Deserialize, Serialize};
+
+use crate::layouts::Layout;
 
 #[derive(Default, Serialize, Deserialize, Debug, Clone)]
 pub struct Tag {
     pub id: String,
     pub hidden: bool,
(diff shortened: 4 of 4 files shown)


stevensonmt:
In my head I equate workspaces to actual screen real estate and tags to collections of opened windows. This feature applied per workspace would allow specific workflows to take advantage of specific monitor dimensions/orientation. It would also allow for tags to define a certain workflow that could be set up on any workspace/monitor. 
With that in mind to maximize the utility of the feature I would consider having both options available but then there's the issue of which setting gets precedence and should that be user configurable?
I appreciate as a newcomer to leftwm and development in general the desire to keep the code human readable. The layout manager idea makes a lot of sense in that regard.

VuiMuich:
I wholeheartedly agree on the idea of a layout manager!

Maybe instead of a bool it is more like a tuple `apply_layout_to` with options `workspace`, `all_tags`, `individual_tag` and the toggle command cycles through the modes?
We might move this option to be a optional field of the detailed workspace config, same should go for the layout option.
Maybe something like this:
```toml
[[workspaces]]
x = 0
y = 0
height = 1200
width = 1920
layouts = ["MainAndDeck", "Monocle"]
apply_layout_to = ["workspace", "all_tags", "individual_tags"]

[[workspaces]]
y = 1200
x = 0
height = 1440
width = 1280
layouts = ["MainAndDeck"]
apply_layout_to = ["workspace"]

[[workspaces]]
y = 1200
x = 1280
height = 1440
width = 1280
layouts = ["MainAndDeck", "Monocle", "CenterMainBalanced", "LeftWiderRightStack"]
apply_layout_to = ["individual_tags"]
```

stevensonmt:
Maybe this is not a configurable setting at all but a command that takes a layout plus a workspace or tag as arguments. Users can run the command to set a desired layout per tag or workspace in their up scripts to make it theme-able. Keeps it from propagating another setting throughout the code. Is that overly simplistic?

lex148:
Sorry a bit late to the conversation. The layout manager is a great idea. Abstracting code and concepts away is always a big win. I would even take it one step further and say you should pull each behavior into its own file under the layout manager module.

AethanFoot:
So this has a base implementation of layout per tag being togglable between tag and workspace. It is ready and should be easy and extend this with greater customisation (don't know if we should wait for the new config format before deciding more complex configuration setups?). This has also simplified routing through the code with more things directly accessing what they need directly, and we are no longer passing around  vectors of tags or layouts when not needed.

This should be merged after the fixes #499 and #500, and I will fix any conflicts.

Thanks!

lex148:
@AethanFoot I am excited to get this one merged in but am having a little trouble testing. I see the config setting to change from Tag to Workspace mode, but it doesn't seam to change anything. Am I missing something? I also see you are defaulting to Workspace. I think that is a really good Idea while we are still developing this. Lets keep the default at Workspace until we are sure we have all the kinks worked out. After that we can set it to whatever everyone wants.

VuiMuich:
I am experiencing three regressions:
- `MoveFocusUp/Down` behaves like `MiveWindowUp/Down`
- `leftwm state -w 0 -s "{{workspace.layout}}"` does not update on layout change and always prints `Monocle`
- layout state does not survive `SoftReload` (log: `leftwm-worker[738246]: Cannot load old state: missing field `layout` at line 1 column 17062`)

AethanFoot:
@VuiMuich hopefully those are fixed now?

lex148:
I am still experiencing layouts being applied by tags while in workspace mode.

Here are the steps I did for testing:
1) rebuild everything on this branch
2) deleted my config file. `./config/leftwm/config.toml`
3) `pkill leftwm-worker` 
4) verified the new config.toml file is in `layout_mode=Workspace`
5) open a bunch of windows on both monitors. 
6) changed layout to all horizontal for wide screen monitor
7) changed layout to all vertical for vertical monitor
8) swapped the two tags

Result:
the wide screen monitor is now displaying windows using the vertical layout
the vertical monitor is now displaying windows using the horizontal layout

Expected:
I expected the layouts to remain the same on the two workspace.

PR #455, opened 2021-08-22
Reduce amount of dependencies for libraries

cecton:
Hello :wave: 

I would like to introduce a few PRs that will improve the library-ness of leftwm. My goal is to leave the binary leftwm as close as possible to what it is but to make leftwm usable as a lib and strictly limited to a tiling manager.

This is because I found leftwm's description the most interesting but I believe it doesn't hold enough to what it says it is:

> LeftWM is designed to do one thing and to do that one thing well: *be a window manager.*

I think we can do better. The first step is to reduce the unnecessary dependencies when used as a library. More PR will come as I work to integrate in my personal project.

### What this PR does

A user who would want to use leftwm as a library is now able to deactivate the unnecessary dependencies by using `default-features = false`. Nothing else changed in this PR.


lex148:
@cecton this is a great goal. I cant wait to see what you come up with.

lex148:
this clippy error has nothing to do with you I'm going to go ahead and merge

PR #461, opened 2021-08-25
Add check for modifier keys

TotalDarkness-NRF:
Adds a check for the modifier keys when editing leftwm config.
Previous behaviour would allow modifier keys to be anything even when they were not a valid key.
Will output an error when the modifier key is not valid!

### DIFF
--- a/src/bin/leftwm-check.rs
+++ b/src/bin/leftwm-check.rs
@@ -157,15 +157,11 @@ fn check_keybinds(keybinds: Vec<Keybind>, verbose: bool) -> bool {
             ));
         }
 
-        keybind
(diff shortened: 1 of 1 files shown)


VuiMuich:
Good catch!

I was thinking if the error message could be improved a tiny bit to help understanding that the key is not valid as modifier key but couldn't come up with satisfactory wording yet.
Best idea is like
`"The key '{}' is not valid for use as modifier."`

lex148:
@TotalDarkness-NRF awesome find. thanks for this fix! I can really imagine this one messing with someone if they got it wrong.

PR #462, opened 2021-08-25
store hidden windows on correct sides so picom animation works

lex148:


lex148:
If you are using a picom fork like picom-jonaburg-git, windows slide in from there last location. This is a great affect, but is all messed up currently when switching tags. That is because invisible windows are currently all stored to the far left of the screen. This update hides invisible windows on the correct size so they side in appropriately

PR #463, opened 2021-08-28
Use traits and move away non-related tiling manager logic

cecton:
Hello again! :wave:

This time it looks like a big refactoring as you can see. I understand it will
take time to review. No worry.

There are 2 major changes:

1.  The `Config` struct has been moved to a module `common` that is only for
    binaries. A trait `Config` has been added to the lib to abstract its
    functionalities.
2.  Same for `State`.

This PR is part 2 of the whole libraryfication process. The first part was to
make a few dependencies optionals which is nice but it's not as important as
this new PR which actually make the lib usable.

To do this, I created my own binary crate where I copy-pasted the code of
leftwm-worker. Then I added leftwm to my dependencies and I fixed all the
issues in order to make it worker. These changes include essentially making
things public and adding a trait here and there.

It's now working! You can create a tiling manager binary based on leftwm. But
it's not perfect and will need to be refined. I propose to do that in at least
2 other PRs. My reasoning here is that if I change the code too much it will
become totally unrecognizable to the maintainers and it will be harder to
review. That is why I try to stay as close as possible to the current code.

Please do not hesitate to ask me to change things or move things around. Just
keep in mind that I will also refine things a lot in a future PR.


AethanFoot:
Having a quick look, it looks great. Amazing work!
Thanks.

cecton:
I forgot to do the ThemeSetting. It was a bit more complicated there: I kept the struct and I added a ThemeLoader trait to load the settings.

lex148:
Great work! I have been looking over this PR and all looks good except one little problem. It appears to be loading the default theme.toml instead of the one set using the LoadTheme command

cecton:
> Great work! I have been looking over this PR and all looks good except one little problem. It appears to be loading the default theme.toml instead of the one set using the LoadTheme command

Thanks!! :tada: 

This is weird, the code looks good... :thinking: I suspect the file couldn't be loaded for some reason so it fallsback to the default theme. I've improved the logging to log an error if the file couldn't be loaded. If you don't mind giving another try?

cecton:
> Again awesome work. I can't wait to get this one merged. I always really like cleanup PRs like this.

Thanks a lot!! It's very reassuring and very appreciated. I hope to do a few more in the coming months.

> I hate to have to ask this of you but could you please update your PR to not make `workspace_margin` required. It was renamed from `margin` a while back but we still have so many old themes that use margin not workspace_margin.

Sure no problem! Though there was no default... so I used `Margins::Int(0)`? Or should it default to copy the margin of the window (field `margin` in `ThemeSetting`)?

VuiMuich:
> Sure no problem! Though there was no default... so I used `Margins::Int(0)`? Or should it default to copy the margin of the window (field `margin` in `ThemeSetting`)?

I think this is a sane default.

cecton:
Ok I'll try to keep that in mind for the future :+1:

PR #472, opened 2021-09-05
Move away more non related tiling manager logic

cecton:
Coming in next PR (not this one): split crate in 2: leftwm-core and leftwm

Hopefully this is the last "intensive" PR. The next one will be mostly about moving things around and the one after should be more scoped to specific issues.

### Changes
- move independent functions to `Manager`
- move event loop to `Manager`
- merge lib's `ThemeSetting` into lib's `Config`
- move theme loading/unloading commands to bin's custom commands
- removed ExternalCommand in favor of `Command`
- proper types for `Command`
- ~add a generic type for implementing custom commands~ add an `Other(String)` variant for implementing custom commands
- introduce anyhow for the bin to simplify error handling (some coloring has been lost... oh well)
- better logs when sending command to unix pipe
- really happy with all the stuff clippy found this time, great
- move `Config` instance and `DisplayServer` instance to `Manager`, the worker's code is greatly simplified
- introduce `thiserror` to simplify error declarations
- dropped ErrorKind (I didn't see any value)


AethanFoot:
Having a "quick" look over this, everything is looking good. One note I have to make is I'm not a fan of the "C" and "CMD", the CMD is ok but feels weird for some reason but I feel the C is more ambiguous (maybe an abbreviation would work but with the dual C's having full words would make them more distinguishable). 
Awesome work!
Thanks.

cecton:
> Having a "quick" look over this, everything is looking good. One note I have to make is I'm not a fan of the "C" and "CMD", the CMD is ok but feels weird for some reason but I feel the C is more ambiguous (maybe an abbreviation would work but with the dual C's having full words would make them more distinguishable).
> Awesome work!
> Thanks.

Yes it was already my intention to rename it to CONFIG. But this PR overall is WIP so don't worry too much

VuiMuich:
Hey, may I ask a question not totally related to this #472 PR, but still to your big refatoring project @cecton?

While working on #473 I noticed we now have `src/state.rs` and `src/bin/common/state.rs`.
Now I am wondering for the reasoning, why keep the `src/state.rs` at all, when it only holds `pub trait State`?
Excuse me if this appears to be a dumb question and is obvious to properly educated developers, but at least to my layman pov it is slightly confusing to have to files of the same name in two places of the repo.

cecton:
> Now I am wondering for the reasoning, why keep the `src/state.rs` at all, when it only holds `pub trait State`?
> Excuse me if this appears to be a dumb question and is obvious to properly educated developers, but at least to my layman pov it is slightly confusing to have to files of the same name in two places of the repo.

No question is dumb, no worry!! :hugs: 

Yes the naming is something I would like to address at some point. I'm just afraid that doing it will confuse the existing knowledge every contributor has at the moment on the project and the review process of my changes. I was planning to do it in the next PR when the crate gets split.

The trait `State` was used to let the user of the library decide what will be the implementation for saving the state. This is the same reasoning for the trait Config. Please note though that in this pr (#472) I'm going to actually merge that functionality to the Config trait.

My idea is that leftwm-core will hold the tiling manager logic: connecting to the X.org session, managing windows, bind commands to keys, layouts, ... but not the way the configuration is loaded: border size, margins, actual keymapping, theme, etc... The configuration will be provided by a trait so leftwm (bin) will dictate how things are provided.

In my personal use case I like having the configuration compiled so I make my own bin and hardcode everything (like dwm, xmonad).

VuiMuich:
Hey, thanks for this explanation. 

Now that makes a lot of sense.

Really appreciate the work you are doing here, so great and so much to learn from!!
Also appreciate your way of doing it in several PRs so we can keep up with the changes.

Thanks a ton!

cecton:
One TODO in my list solved thanks to your work @VuiMuich :heart: (590b636)

I knew there was something wrong this this field "tags" but couldn't find out what exactly :grin:

cecton:
@AethanFoot @VuiMuich @lex148 part 3 is ready :sweat_smile: 

I already have some code for part 4 but it breaks too much stuff if I don't split the crate so... it will wait in a branch :sweat: 

I tried to list all the changes I did. Don't hesitate to commit directly in the branch if you want! And I hope you have a good weekend :hugs:

src/layouts/mod.rs
```diff
@@ -123,13 +123,17 @@ impl Layout {
         if index < 0 {
             index = layouts.len() as isize - 1;
         }
-        layouts[index as usize].clone()
+        layouts[index as usize]
     }
 }
 
-// TODO: Perhaps there is a more efficient way to impl FromStr using serde
```
cecton, line 130: I removed this TODO because there is no way using serde. That would enforce a serialization format just for a few string... not great.

I could make a macro_rule if you want... :thinking:

src/bin/common/command.rs
```diff
@@ -0,0 +1,42 @@
+use serde::{Deserialize, Serialize};
+
+// TODO this code is temporary. Due to the limitations of TOML we cannot serialize leftwm::Command
+//      easily. If we replace TOML by JSON/JSON5/YAML we will be able to remove this code and a
+//      bunch of validation in leftwm-check.rs. This requires to deprecate the TOML config file,
+//      thus making a breaking change.
+//
+//      https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=59232ae3a6f902fc3a3a7a09d1d48c80
```
cecton, line 8: We should create an issue to deprecate the toml config and use something else. I love TOML but in this case it really fails... The issue is that it can't serialize enums that has variants with values (struct variants and tuple variants).

VuiMuich: I am all in favour f this.
I started looking into alternative config languages, when `eww` did their poll/disussion when they moved away from xml.
They wrote their own lisp-like config language called `yuck`, but I don't think that something like this would be a good fit for us.
Back then, I thought `dhall` might be nice, but it might as well be a bit over the top, plus it comes with a ton of crates it depends on.

AethanFoot (commented):
Overall looks good.
Are the three changed commands (swapscreens, sendwindowtotag, and movewindowtolastworkspace) the only config breaking changes? I do agree these are clearer names but should we change them in #292? If we feel we can change them now we should create a warning for users that update to this.
Thanks.

cecton:
> Are the three changed commands (swapscreens, sendwindowtotag, and movewindowtolastworkspace) the only config breaking changes?

That's a mistake on my side. I actually tried to preserve the original names but I probably copy-pasted the wrong thing. I will fix it.

> I do agree these are clearer names but should we change them in #292? If we feel we can change them now we should create a warning for users that update to this.

Yes I prefer that this change come in another PR.

src/handlers/window_handler.rs
```diff
@@ -128,182 +255,82 @@ fn setup_window(
...
-}
-
-fn find_window<'w>(manager: &'w Manager, handle: &WindowHandle) -> Option<&'w Window> {
-    manager.windows.iter().find(|w| &w.handle == handle)
+fn find_window<'w, C: Config, SERVER: DisplayServer>(
+    manager: &'w Manager<C, SERVER>,
+    handle: &WindowHandle,
+) -> Option<&'w Window> {
+    manager.state.windows.iter().find(|w| &w.handle == handle)
 }
```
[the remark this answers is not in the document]
cecton: Ok done

cecton:
(This is ready again for review btw)

src/display_servers/xlib_display_server/xatom.rs
```diff
@@ -1,6 +1,9 @@
 use std::ffi::CString;
 use x11_dl::xlib;
 
+// Specifications can be found here:
+// https://specifications.freedesktop.org/wm-spec/1.3/ar01s03.html
```
VuiMuich, line 5: This is very helpful ❤️

src/display_servers/xlib_display_server/xwrap.rs
```diff
@@ -1444,11 +1443,11 @@ impl XWrap {
         }
     }
 
-    pub fn load_colors(&mut self, theme: &ThemeSetting) {
+    pub fn load_colors(&mut self, config: &impl Config) {
         self.colors = Colors {
-            normal: self.get_color(&theme.default_border_color),
-            floating: self.get_color(&theme.floating_border_color),
-            active: self.get_color(&theme.focused_border_color),
+            normal: self.get_color(config.default_border_color()),
```
VuiMuich, line 1448: Just for my understanding, why is no borrow needed here?

src/utils/command_pipe.rs
```diff
@@ -52,107 +53,96 @@ impl CommandPipe {
...
-        "FocusNextTag" => Ok(ExternalCommand::FocusNextTag),
-        "FocusPreviousTag" => Ok(ExternalCommand::FocusPreviousTag),
-        "FocusWorkspaceNext" => Ok(ExternalCommand::FocusWorkspaceNext),
-        "FocusWorkspacePrevious" => Ok(ExternalCommand::FocusWorkspacePrevious),
-        "NextLayout" => Ok(ExternalCommand::NextLayout),
-        "PreviousLayout" => Ok(ExternalCommand::PreviousLayout),
-        "RotateTag" => Ok(ExternalCommand::RotateTag),
-        "CloseWindow" => Ok(ExternalCommand::CloseWindow),
-        // These require arguments and might be more finicky
-        "LoadTheme" => build_load_theme(s),
```
VuiMuich, line 98: If we lose `LoadTheme` and `UnloadTheme`, we cannot switch themes, without at least a `SoftReload`, right?
Personally I haven't used this possibility yet, but I believe it might actually be nice to keep this. Or am I missing the point where this is still possible, but the way its triggerd has changed?

src/bin/leftwm-check.rs
```diff
@@ -246,10 +222,7 @@ fn check_elogind(verbose: bool) -> Result<()> {
             }
             println!("\x1b[1;91mERROR: XDG_RUNTIME_DIR not set and elogind not found.\nSee https://github.com/leftwm/leftwm/wiki/XDG_RUNTIME_DIR for more information.\x1b[0m",);
```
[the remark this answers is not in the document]
cecton: I probably left it because it's not part of the error instance itself... :thinking: I don't see any harm to remove it (or merge it with the error message below), it was probably a duplicate message before.

src/display_servers/xlib_display_server/xatom.rs
cecton: I'm often confused by signed and unsigned numbers from C stuff :sweat_smile:

src/display_servers/xlib_display_server/xwrap.rs
cecton: The functions already return a borrowed value.... Though I'm not entirely sure if the Config trait's functions should return owned values or borrowed value :thinking: Right now it's a bit mixed up... There should be a right answer, maybe it will reveal itself in a future PR. Don't hesitate to let me know if you have any idea on this.

src/utils/command_pipe.rs
cecton: Yes it is still possible. When a command string is not recognized it is forwarded to Config::command_handler. This function works for both CommandPipe **AND** for keybind Command::Other. You will see that the code to do the theme load/unload is in `src/bin/common/config.rs`.

My idea is that, if I have a tiling manager library, there is no concept of "theme". It's a great concept for leftwm to abstract to the user that kind of configuration but it is an unneeded abstraction for the lib itself.

src/bin/leftwm-check.rs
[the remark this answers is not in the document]
VuiMuich: Yeah I guess the link would be helpful to be merged, the rest is kind of duped

src/display_servers/xlib_display_server/xwrap.rs
VuiMuich: Oh, thanks should have followed the rabbit hole a bit deeper I guess.
At this point I'm also not sure what would be best.

VuiMuich:
> > Are the three changed commands (swapscreens, sendwindowtotag, and movewindowtolastworkspace) the only config breaking changes?
> 
> 
> 
> That's a mistake on my side. I actually tried to preserve the original names but I probably copy-pasted the wrong thing. I will fix it.
> 
> 
> 
> > I do agree these are clearer names but should we change them in #292? If we feel we can change them now we should create a warning for users that update to this.
> 
> 
> 
> Yes I prefer that this change come in another PR.

Should we 'alias' the old commands then and print a deprecation warning in logs and `leftwm-check`?

cecton:
> Should we 'alias' the old commands then and print a deprecation warning in logs and `leftwm-check`?

I don't think it's worth it if we deprecate the toml config entirely.

mautamu:
Do we have something to replace it with yet to work out #292 with? 

> > Should we 'alias' the old commands then and print a deprecation warning in logs and `leftwm-check`?
> 
> I don't think it's worth it if we deprecate the toml config entirely.

VuiMuich:
Valid point.

I'll open a discussion to collect possible replacements for toml.

src/bin/leftwm-check.rs
```diff
@@ -48,9 +47,6 @@ async fn main() -> Result<()> {
     match load_from_file(config_file, verbose) {
         Ok(config) => {
             println!("\x1b[0;92m    -> Configuration loaded OK \x1b[0m");
-            if config == Config::default() {
-                println!("\x1b[1;93mWARN: The file loaded was the default. Your configuration is likely invalid \x1b[0m");
```
mautamu, line 52: I'm not sure I follow this, what're we replacing this warning with? (Apologies if it's covered elsewhere).

src/bin/leftwm-check.rs
```diff
@@ -133,31 +126,14 @@ fn check_workspace_ids(workspaces: Option<Vec<Workspace>>, verbose: bool) -> boo
...
-        if value_required_commands
-            .iter()
-            .any(|i| i.clone() == keybind.command)
-            && keybind.value.is_none()
-        {
-            returns.push((
-                Some(keybind.clone()),
-                "This keybind requres a `string`".to_owned(),
-            ));
+        if let Err(err) = leftwm::Keybind::try_from(keybind.clone()) {
```
mautamu, line 135: This is awesome; thank you very much for this!

src/bin/leftwm-check.rs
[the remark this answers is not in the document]
mautamu: Yeah, would be good to merge with the link I think

src/bin/leftwm-check.rs
```diff
@@ -330,77 +300,47 @@ fn check_current_theme_set(filepath: &Option<PathBuf>, verbose: bool) -> Result<
...
             }
-            Ok(filepath)
         } else {
-            let error = "\x1b[1;91mERROR: No `theme.toml` found\x1b[0m".to_string();
-            println!("{}", error);
-            Err(leftwm::errors::LeftError::from(std::io::Error::new(
-                std::io::ErrorKind::Other,
-                error,
-            )))
+            bail!("No `theme.toml` found at path: {}", filepath.display());
```
mautamu, line 343: Changes like this make leftwm-check much more ergonomic, thanks!

mautamu (commented):
This is fantastic; thank you for this! I'm not *entirely* on board with dropping Un/LoadTheme, but I appreciate where we're going and what we're trying to accomplish by doing so.

VuiMuich:
> This is fantastic; thank you for this! I'm not _entirely_ on board with dropping Un/LoadTheme, but I appreciate where we're going and what we're trying to accomplish by doing so.

It not actually dropped, but got moved to `src/bin/common/config.rs` which is a bit of an uncommon place to look for commands, but I follow the reasoning so far..
```rust
fn command_handler<SERVER: DisplayServer>(
        command: &str,
        manager: &mut Manager<Self, SERVER>,
    ) -> bool {
        let mut args = command.split_whitespace();
        let command = args.next().unwrap();
        match command {
            "LoadTheme" => {
                if let Some(path) = args.next() {
                    manager.state.config.theme_setting.load(path);
                } else {
                    log::warn!("Missing file argument to load theme");
                }
                manager.update_for_theme()
            }
            "UnloadTheme" => {
                manager.state.config.theme_setting = Default::default();
                manager.update_for_theme()
            }
            _ => {
                log::warn!("Command not recognized: {}", command);
                false
            }
        }
```

src/bin/leftwm-check.rs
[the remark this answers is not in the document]
cecton: Done in 91b5069

src/bin/leftwm-check.rs
cecton: In the worker this code's logic did make sense because if the file failed to load, the default config will be loaded. So to detect that the config has been loaded properly, we would check that the config loaded is different than the default config.

This logic doesn't make sense here in leftwm-check because the `load_from_file` function returns an error if the file could not be read or could not be parsed. So if the configuration of the user is the same than the default configuration it only means they are using the default configuration. It doesn't mean that the configuration was invalid in any way.

```rust
    if Path::new(&config_filename).exists() {
        let contents = fs::read_to_string(config_filename)?; // Error if file could not be read
        if verbose {
            dbg!(&contents);
        }
        Ok(toml::from_str(&contents)?) // Error if file could not be parsed
    } else {
        bail!("Configuration not found in path"); // Error if file does not exists
    }
```

(Actually... :thinking: we can remove the check that the file exists. This is already covered by `fs::read_to_string()` and should display something like "file not found")

cecton: Done in ceb1a78

cecton:
> It not actually dropped, but got moved to `src/bin/common/config.rs` which is a bit of an uncommon place to look for commands, but I follow the reasoning so far..

Ahhh I think I see it now. The trait Config should return owned types and the manager will store them somehow (or it will be stored by sub components). Then `update_for_theme` should be renamed to `reload_config` and it will re-call the functions of the trait Config to take new instances of everything. This means also that `config` can be removed from the `State`.

src/bin/leftwm-check.rs
mautamu: Ah; thank you for the explanation! Hadn’t tracked with the difference. :)

VuiMuich:
> > It not actually dropped, but got moved to `src/bin/common/config.rs` which is a bit of an uncommon place to look for commands, but I follow the reasoning so far..
> 
> Ahhh I think I see it now. The trait Config should return owned types and the manager will store them somehow (or it will be stored by sub components). Then `update_for_theme` should be renamed to `reload_config` and it will re-call the functions of the trait Config to take new instances of everything. This means also that `config` can be removed from the `State`.

Do I then get it right, that this will open whole new possibilities for partial config reload or my long desired 'actually soft reload'?

cecton:
> Do I then get it right, that this will open whole new possibilities for partial config reload or my long desired 'actually soft reload'?

Actual soft reload yes definitely. Partial.... I think yes. You can already do partial reload using a custom command because you have a &mut on Manager.

lex148:
@cecton Sorry its taken me so long to get to testing this one. I have been unavailable the last two weeks. I have been running this branch all day and all looks great! Code looks good too. I don't have any reason to keep it from being merged other then the one little conflict.

cecton:
Ok let's do this. This PR is big enough anyway xD

AethanFoot:
Might need to revert as external commands seem broken, they don't seem to want to cause an update (e.g. when switching tags with the bar the windows stay the same until you do something internal, e.g. open a window, then it refreshes). Also I've noticed the CustomMargin isn't accepting [x], [x, y], and [x, y, z] vectors correctly, any value not included is set to 0.

Thanks.

AethanFoot:
The issues appear to be that update_windows from window_updater is not called when an external command is called. Would be easiest to add this in the event loop when update_needed is checked, and remove the call from display_event_handler. And the custom margins is just nor implemented completely.
Thanks.

hertg:
> Might need to revert as external commands seem broken, they don't seem to want to cause an update (e.g. when switching tags with the bar the windows stay the same until you do something internal, e.g. open a window, then it refreshes).

Just updated `leftwm-git` on Arch. I can also confirm this issue.

VuiMuich:
as far as I see this, all commands that take argument (except `LoadTheme`) are broken.

btw: for some reason I don't remember anymore, I use `wmctrl` to switch tags in eww.

VuiMuich:
How are your bars switching tags?
After I noticed I am dumb and hat a off-by-one (being on tag_id 3 and sending to tag_num 2)  in the command I used to test, I find every external command to be working.
`leftwm command "SendWorkspaceToTag 0 0"`
`leftwm-command "SendWindowToTag 5"`
`leftwm command "ToggleScratchPad Alacritty"`
`leftwm command "SetLayout Monocle"`
`leftwm command "SetMarginMultiplier 90"`
`leftwm command "RotateTag"
`leftwm command "FocusNextTag"`
Thats all I tried

AethanFoot:
Ye. Any external command, that requires a visual update, wont display its change until window_updater::update_windows is called.
Which was omitted accidentally when removing external_command_handler.
Thanks.

VuiMuich:
> Ye. Any external command, that requires a visual update, wont display its change until window_updater::update_windows is called.
> 
> Which was omitted accidentally when removing external_command_handler.
> 
> Thanks.

Oh right you wrote that earlier. Just wondering then, what is triggering the update for me then, as all worked instantly.

cecton:
Ok will check that out. Sorry for the inconvenience

src/handlers/external_command_handler.rs
```diff
@@ -1,163 +0,0 @@
...
-    state: &impl State,
-    config: &impl Config,
-    theme_loader: &impl ThemeLoader,
-    command: ExternalCommand,
-) -> bool {
-    let needs_redraw = process_work(manager, state, config, theme_loader, command);
-    if needs_redraw {
-        update_windows(manager);
-    }
-    needs_redraw
```
cecton, line 21: Ohhh that code? @AethanFoot Ok will make a PR quickly oops

AethanFoot: Ye, idk it will be better to put it in the event loop, and move the one out of the display event handler

PR #473, opened 2021-09-10
Proper handling of changes to tags in config during soft reload

### LINKED ISSUE #207: Renaming a tag while it is "busy": true makes it "busy": false
Open at least one window (tried alacritty, pavucontrol and ffplay) on a tag (eg. "5").
Rename the tag in config.toml (eg to "debug") and softreload.
The tag appears now empty (leftwm-state reports also "busy": false, as mentioned in title) but the processes stay alive and are back on this tag after reverting the renaming.
Also I always get this Error Message in the Polybar modules-left area, but not sure if related:
leftwm_polybar_error
I was not able to catch this anywhere else then this screenshot, yet. And after a few seconds the tags properly launch.

(BTW: is it possible, that leftwm-state -n is no different then without the flag?)

VuiMuich:
This should handle a change to array of tags in config.toml better during soft-reload.

In case any tags differ from the old state the tag ids get assigned based on ther index in `manager.tags`, if the lenght of this Vec was longer in the old state windows are assigned to index 0.

marked as draft, as there is still a regression with preserving the state if the config was unchanged. But wanted to give a heads up as this might conflict in merging if #472 gets merged first.

Closes #207


VuiMuich:
Hmm, looks like for some reason, the `tag` get lost, while being written to `/tme/leftwm.state`

When I print `log::info!("Current Manager: {:?}", &manager)` in `leftwm::State::save()` and `log::info!("Old Manager: {:?}", &old_manager)` in `leftwm::State::load()` I get 
```
...
tags: [Tag { id: "\u{f46d}",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "\u{e22e}",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "\u{f7a1}",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "\u{e000}",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "ﲾ",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "異",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "\u{f496}",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "ﭮ",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "\u{f454}",
hidden: false,
main_width_percentage: 50,
flipped_horizontal: false,
flipped_vertical: false },
Tag { id: "NSP",
hidden: true,
main_width_percentage: 0,
flipped_horizontal: false,
flipped_vertical: false }],
...
for the current, but for the old:
```
```
...
tags: []
...
```
So far no clue why this is so, though.

VuiMuich:
Oh, I found why: the tags field of the manager (and several fields of the tags them self) where `#[serde(skipp)]`

So I got one step further, now I have still a bug that my mapping of tags isn't working as expected..

VuiMuich:
Ahh, slowly I'm making progress. 
Obviously `old.tags` is the vector of the windows and it `.position()` would always return 0, as currently windows never get assigned a second tag (although they should be able to, right?).

I would need to get the positions of the tags in `old_manager.tags` of course.
Just that gave me a bad loop of crashing worker.
Will have a look later after a little break into Reallive 😁

VuiMuich:
Btw: @lex148 and @AethanFoot (I believe you introduced the `"[serde(skip)]` in tags right?)
Where there any reasons beyond "won't need that info on reload" for excluding tags (and some of their fields) from being serded?

AethanFoot:
I believe those could be removed now, they where there for when the tag was an arc and the fields excluded were types that serde didnt support.
Thanks.

VuiMuich:
I thought I got the crashes, but still can crash it..
Need to look at it on another day..

VuiMuich:
Finally this is going somewhere.

I believe basically this should work now.
Still would like to tidy things a bit. 

If anyone wants to test already, would you please also look out for a possible regression with focusing (especially with layout MainAndDeck and scratch pads).

Speaking of scratch pads, I believe this might be broken with the last commit.

VuiMuich:
I believe other than the two above questions this is good for review.

Also, hope it's ok, I snuck in the clippy pedantic fix in `leftwm-worker.rs`.

BTW: for testing this I used these in `config.toml` and un/commented one or the other (glyphs ar from nerdfonts):
```
tags = ["", "", "", "", "ﲾ", "異", "", "ﭮ", ""]
# tags = ["1", "2", "3", "4", "5", "6", "7", "8", "9"]
# tags = ["1", "2", "3", "4"]
```
then I manually put some windows on random tags and watched what happens during reload (plus I used a bunch of `log::info!()` which I removed now)

Issue #96, opened 2020-05-20
Changing main width doesn't work in fibonacci layout


mathias234:
There seems to be no way to change the width of the main window in a (main and vert) layout. This functionality is present in most tiling WMs

mathias234:
I did an implementation of this on my fork https://github.com/mathias234/leftwm/commit/7ecb1d23bd05ef89a0f79bea23260f8c21b0a4bf

hertg:
Wait, this one isn't closed. It still doesn't work in the fibonacci layout

PR #474, opened 2021-09-12
allow changing main_window size in CenterMain/CenterMainBalanced layouts

hertg:
This PR partly addresses issue #96.

With this update, changing the main-width (`IncreaseMainWidth` and `DecreaseMainWidth`) is now possible in the `CenterMain` layout.

output

Having an ultra-ultra wide 49' inch monitor, this feature is a lifesaver for me :)

**Update 2021/09/12**
I made an update to the code:
- Fixed a pre-existing bug in the `x` positioning of the primary window, when two windows were open on a workspace where `workspace.x()` is **not** `0`.
- Added the same feature to the `CenterMainBalanced` layout.

output

### DIFF
--- a/src/layouts/center_main.rs
+++ b/src/layouts/center_main.rs
@@ -109,7 +109,7 @@ pub fn update(workspace: &Workspace, windows: &mut Vec<&mut Window>, tags: &mut 
         if let Some(first) = iter.next() {
             first.set_height(workspace.height());
             first.set_width(primary_width);
-            first.set_x(workspace.x() + primary_x);
+            first.set_x(primary_x);
             first.set_y(workspace.y());
         }
     }
--- a/src/layouts/center_main.rs
+++ b/src/layouts/center_main.rs
@@ -66,13 +66,13 @@ pub fn update(workspace: &Workspace, windows: &mut Vec<&mut Window>, tags: &mut 
 
     let primary_width = match window_count {
         1 => workspace.width() as i32,
-        _ => (workspace.width() as f32 / 2.0).floor() as i32,
+        _ => ((workspace.width() as f32 / 100.0) * workspace.main_width(tags)).floor() as i32
     };
 
     let secondary_width = match window_count {
         1 => 0,
-        2 => (workspace.width() as f32 / 2.0).floor() as i32,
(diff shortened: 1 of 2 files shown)


hertg:
btw, I'd be happy to implement the same for the `CenterMainBalanced` layout if you'd like that

VuiMuich:
> btw, I'd be happy to implement the same for the `CenterMainBalanced` layout if you'd like that

I guess it would be actually nice to have this in each layout that sports a "dedicated main" window, that's basically ever layout except `EvenHorizontal`,`EvenVertical`,`GridHorizontal` and `Monocle`. TBH have not checked, which ones already have this implemented.

src/layouts/center_main.rs
```diff
@@ -109,7 +109,7 @@ pub fn update(workspace: &Workspace, windows: &mut Vec<&mut Window>, tags: &mut
         if let Some(first) = iter.next() {
             first.set_height(workspace.height());
             first.set_width(primary_width);
-            first.set_x(workspace.x() + primary_x);
+            first.set_x(primary_x);
```
hertg, line 112: This change actually fixes a pre-existing issue where `x` was calculated wrong if two windows were open and`workspace.x()` is not `0`.

In the previous version, `workspace.x()` was added to `primary_x`, but `primary_x`already contained the `workspace.x()` value when two windows were open, see line number 83.

I changed that so that `primary_x` always accounts for `workspace.x()`, so there is no need to add it again here.

hertg:
I just updated my PR with the same fix also applied to the `CenterMainBalanced` layout and updated the PR description.

I don't know exactly what the "Balanced" stands for, but with this feature, it doesn't really differ from the `CenterMain` layout anymore in the sense of column widths. But this shouldn't matter because you can now dynamically change them to be approximately equal in both layouts.

The only difference is now that `CenterMain` appends new windows only to the right stack in an even vertical pattern, where `CenterMainBalanced` adds windows to the left and right stack in a fibonacci pattern. 

I think it's great to have those two options, but the "Balanced" in the layout name might not make as much sense now as it did before. :smile:

lex148:
@hertg this is great! awesome work

PR #479, opened 2021-09-12
add commands to cycle through tags

the-other-autumn:
adds two commands:

NextTag which cycles through all tags and loops when it reaches the last tag
` 1->2->3->4->5->6->7->8->9->1 `

PrevTag which cycles through all tags in reverse and loops when it reaches the first tag
`9->8->7->6->5->4->3->2->1->9`

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -10,6 +10,8 @@ pub enum Command {
     ToggleScratchPad,
     ToggleFullScreen,
     GotoTag,
+    NextTag,
+    PrevTag,
     FloatingToTile,
     MoveWindowUp,
     MoveWindowDown,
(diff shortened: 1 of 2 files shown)


mautamu:
Hmm, is this the same as FocusNextTag and FocusPreviousTag?

the-other-autumn:
whoops, didn't notice that command

mautamu:
It's all good mate, do you mind sharing where you looked? Would probably be helpful to add some documentation. Want to make leftwm as friendly and easy to understand as possible.

the-other-autumn:
checked through the wiki and the default config file a bit, must have just missed it, 
that's what i get for doing that while im tired i suppose

PR #480, opened 2021-09-13
Make path to state_file configurable

VuiMuich:
First of, as I once again butchered my git tree this PR depends on #473 to be merged. Sorry.

In order to set a custom path just add `state_file = /home/username/.cache/leftwm.state` to your `config.toml`.
When none is present I default to `pub const STATE_FILE: &str = "/tmp/leftwm.state" in `common/state.rs`. I felt I would like to keep this here for clarity.

I am still working on a bug that relative paths cant be handled, so gonna mark this 'draft' for now.

Also not sure what to think about the `&dyn Config` with the `State::save()` and `State::load()`? There is a warning, that this will be needed in Rust2021, but it feels like this is not the proper way how to handle this then, at least to me. So, if anyone has a hint on how to handle this properly, would be very welcome.

### DIFF
--- a/src/bin/common/config.rs
+++ b/src/bin/common/config.rs
@@ -31,6 +31,7 @@ pub struct Config {
     pub focus_behaviour: FocusBehaviour,
     pub focus_new_windows: bool,
     pub keybind: Vec<Keybind>,
+    pub state_file: Option<String>,
 }
 
 #[must_use]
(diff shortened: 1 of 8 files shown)


VuiMuich:
Would we be ok with just supporting absolute paths for `state_file`?
As far as I get it, there is currently no way in the std library to handle `$HOME` and `~`, except `std::env::home_dir` which is deprecated. There seem to be intentions to bring back proper home directory handling in std, but no prospect when this might happen.
Cargo and rustup us the `home` crate, but I am always hesitant to introduce new dependencies.

Also in the case of the `state_file` it makes little sense to rely on `BaseDirectories` like writing the default config does, as this would limit the `state_file` to exist in XDG_ standad directories, I guess.

VuiMuich:
@lex148 may I ask for your opinion if support for relative paths would be mandatory?

VuiMuich:
I guess at this point it's easier to start over with this from a fresh branch, then fixing all conflicts.

Issue #163, opened 2021-01-25
Dvorak keyboard does not trigger the commands


Arash-Afshar:
Change the layout using `setxkbmap dvorak`, then triggering a command for opening dmenu does not work. I tried both `Mod4 + p` (pressing the physical key with label `r` on it) and `Mod4 + l` (the physical key with the label `p` on it). Neither worked.

btw, I am interested in helping fix this but I'd need some help with getting started. If you can point me to a location in the code and how to view the debug logs, I'll get started on it :)

lex148:
Thats great, its awesome you want to help.
Here is where I would start: [https://github.com/leftwm/leftwm/blob/master/src/command_builder.rs](https://github.com/leftwm/leftwm/blob/master/src/command_builder.rs)
This is where keyboard events get translated into actions for LeftWM to act on. 

Debugging a window manager can be rather tricky. There are better ways if you want to really do a lot of work in LeftWM, but simplest is probably the following. 

1) symlink the executables from `target/debug` to somewhere that everyone can access like `/usr/bin` (see the manual setup proccess )
1b) you can verify with `which leftwm-worker` 
2) build with the journald enabled `cargo build --features=fournald` 
3) set the log level you want to see in `src/bin/leftwm-worker.rs` LINE: 165  default is only logging to info

You are now all ready to go:
You can watch the logs with `journalctl -f`
To see your changes make a change in code, compile with `cargo build --features=fournald`
The worker process auto restarts, so just kill it and you will see the latest compiled version. 
`pkill leftwm-worker` 

If your keybindings are working :) `Mod+Shift+R` keeps the state you are in so its a little better then just using pkill

Good luck and have fun!

Arash-Afshar:
So it turned out it works with dvorak keyboard. The problem was "changing" the keyboard layout. So when I change the layout and kill the `left-worker` it works fine, because it picks up the current layout. I'll look into why changing layout is problematic when I have a chance.

oukemi:
I also have the same problem with a bepo keyboard.. .

VuiMuich:
if this issue still persists after the fix in #481 please reopen

seungjin:
I am a Colemak layout user. 
Leftwm does not pick up my colemak layout - means it does not trigger command (ie. modkey+p dose not do any). 
If I do `setxkbmap -layout us -variant colemak` and `leftwm-command SoftReload` in terminal, it makes my colemak layout keys binds work. 

[...]

PR #481, opened 2021-09-14
Fix broken keybinds when keyboard layout changes

SamClercky:
Hi,

Lately I changed my keyboard layout to `dvorak`, but Leftwm wouldn't pick up the changes resulting in me not being able to use the keybindings. The symptoms and temporary solution, `pkill lefwm-worker` on startup, were similar to issue #163 (this pull request possibly fixes this issue).

**The fix:**
- Leftwm did not refresh the key mapping on a `MappingNotify` event -> added an extra `match` case
- After the remapping, the keys needed to be regrabbed
- Some plumbing to get all necessary information together for the regrabbing

Kind regards,

SamClercky

Ps. Some useful links:
- [Stackoverflow xlib keyboard changes](https://stackoverflow.com/questions/35569562/how-to-catch-keyboard-layout-change-event-and-get-current-new-keyboard-layout-on)
- [Example implementation in sxhkd](https://github.com/baskerville/sxhkd/blob/fe241b0d2d70c9c483b23cf3cd14f1383f0953a2/src/sxhkd.c#L260)
- Useful xlib documentation: [`XRefreshKeyboardMapping`](https://tronche.com/gui/x/xlib/utilities/keyboard/XRefreshKeyboardMapping.html) and [`MappingNotify`](https://tronche.com/gui/x/xlib/events/window-state-change/mapping.html)

### DIFF
--- a/src/display_servers/xlib_display_server/event_translate.rs
+++ b/src/display_servers/xlib_display_server/event_translate.rs
@@ -111,7 +111,7 @@ fn from_mapping_notify(raw_event: xlib::XEvent, xw: &XWrap) -> Option<DisplayEve
     if event.request == xlib::MappingModifier || event.request == xlib::MappingKeyboard {
         // refresh keyboard
         log::info!("Updating keyboard");
-        xw.refresh_keyboard(&mut event).unwrap();
+        xw.refresh_keyboard(&mut event).ok()?;
 
         // Reload keybinds
         Some(DisplayEvent::KeyGrabReload)
--- a/src/display_servers/xlib_display_server/event_translate.rs
+++ b/src/display_servers/xlib_display_server/event_translate.rs
@@ -22,6 +22,9 @@ impl<'a> From<XEvent<'a>> for Option<DisplayEvent> {
             // new window is created
             xlib::MapRequest => from_map_request(raw_event, xw),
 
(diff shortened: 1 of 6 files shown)


src/display_servers/xlib_display_server/event_translate.rs
```diff
@@ -103,6 +106,20 @@ fn from_map_request(raw_event: xlib::XEvent, xw: &XWrap) -> Option<DisplayEvent>
     Some(DisplayEvent::WindowCreate(w, cursor.0, cursor.1))
 }
 
+fn from_mapping_notify(raw_event: xlib::XEvent, xw: &XWrap) -> Option<DisplayEvent> {
+    let mut event = xlib::XMappingEvent::from(raw_event);
+    if event.request == xlib::MappingModifier || event.request == xlib::MappingKeyboard {
+        // refresh keyboard
+        log::info!("Updating keyboard");
+        xw.refresh_keyboard(&mut event).unwrap();
+
```
AethanFoot, line 115: Instead of unwrap you should be able to use ?, e.g ```xw.refresh_keyboard(&mut event)?```.
As unwrap will lead to panics if there is an error.

lex148:
This looks great! I will give it a test to day and then merge.

Issue #488, opened 2021-09-22
Intellij pop-up breaks tile mode


lkadalski:
**Describe the bug**
I'm using Intellij to work on Java projects. The unwanted situation is when I have two Intellij instances open at the same workspace and trigger popup window in one on them(rename a file). Then parent Intellij of that pop-up is taking whole screen and ignores the other (tiled) one. Even If I move the infected intellij window to other workspaces and back it ignores tiled windows on that workspace. It behaves like it is the only one tile window in workspace. I tried to move it to floating mode and back but once moved to floating mode it stays floated also. 
I have a feeling that leftwm mark parent intellij as pop-up, but I can't confirm.
Other programs work as expected. 

**To Reproduce**
1. Open Two instances of Intellij on one workspace
2. In one of the instances use pop-up window for instance: R click on a file ->Refactor -> Rename file. 
3. Pop-up shows and parent intellij is behaving like it's only one tile in workspace(error)
4. When tried to move to floating mode(it's working) then cannot return to tiled mode.(2nd error)

**Expected behavior**
After triggering pop-up, two instances of intellij are visible as two tile windows in the same workspace. Parent Intellij of pop-up acts as normal tile window. It can be moved to other workspaces as tiled one and it's recognized by leftwm as tiled one.
It can be change freely to flaoting mode and back to tiled one.  

**Environment:**
 - ArcoLinux
 - Output of `leftwm-check`:
 ```
 :: LeftWM version: 0.2.8
:: LeftWM git hash: 3c80a4a
:: Loading configuration . . .
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
```

Do you need any other information?

hertg:
I planned on opening the exact same issue :smile:. I did a little research ahead to find out if it's something I could address with a PR myself. I had the exact same issue in dwm and it seems to stem from something that JetBrains messed up. I saw reports of the exact same issue for WebStorm, another JetBrains product.

Here are some resources I found:

- [Prevent my IDE to become floating in Awesome WM](https://superuser.com/questions/999486/prevent-my-ide-to-become-floating-in-awesome-wm) (superuser.com)
- [[DWM] Java App put itself in floating mode when child window opens.](https://www.reddit.com/r/suckless/comments/k4j89d/dwm_java_app_put_itself_in_floating_mode_when/) (reddit.com)
[...]

PR #491, opened 2021-09-22
Handling Normal window type as non transient

lkadalski:
This is answer for problem with Intellij. Issue explained in [ticket](https://github.com/leftwm/leftwm/issues/488)

Signed-off-by: lkadalski <kadalski.lukasz@gmail.com>


AethanFoot:
I have also found another place where this would fix, which if with timeshift (or probably any other app which does the same). This issue is that timeshift creates a new normal window and hides the original window when going to the settings page (instead of dialog, and which isn't very twm friendly). This cause a floating window which takes the whole screen as the parent is hidden so our sizing doesnt work properly, with no parent accessible. Anything that works like this is just bad app design, but this change should prevent us creating any extra side effects.

However I would say it would be better to address in window_handler, where we actually set the window floating. As knowing the windows parent should still be stored in case we want to use it.

Thanks.

lex148:
@lkadalski this is great. I am out of the office until the 1st, so its going to take be just a bit to test this but cant wait to get it merged. little fixes like this really make such a big difference.  

If anyone else gets a chance to test please give it a thumbs up so I know. I'm a little unavailable this week.

lkadalski:
Thanks, this is mainly done by research work of @hertg
Do you think it should be moved to other component as @AethanFoot suggested ?

hertg:
I tested it on my machine and can confirm that it resolves the issue :tada: 
Are there any potential side effects of this update? I am happy to test those out as well.

@lkadalski 
Can you update your branch to the latest upstream? I would be willing to switch my local installation to your branch and see if something comes up in my day-to-day work.

lkadalski:
Updated @hertg 
I think that someone mentioned in your research that is bad design when one Normal window creates new normal window as transient. It'd rather dialog etc.
But maybe **there is** some program which acts like this and it's expected. I'm not aware of any of those.
Currently I'm using Brave, Insomnia, Intellij, Alacritty, Nemo(sometimes) in many instances and don't have any unexpected behavior with this setup.
Please don't trust me though, and test your workflow :)

Issue #483, opened 2021-09-16
simple patch for scratchpad position


tramhao:
Probably we need more control of the size and position of scratchpad, and here is my patch for a scratch pad on right bottom which is my personal preference. I'm not sure if there is any plan on this issue so I posted it here. If you guys have any suggestions, please kindly let me know. Since this is quite personal, I'll not submit a PR until it's considered more seriously.

```
diff --git a/src/handlers/window_handler.rs b/src/handlers/window_handler.rs
index 03ccbe2..4903302 100644
--- a/src/handlers/window_handler.rs
+++ b/src/handlers/window_handler.rs
@@ -85,7 +85,7 @@ fn setup_window(
 
         if is_scratchpad {
             window.set_floating(true);
-            let new_float_exact = ws.center_halfed();
+            let new_float_exact = ws.right_bottom();
             window.normal = ws.xyhw;
             window.set_floating_exact(new_float_exact);
         }
diff --git a/src/models/workspace.rs b/src/models/workspace.rs
index 3c4dc40..69e1351 100644
--- a/src/models/workspace.rs
+++ b/src/models/workspace.rs
@@ -289,6 +289,10 @@ impl Workspace {
         self.xyhw_avoided.center_halfed()
     }
 
+    pub fn right_bottom(&self) -> Xyhw {
+        self.xyhw_avoided.right_bottom_corner()
+    }
+
     pub fn update_avoided_areas(&mut self) {
         let mut xyhw = self.xyhw;
         for a in &self.avoid {
diff --git a/src/models/xyhw.rs b/src/models/xyhw.rs
index 4f4f9ce..e612da2 100644
--- a/src/models/xyhw.rs
+++ b/src/models/xyhw.rs
@@ -268,6 +268,17 @@ impl Xyhw {
         .into()
     }
 
+    pub fn right_bottom_corner(&self) -> Xyhw {
+        XyhwBuilder {
+            x: self.x + self.w - (self.w / 2 - self.w / 20),
+            y: self.y + (self.h / 5), //- (self.h / 4),
+            h: (self.h - self.h / 5),
+            w: (self.w / 2 - self.w / 20),
+            ..XyhwBuilder::default()
+        }
+        .into()
+    }
+
     pub fn center_relative(&mut self, outer: Self, border: i32, change: Option<XyhwChange>) {
         // If this is a Splash screen, for some reason w/h are sometimes not sized properly. We
         // therefore want to use the minimum or maximum sizing instead in order to center the
```

[...]

PR #493, opened 2021-09-24
Scratchpad position

tramhao:
can read from config file the below section:

[[scratchpad]]
name = "st"
value = "st"
width = 51
height = 80
x = 49
y = 20

these values are relative numbers so 51 means 51% of workspace width.


src/models/xyhw.rs
```diff
@@ -268,6 +269,38 @@ impl Xyhw {
...
+        XyhwBuilder {
+            x: self.x + self.w * x_sane / 100,
+            y: self.y + self.h * y_sane / 100, //- (self.h / 4),
+            h: self.h * height_sane / 100,
+            w: self.w * width_sane / 100,
+            ..XyhwBuilder::default()
+        }
+        .into()
+    }
+
```
hertg, line 303: Having something scratchpad specific in `Xyhw` seems a bit weird, Maybe it should be the other way around?
Scratchpad can know about Xyhw, but Xyhw probably shouldn't have to know anything about Scratchpad.

src/config/scratchpad.rs
```diff
@@ -4,4 +4,29 @@ use serde::{Deserialize, Serialize};
...
+    // relative y of scratchpad, 25 means 25% of workspace y
+    #[serde(default)]
+    pub y: i32,
+}
+
+impl Default for ScratchPad {
+    fn default() -> Self {
+        Self {
+            name: "st".to_string(),
+            value: "st".to_string(),
```
hertg, line 25: Should `st` really be the default?

src/models/xyhw.rs
tramhao: Because I still don't have a whole picture of the project, would you mind let me know the correct place for this? I don't want to screw the structure of the project.

src/config/scratchpad.rs
tramhao: That's a default which was not called before. Actually without this could be better. When no scratch pad configured, simply don't show it. What do you think?

src/models/xyhw.rs
tramhao: I checked that currently only config have a scratchpad module. But it's only for configuration.

hertg:
Is there an open issue you are addressing with this PR? Or could you elaborate a little bit on the pain-points you have and wish to address with this feature? I don't use the scratchpad too often, and when I do, I usually resize and move it around quite a bit, that's why it isn't as obvious to me as it may be to you :)

I'm fairly new to the project myself, so don't worry too much about my comments. I just wanted to point some things out in the code that I felt were a bit odd. It's really just a hint to the more experienced contributors who will have to review your PR. :smile: 

Btw. I just recently added the [Size](https://github.com/leftwm/leftwm/blob/master/src/models/size.rs) enum, you could also use that if you want to allow absolute values too.

tramhao:
Any comments are helpful and thanks for your reply. I mentioned it in https://github.com/leftwm/leftwm/issues/483 .
Basically my preference is open a scratchpad at right_bottom corner, with width of 80 columns. I use it a lot when I use dwm. After switching to leftwm, I cannot configure it so here comes the PR. The community is very active and welcoming.

tramhao:
> Btw. I just recently added the [Size](https://github.com/leftwm/leftwm/blob/master/src/models/size.rs) enum, you could also use that if you want to allow absolute values too.

Seems a good resolution for percentage. Let me dig into it.

hertg:
> Seems a good resolution for percentage. Let me dig into it.

The percentage is represented as `0.5 => 50%`, just in case that trips you up.
I'm still not 100% sure if that's really how it should behave, see my comment here: https://github.com/leftwm/leftwm/pull/482#discussion_r712536542. But since it was already merged, I guess you can use it. Just be aware that there is a possibility that the behavior of `Size` could change a little bit.

src/models/xyhw.rs
hertg: > Because I still don't have a whole picture of the project

We're on the same boat :wink: 
Unfortunately, you'll have to wait for another contributor to advise you on that, I'm not yet comfortable enough with the code to give you a clear suggestion.

VuiMuich:
I believe we discussed this in the comments of a somewhat related issue as well, just can't remember it right now.
Maybe it was one regarding tilda/yakuake.. will look it up later.

VuiMuich:
Hmm, I cant find the comment I was talking about earlier.. might have been about default sizes for floating windows, but i cant remember unfortunately.

But back on your PR:
This is a great first I'd say, thanks for your contribution and 'Welcome Adventurer!'

I'd suggest to make the fields `width`, `height`, `x` and `y` of type `Option<i32>` and handle this in `common::config` as this needs to be available each time, the scratchpad window is created. But when it gets resized and positioned, it should keep this new values when getting hidden until it is destroyed.
For a more in-depth recommendation, where to exactly set the defaults I might have a chance to have a proper look on Sunday.

tramhao:
Thanks for your reply. I change scratchpad x y height width to option. But in config, I didn't find the proper place to handle it. I handled it in xyhw.rs and works. About the preserve of adjusted size, it's now preserved as only when create window it's using the configuration data.

AethanFoot:
I can't really look properly as i am on my phone. However you should be able to move scratchpad_size_position (maybe scractchpad_xyhw flows better?) into window_handler without any issues, as you should be able to get the workspaces xyhw publicly.

Other than this this looks like a good solution!
Thanks.

tramhao:
> I can't really look properly as i am on my phone. However you should be able to move scratchpad_size_position (maybe scractchpad_xyhw flows better?) into window_handler without any issues, as you should be able to get the workspaces xyhw publicly.
> 
> Other than this this looks like a good solution!
> Thanks.

Thanks for the reply and I change it to window_handler.rs. Looks cleaner now.

src/handlers/window_handler.rs
```diff
@@ -307,3 +317,41 @@ pub fn get_next_or_previous(manager: &mut Manager, handle: &WindowHandle) -> Opt
...
+    let scratch_pad_x = scratch_pad.x.unwrap_or(25);
+    let scratch_pad_y = scratch_pad.y.unwrap_or(25);
+    let scratch_pad_height = scratch_pad.height.unwrap_or(50);
+    let scratch_pad_width = scratch_pad.width.unwrap_or(50);
+
+    let x_sane = if (10..90).contains(&scratch_pad_x) {
+        scratch_pad_x
+    } else {
+        25
+    };
```
VuiMuich, line 332: Not an actual critique on this PR, but asking to improve my own understanding of code quality/best practices/good style:
Would it be good to extract this to
```rust
fn sane_dimension(config_value: Option<i32>, default_value: i32 {
    match config_value {
       Some(dim) if (10..90).contains(&dim) => dim,
       _ => default_value,
    }
}
```
and then call `let x_sane = sane_dimension(scratch_pad_x, 25)` and so on?
(Writing on mobile, so please excuse if this is actually broken code 🤷🏼‍♂️, not entirely sure about that `Some() if`)

tramhao: Great idea and thanks for your help! I extracted sane_dimensions and now even more cleaner.
Meanwhile, I tried with impl Default for Config before. But the problem is, when no scratchpad is configured, the default will still work and get a default value. Probably it's best to show nothing when no scratchpad is configured.

tramhao: Because I have to face a situation: when people upgrade, they'll not change configuration file unless necessary. Thus current solution can work with the situation when scratchpad is configured, but no xyhw is provided. Here a sane defaults will be provided, same as behavior before the patch.

VuiMuich: Well I believe this is exactly the purpose of the default config and the reason why I suggested `Option<i32>` so the parser can handle backwards compatibility.
I'm not sure if this wold work but I'd try to put `scratchpad.x: 50` and so on in the `Self {}` element in `impl Default for Config`

VuiMuich:
Sorry for another idea, but it would be actually nice if you could use `Option<Size>` which was introduced by @hertg in #482 so it's up to the user to use pixel values or percentages.
Thanks a lot for your efforts!

tramhao:
> Sorry for another idea, but it would be actually nice if you could use `Option<Size>` which was introduced by @hertg in #482 so it's up to the user to use pixel values or percentages.
> Thanks a lot for your efforts!

Met a problem. enum Size cannot derive eq and hash, because float is used. Is there a proper way to handle it?

hertg:
> Met a problem. enum Size cannot derive eq and hash, because float is used. Is there a proper way to handle it?

You could turn the `Size` into an absolute `i32` by using `into_absolute(<whole>).floor() as i32`. You probably need to turn any relative size into an absolute pixel size anyway in order to compare them? But I may be completely missing the point :smile:

src/handlers/window_handler.rs
VuiMuich: That's what I thought might work, but now that I rethink it I guess you are right, that this won't work as intended.
My idea was to define the sane defaults globally so when the config for the scratchpad is set each scratchpad can either take the values from the parser or fallback.
Setting this default values would naturally belong in the above section I guess.
Unfortunately I won't make it to may dev machine anymore today, so I just can look at this on mobile.
If you feel you're stuck here I guess leaving a TODO note in the `Default for Config` to refactor this at a later point and where the default is currently applied, might be ok for now.

As far as I read into branch 4 of @cecton's libification rework a few things might get quite a bit clearer along the way.

tramhao:
> Sorry for another idea, but it would be actually nice if you could use `Option<Size>` which was introduced by @hertg in #482 so it's up to the user to use pixel values or percentages.
> Thanks a lot for your efforts!

I change it to Option<Size>. But my implementation is not correct. Should I pass the xyhw of workspace into sane_default fn? Otherwise I cannot get a sane xyhw if the user specified a pixel value. Please kindly check below:
```
fn sane_dimension(config_value: Option<Size>, default_value: f32) -> i32 {
    match config_value {
        Some(size) => match size {
            Size::Percentage(percentage) if (0.1..0.9).contains(&percentage) => {
                size.into_absolute(100.0) as i32
            }
            Size::Percentage(_) => (default_value * 100.0) as i32,
            Size::Pixel(dim) if (10..90).contains(&dim) => dim,
            Size::Pixel(_) => (default_value * 100.0) as i32,
        },
        _ => (default_value * 100.0) as i32,
    }
}
```

VuiMuich:
I am really sorry I am causing you such headaches..

I believe on thing you could simplify would be to lose `* 100` here and `/ 100` in the xyhw_builder, right?
And I am not sure if the match arms `Size::Percentage(_)` and `Size::Pixel(_)`, if it was possible to flatten this to a single match not a cascaded one?

tramhao:
> I am really sorry I am causing you such headaches..
> 
> I believe on thing you could simplify would be to lose `* 100` here and `/ 100` in the xyhw_builder, right?
> And I am not sure if the match arms `Size::Percentage(_)` and `Size::Pixel(_)`, if it was possible to flatten this to a single match not a cascaded one?

No, your help is very appreciated. I've learned a lot from your guide, and thanks so much:)
I'm also thinking the same thing and got it working this time. Now it's possible to mix percentage with pixel in the configuration, and I think it's good, because for x,y we prefer pixel, but for h/w percentage is better. I'll push it in a minute.

tramhao:
> I am really sorry I am causing you such headaches..
> 
> I believe on thing you could simplify would be to lose `* 100` here and `/ 100` in the xyhw_builder, right?
> And I am not sure if the match arms `Size::Percentage(_)` and `Size::Pixel(_)`, if it was possible to flatten this to a single match not a cascaded one?

About the 100 * and /100, it's because I need to return an i32 value. If less than 1, it'll become 0. If I return the f32, in later calculation I need to convert other value then convert it back. Both methods are not perfect.

tramhao:
> > I am really sorry I am causing you such headaches..
> > I believe on thing you could simplify would be to lose `* 100` here and `/ 100` in the xyhw_builder, right?
> > And I am not sure if the match arms `Size::Percentage(_)` and `Size::Pixel(_)`, if it was possible to flatten this to a single match not a cascaded one?
> 
> About the 100 * and /100, it's because I need to return an i32 value. If less than 1, it'll become 0. If I return the f32, in later calculation I need to convert other value then convert it back. Both methods are not perfect.

I put the calculation into sane_dimension so now it should be ok. For the default implementation, I still cannot figure out a way to do it. Thanks.

VuiMuich:
> About the 100 * and /100, it's because I need to return an i32 value. If less than 1, it'll become 0. If I return the f32, in later calculation I need to convert other value then convert it back. Both methods are not perfect.

Ohh, right.
Thanks for all your efforts again. This looks good now. And don't worry about the default too much, maybe it's just me having a silly idea that actually won't work. As a last little request. could you add to the TODO comment, that default sizes currently are set in the `sane_dimension` function? Thanks.

tramhao:
> And don't worry about the default too much, maybe it's just me having a silly idea that actually won't work. As a last little request, could you add to the TODO comment, that default sizes currently are set in the `sane_dimension` function? Thanks.

Of course.

PR #494, opened 2021-09-26
fix #397

### LINKED ISSUE #397: [Feature Request] Sharing single Windows for screen share
A the title says, I'd like to be able to share single Windows for screensharing. I would also try to make this myself if this is planned to be implemented.

I use leftwm-git from the AUR right now

AethanFoot:
Fixes #397 hopefully.
Thanks.

Edit:
Just for those curious this sets the WM_STATE (different then _NET_WM_STATE) to "NormalState" (the 1), which is a constant in Xutil.h (however there are no binding for Xutil.h for rust). An example of setting WM_STATE can be found in dwm.c setclientstate function.


hertg:
I just tested this on my machine, and it works great :tada: 
Thanks a lot for fixing that so fast! :)

PR #500, opened 2021-10-02
Fix margin parser

cecton:
As mention by @AethanFoot https://github.com/leftwm/leftwm/pull/472#issuecomment-932315027 I accidentally broke the parsing behavior of Margins.

Existing behavior before my change:
 - no values: gives a warning, assume 10px on all side
 - 1 value: assume that value on all side (match CSS)
 - 2 values: assume the first value for top and right and the second value for bottom and left :warning: This is contradicting CSS where 2 values means: first value for top and bottom and second value for right and left
 - 3 values: assume the first value for top, second value for right, third value for bottom and left :warning: This is contradicting CSS where 3 values means: first value for top, second value for right and left, third value for bottom
 - 4 values: top, right, bottom, left (match CSS)
 - more than 4 values: gives a warning, assume 4 values

What I implemented:
- no values: assume 0 on all side, no warning ❌
- 1 value: assume that value on all side (match CSS) :heavy_check_mark: 
- 2 to 4 values: assign value to top, right, bottom and left, use 0 if one is missing ❌
- more than 4 values: does not parse (fatal) ❌

This PR restore the original behavior but I think we should modify it to comply with CSS so it will feel more intuitive.

Let me know your opinion.

### DIFF
--- a/src/bin/common/theme_setting.rs
+++ b/src/bin/common/theme_setting.rs
@@ -70,15 +70,39 @@ impl std::convert::TryFrom<CustomMargins> for Margins {
                 left: size,
             }),
             CustomMargins::Vec(vec) => {
-                if vec.len() > 4 {
-                    Err("too many values for margin (max is 4)")
-                } else {
-                    Ok(Self {
-                        top: *vec.get(0).unwrap_or(&0),
-                        right: *vec.get(1).unwrap_or(&0),
-                        bottom: *vec.get(2).unwrap_or(&0),
-                        left: *vec.get(3).unwrap_or(&0),
-                    })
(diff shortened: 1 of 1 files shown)


mautamu:
I believe CSS-like was the original goal when it was implemented; would probably be good to do so now.

cecton:
> I believe CSS-like was the original goal when it was implemented; would probably be good to do so now.

Ok so I adapted the 2 and 3 input values to match CSS better.

I also changed the 0 or more than 4 values to fail instead of being permissive. If you don't like this change, feel free to revert the last commit.

PR #501, opened 2021-10-02
clippy 0.1.57

AethanFoot:


AethanFoot:
Can someone check 3d96e7c, I believe this should work the same (if im remembering my demorgan's laws correctly) given https://rust-lang.github.io/rust-clippy/master/index.html#if_then_panic.
Thanks.

mautamu:
Appears so;
```rust
fn main() {

assert_eq!(!(!true && false), true || !false);
assert_eq!(!(!true && true), true || !true);
assert_eq!(!(!false && true), false || !true);
assert_eq!(!(!false && false), false || !false);
}
```
passes.

PR #508, opened 2021-10-05
Make `systemd-logging` default

VuiMuich:
As error messages were greatly improved recently (#472 iirc) and the need to get a common log feedback from users during support requests increases as well, I want to suggest to make systemd-logging the default.

For users who don't want systemd logging, or run a different init system we have the command `cargo build --release --no-default-features --features=cli` mentioned in the `README.md`.


hertg:
Does this mean that users with different init systems will get a compilation error when using `cargo build` without any flags? Or does it just not log anywhere, as it does now for systemd users? I think the first would be a controversial step for anyone using an alternative init system, the latter seems fine.

mautamu:
I just build it on Artix (runit) with `--features journald` without any issues.

VuiMuich:
> I just build it on Artix (runit) with `--features journald` without any issues.

Is it also running well?
I didn't expect it to fail during build, but possibly panic if worker can't drop its messages.
Probably this is taken care properly by the systemd logging crate though.

lex148:
I am a bit torn on this one. Absolutely see the desire here to have better logging. Our logs are extremely verbose. My concern is we might fill up someones hard drive. Has anyone calculated the rate at which our log file grows? Just seams like an important think to check if we are defaulting people into this feature.

lex148:
> What do you think about `leftw conmand "SetLogLevel Debug"` with options `Off`, `Debug` and `Error`. What else would make sense here?

oh yeah. I really like that Idea :)

VuiMuich:
With
```shell
leftwm >>> rg info! .
./src/bin/leftwm-worker.rs
10:    log::info!("leftwm-worker booted!");
25:        Ok(_) => log::info!("Completed"),

./src/display_servers/xlib_display_server/event_translate.rs
113:        log::info!("Updating keyboard");
```
I'd say `log::info!` should even print in `off`, this way we keep the current behaviour.

Agree on excluding `trace`. If really needed there is still `RUST_BACKTRACE` env variable.

hertg:
Just wanted to mention that journald has the following priority levels ("log levels"), as stated in [rfc5424](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1).

```
              0       Emergency: system is unusable
              1       Alert: action must be taken immediately
              2       Critical: critical conditions
              3       Error: error conditions
              4       Warning: warning conditions
              5       Notice: normal but significant condition
              6       Informational: informational messages
              7       Debug: debug-level messages
```

I'm not sure if the verbose logging of LeftWM would actually be a problem, as long as it's on the `debug` level, I think the default setting for systemd is set to log-level 6 (ie. `info`). So it probably ignores the debug logs already, unless configured otherwise. You can check that on your system with the following command:
```
systemctl log-level
```

VuiMuich:
looks like auto-closing this with #821 didn't work. well...

PR #510, opened 2021-10-06
feature(autostart): follow the xdg-autostart spec

hxzhao527:
Some applications like `fcitx5` put desktop file in `/etc/xdg/autostart`, not user's home.

change:
1. add doc
2. fix autostart  directories
3. change `XDG_CURRENT_DESKTOP` to `LeftWM` (case-sensitive)

English is not my mother tongue. If there are errors in the document or a better description, I am happy to update it. :smile:

### DIFF
--- a/leftwm.desktop
+++ b/leftwm.desktop
@@ -4,3 +4,4 @@ Name=LeftWM
 Comment=A window manager for the adventurer
 Exec=leftwm
 Type=Application
+DesktopNames=LeftWM


VuiMuich:
Looks good to me, great to see some doc (Memo to self: make it a habit too :wink:). 
Only thing I was wondering, that `cargo fmt` didn't complain about the length of some doc comments. Maybe you could wrap them manually to avoid overly long lines?
Thanks for another great PR.

src/utils/child_process.rs
```diff
@@ -18,19 +17,36 @@ use xdg::BaseDirectories;
...
+            .map(|xdg_dir| {
+                xdg_dir
+                    .list_config_files_once("autostart")
                     .iter()
+                    .filter(|path| {
+                        if let Some(ext) = path.extension() {
+                            return ext == "desktop";
+                        }
+                        false
+                    })
```
AethanFoot, line 49: Here you could do something like;
```
path.extension().unwrap_or('') == "desktop"

PR #511, opened 2021-10-07
Document and organise xwrap.rs (Questions)

AethanFoot:
Tag @VuiMuich, @mautamu, @lex148, @hertg. Opening now as have some questions. I was going to organise all the getters together and setters together, etc (and maybe some refactoring to reduce code). However I think maybe splitting them into different files would make it easier to find and understand what you want easier (which is the goal of this PR, to make the xlib side easier for new users tackle), which would all be in a new directory called xwrap (where xwrap.rs is currently).

This is part of a string of documentation I hoping to tackle (anyone can help), where I am following https://doc.rust-lang.org/rust-by-example/hello/comment.html. We should also try and add as many links to relevant documentation that new users can go to to understand that thing better. E.g. I thinking of putting a link to the xlib documentation for each given function called in xwrap.

Many Thanks.


VuiMuich:
This some great endeavour and it looks really nice!

AethanFoot:
I feel this is ready now, however if there are any addition/change suggestions please let me know. I would suggest some extra testing in case of regression if I made an error reducing the functions.
Thanks!

PR #513, opened 2021-10-09
Split leftwm in 2 crates: leftwm and leftwm-core

cecton:
Hello @lex148 :wave: 

So this is finally the goal: separating core logic from binary logic. The core logic will allow other tiling manager to be based on leftwm. It will also allow people to make their own specific tiling manager (this is my personal intention).

With this change you can see that the optional dependencies of "cli" are now not optional anymore and moved to the crate `leftwm`. On the other hand, `leftwm-core` will see its dependencies reduce: all cli's feature dependencies but also anyhow, toml, liquid, shellexpand. The more reduced, the better (for projects who use leftwm but will make different choices). This will improve in future PR but I want to keep this one strictly about the splitting.

The "common" module has now become the `leftwm`'s lib. This allows easy access and resolve a few "dead_code" warnings.

### The downside

The split means that you will need to maintain and publish 2 crates on crates.io. You can use [`cargo-release`](https://crates.io/crates/cargo-release) to help you on that. I personally usually do the command `cargo release --workspace --no-dev-version minor` (or major or patch). I suggest to keep the version of the 2 crates synchronized as it is easier to maintain but you can have different version if you want. If you want you can add me to owner and I will do the first publish of everything for you.


AethanFoot:
Im not sure why the test is having issues, is it possible that the first pipe isnt always dropping correctly or in time before the second pipe is opened? If I remember correctly it fails on the second one.

cecton:
> Im not sure why the test is having issues, is it possible that the first pipe isnt always dropping correctly or in time before the second pipe is opened? If I remember correctly it fails on the second one.

I just created #514 for that. There is some timed logic in the test if I remember... so it's possible the CI machines is just not fast enough for the time we allocated in test. Further discussion can go in #514.

AethanFoot:
>  If you want you can add me to owner and I will do the first publish of everything for you.

I would always be happy to help as well with this, sorting issues and pr's (however if you prefer, you can always have the final say on the merging). I'm sure @VuiMuich and @mautamu would also be happy to help.

VuiMuich:
This is awesome.
I surly will help where I can, but would it even be possible to submit to crates.io without being the repo owner/maintainer?

On another topic, in my head #508 is growing to a bigger logger-rework. As currently it is part of `leftwm-worker` I feel this belongs to the binary and not to the core, right? Jut not 100% sure as some kind of logging would definitely be helpful in the core as well, I guess.

And finally @cecton do you know the sminez/penrose project? I just learned about it last week and tried it in a VM and it's quite nice. Is this where you wanted to go to with the leftwm-libification?

PR #515, opened 2021-10-09
readme: mention journalctl command to debug leftwm

hertg:
Just a stupid simple addition to the readme, mentioning how to view the leftwm logs when using journald.
Might not be as obvious to new contributors :)


VuiMuich:
I also had this in #508 but as I believe it will take a bit longer till I get this ready (in my head it is growing to a bigger logging rework) we surely can merge this now.

PR #521, opened 2021-10-10
Support sticky windows

### LINKED ISSUE #512: [feature] sticky windows
This feature request is inspired by the "sticky window" feature of dwm.

**Feature**
Allow to set a window as `sticky` dynamically, similar to setting a window as `floating`. A sticky window should always be visible, and "follow" you if you switch to another tag.

An example: When multitasking, eg, watching a video but working on something simultaneously, the user should be able to switch between tags to work with different windows, but the window playing the video should always follow the user and stay visible.

**Going into the implementation details**
Now I know that this feature can't be implemented to behave the same as it does in `dwm`. One key difference between `dwm` and `LeftWM` is that in `dwm` each workspace (using LeftWM terms) has its own set of tags, while in LeftWM all workspaces share the same set of tags. 

Obviously, just displaying the sticky window on all tags won't work, given you can have multiple workspaces visible on a multi-monitor setup, but we can't display the same window twice. So implementing this could become a little bit more complex. 

The following paragraphs are just some brainstorming on how this could be solved, they are not completely thought out solutions:
1. Instead of belonging to a specific tag, a sticky window could belong to a specific workspace, so it will be displayed on every tag on the given workspace. This would cause a behavior that might be weird though: On a two-monitor setup, swapping the workspaces between the monitors would leave the sticky window on the same monitor it was on, but all the other windows swapped. Also, implementing that might be a pain because it might contradict some current concepts. (?)
2. Sticky windows are handled the same in code, but on each tag-change-event we check if any currently visible tag goes out of sight. If it does, we iterate over the windows on it and search for all the windows marked as "sticky". Then we move the sticky windows to the tag which will be displayed next on the same workspace.

Do you guys have any other ideas or opinions on this feature request?

I'd be happy to send in a PR for that, but would like to receive some feedback first. :)

AethanFoot:
Close #512. Also updates window states in all situations allowing the simplification of previous fullscreen code.
Thanks.

### DIFF
--- a/src/display_servers/xlib_display_server/xwrap/setters.rs
+++ b/src/display_servers/xlib_display_server/xwrap/setters.rs
@@ -1,5 +1,5 @@
 //! `XWrap` setters.
-use super::{Window, WindowHandle};
+use super::WindowHandle;
 use crate::XWrap;
 use std::ffi::CString;
 use std::os::raw::{c_long, c_ulong};
@@ -83,8 +83,8 @@ impl XWrap {
     }
 
     /// Sets a windows fullscreen state.
-    pub fn set_fullscreen(&self, window: &Window, fullscreen: bool) {
-        if let WindowHandle::XlibHandle(h) = window.handle {
+    pub fn set_fullscreen(&self, handle: WindowHandle, fullscreen: bool) {
+        if let WindowHandle::XlibHandle(h) = handle {
             let atom = self.atoms.NetWMStateFullscreen;
             let mut states = self.get_window_states_atoms(h);
             if fullscreen {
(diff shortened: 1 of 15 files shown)


src/display_servers/xlib_display_server/xwrap/setters.rs
```diff
@@ -126,6 +105,26 @@ impl XWrap {
         }
     }
 
+    /// Sets a windows state.
+    pub fn set_state(&self, handle: WindowHandle, toggle_to: bool, atom: xlib::Atom) {
```
mautamu, line 109: Nice and elegant!

mautamu (approved):
Overall really nice! Solid PR.

Not entirely sure about the behavior, if I have the sticky window focused on tab 1 and then go to tab 2, should it stay focused? Not sure what other WMs do/what others expect, but I would expect the current behavior... I enjoy the functionality as is! 

Should probably document this in the wiki.

PR #524, opened 2021-10-11
Finish 165

AethanFoot:
Finishes #165. Do we want this on by default, if so should this window not take focus when spawned (as the mouse will be dragged back to the other screen). I am also unsure if we want a window swallowing feature and if this should be togglable. The other option for window swallowing is to add it to the list of patches we may want.

Thanks.


leftwm-core/src/handlers/window_handler.rs
```diff
@@ -17,22 +19,22 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
...
             &mut is_first,
+            &mut on_same_tag,
         );
-        insert_window(self, &mut window, is_scratchpad, layout);
+        insert_window(self, &mut window, layout);
 
         let follow_mouse = self.state.focus_manager.focus_new_windows
-            || self.state.focus_manager.behaviour == FocusBehaviour::Sloppy;
+            && self.state.focus_manager.behaviour == FocusBehaviour::Sloppy
+            && on_same_tag;
```
VuiMuich, line 37: I don't get it right now, why you changed this OR to AND.

leftwm-core/src/models/workspace.rs
```diff
@@ -16,7 +16,6 @@ pub struct Workspace {
     pub margin_multiplier: f32,
     pub gutters: Vec<Gutter>,
     #[serde(skip)]
-    all_tags: Vec<Tag>,
```
VuiMuich, line 19: Is this some passing by cleanup?

VuiMuich (commented):
Just my two cents after looking through this on mobile.
Review round two, later today.

leftwm-core/src/handlers/window_handler.rs
AethanFoot: Cause if focus_new_windows was false, it would still move the mouse to the window (in turn focusing it) if in focus_behaviour="Sloppy".

VuiMuich: So actually also a passing by bug fix, right?

PR #531, opened 2021-10-14
Reduce passing manager

AethanFoot:
This reduces the unneeded passes of manager, and moves any functions that just use state to State. This also means there is a lot less manager.state and we access things more directly.


VuiMuich:
Didn't have time to test this one actually, just looked at the code so far.

PR #538, opened 2021-10-22
Fixed wrong Monocle and MainAndDeck focus in Driven and ClickTo focus behaviour

Syudagye:
This PR fixes a problem i have with the `driven` focus behavior:
When i switch to the Monocle layout, if i don't focus the main window, the focused widow will be hidden and i will never be able to interact with the currently visible window, because it doesn't have focus.

**How to replicate this situation :**
- set `focus_behaviour = "Driven"` in `config.toml`
- open some windows on a tag
- focus a window other than the main one
- switch to Monocle

**wrong behaviour :** The main window get the full area but doesn't have the focus.

**fixed behaviour :** The main window get the focus.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,9 +34,9 @@ checksum = "d301b3b94cb4b2f23d7917810addbbaff90738e0ca2be692bd027e70d7e0330c"
 
 [[package]]
 name = "arc-swap"
-version = "1.3.2"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b5ab7d9e73059c86c36473f459b52adbd99c3554a4fec492caef460806006f00"
+checksum = "e6df5aef5c5830360ce5218cecb8f018af3438af5686ae945094affc86fdec63"
 
 [[package]]
 name = "atty"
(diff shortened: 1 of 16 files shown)


Syudagye:
No idea what happened with git complaining about merge conflicts, but it seems fine now

VuiMuich:
Commencing test 😆

VuiMuich:
I had one situation, where the windows of a tag with layout `MainAndDeck` got 'detached' aka unfocusable after switching to it from a tag where I followed the steps in OP. But I cent reproduce it a second time.

Syudagye:
> _sidenote: should we change (not necessarily in this PR) that when we switch to layout `Monnocle` the focused window becomes main? Currently the main window is put on top, no matter it had focus before or not._

I thought of this while trying to find a way to fix this, but couldn't find a way to do it. But yes that would feel more natural.

AethanFoot:
@VuiMuich would you say this is ready? Were you able to reproduce the issue.

VuiMuich:
I'll be back on my machine in about 3h. Will give it another try to reproduce. If I can't I'm gonna merge.

VuiMuich:
I tried hard but could not reproduce my earlier issue with `MainAndDeck` so lets assume this was an unrelated issue.

Thanks @Syudagye for your fixes!

PR #539, opened 2021-10-23
Improve configuration soft reloading

cecton:
This PR is an attempt to help doing "soft" reloading.

If I understand it properly. The current reloading system is to hard reload because soft reloading does not work properly for now. In This PR I will try to enumerate the issues and fix them.

### Checklist
- [x] Use owned type in `Config` trait: this change will allow the state to get owned type and not depend on a configuration to be loaded. My idea is that if the configuration changes but the state still use an old configuration, we should load only the state parameters for consistency.
- [x] Move `Config` from `State` to `Manager`: I wasn't sure why the config field was in the state but then I realized it is because it was part of the state. So to solve the previous point, I'm moving the `Config` to the `Manager`. `Manager` now has a `reload_config()` that takes no argument and spread the new configuration across the different states of the application
- [x] Reload config in XWrap should work now: I noticed a few lines of code were missing for it to work. I have not tested this yet but my hope is to fix it.
- [x] `screen_create_handler` is back to `Manager`: it made sense to move it to the State when config was there but now that I changed it it needs to go back to the Manager to be able to access the config. I did not touch the other functions that were moved to the state as I believe they were good :+1: 
- [x] Because `State` does not have the `Config` field, it now loses its generic... which is nice

### Future
- [ ] ~~I believe there is a startup script for theme that should be re-started when doing the reload_config somehow... maybe keeping track of widget processes... it does not sound trivial~~

### DIFF
--- a/leftwm-core/src/config/mod.rs
+++ b/leftwm-core/src/config/mod.rs
@@ -3,8 +3,8 @@ mod scratchpad;
 mod workspace_config;
 
 use crate::layouts::Layout;
-use crate::models::LayoutMode;
 pub use crate::models::{FocusBehaviour, Gutter, Margins, Size};
+use crate::models::{LayoutMode, Manager};
 use crate::state::State;
 pub use keybind::Keybind;
 pub use scratchpad::ScratchPad;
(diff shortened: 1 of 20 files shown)


cecton:
> Oh boy, I'm afraid my appreciation worded to kiddingly came off quite wrong. I am very sorry for that. I really love that you did this PR and just thought I might joke a bit about the oddity that exactly you who did this as this appears to be quite the opposite what you were working for in your other PRs.

haha no worry! I felt the joke appropriate!

I finished updating the branch btw

AethanFoot:
I am hoping to merge #538 first (sorry for the extra merge conflicts).

cecton:
> I am hoping to merge #538 first (sorry for the extra merge conflicts).

No problem! I personally use [`git try-merge`](https://github.com/cecton/git-tools/#git-try-merge) to resolve conflicts one-by-one with the main branch. It makes conflict resolution much easier, I highly recommend it!

cecton:
Branch is updated again ^_^

PR #555, opened 2021-10-31
Rename branch master to main

cecton:
I thought we should rename our main branch to `main` instead of `master`.

Every new projects on GitHub now uses "main" by default and to be honest it feels weird now when I read "master".

More on the subject: https://www.theserverside.com/feature/Why-GitHub-renamed-its-master-branch-to-main

After merging this you need to go to the settings of the project and rename the branch there too and that's all.

Locally though, everybody who don't want to re-clone the project needs to update the default branch using this command:

```
git remote set-head origin --auto
```

It's `origin` if the remote is origin but it can be `upstream` depending on your local set up:

```
git remote set-head upstream --auto
```

**Note:** re-cloning the repo will default to main and no action is needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
 
 env:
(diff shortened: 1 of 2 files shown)


VuiMuich:
I am in favour of this, but it requires @lex148 attention to adjust the project settings, so it makes sens, that he also merges this.

PR #557, opened 2021-10-31
make tag IDs numeric and unique / use tag names as labels only

### LINKED ISSUE #68: Allow the same label in multiple tags
When two tags have the same character in the "tags" section of leftwm/config.toml they are treated as one even if the number corresponding is different.
In my case two tags were named "I": number 1 and 4, this made the wm show the same windows in tag 1 and in tag 4.
Moving windows from and to tags 1:4 works as usual, one window in [eg] tag 2 is then shown in both 1 and 4, a window shown in 1 and 4 is one when moved to any other tag.

hertg:
## :heavy_check_mark: Features
- Refactored tag IDs to be of type `usize` rather than `&str`.
  Tag IDs are unique now and automatically assigned by LeftWM. The new `Tags` struct takes care of assigning numbers, keeping them in order and distringuish between "normal" and "hidden" windows
- Added some docs, mostly in the `tag.rs` file

## :adhesive_bandage: Fixes
- Having multiple tags with the same label is now possible (closes #68)
- Shortened some code here and there

**TODO**
- [ ] ~When moving the focused window to another tag, it should lose focus, and instead another window on the current tag gets focused.~ -> This is not a regression of this PR, [this line](https://github.com/leftwm/leftwm/blob/53e5d11250fbaecf8a125c206406130c8a5b6f62/leftwm-core/src/handlers/window_handler.rs#L143) is the cause though...


AethanFoot:
This is looking really good so far. Great work!

hertg:
I just realized that my TODO "When moving the focused window to another tag, it should lose focus, and instead another window on the current tag gets focused." isn't actually a regression of my changes, it's a pre-existing issue when the focus-behavior is `Sloppy` and the mouse doesn't hover a window after the WindowMove. Then the focus stays on the moved window which is no longer visible. I don't think this is expected behavior, is it? Focus should probably always be on a window that's currently visible.

hertg:
@AethanFoot @VuiMuich @mautamu 

I think I'm setting this as ready for review.
I'd still like to address the inconsistency that Workspace IDs start at 0, but Tag IDs start at 1,
and that in some places the tag is referenced by it's index, rather than its ID. (In `Command::SendWorkspaceToTag` for example).

But as those are breaking changes, it'd be better to do that in another PR, this one is already big enough. It should be non-breaking though. The only thing that may happen is all your windows ending up on the first tag after you update + SoftReload. This is because windows are tagged by ID rather than label now.

I ran my system on this branch for the last 3 days and haven't experienced any regressions related to this PR.

leftwm-core/src/display_servers/xlib_display_server/xwrap/setters.rs
```diff
@@ -21,13 +22,11 @@ impl XWrap {
...
-    pub fn set_current_desktop(&self, current_tags: &str) {
-        let mut indexes: Vec<u32> = vec![];
-        for (i, tag) in self.tags.iter().enumerate() {
-            if current_tags.contains(tag) {
-                indexes.push(i as u32);
-            }
-        }
+    pub fn set_current_desktop(&self, current_tags: &[TagId]) {
+        let mut indexes: Vec<u32> = current_tags
+            .iter()
```
mautamu, line 27: Feels a lot more 'rusty' :)

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -679,44 +680,59 @@ mod tests {
...
+        let current_tag = manager.state.focus_manager.tag(0).unwrap_or_default();
+        assert_eq!(current_tag, 4);
+
+        // test tag history
+        assert_eq!(manager.state.focus_manager.tag(1).unwrap_or_default(), 3);
+        assert_eq!(manager.state.focus_manager.tag(2).unwrap_or_default(), 2);
+        assert_eq!(manager.state.focus_manager.tag(3).unwrap_or_default(), 6);
+    }
+
+    #[test]
```
mautamu, line 704: Thanks for adding these, looks good!

mautamu (approved):
Don't see anything alarming; looks good to me! Thanks for this! :)

leftwm-core/src/models/tag.rs
```diff
@@ -57,30 +245,23 @@ impl Tag {
...
-            return;
-        }
-        *mwp += delta as u8;
-        if *mwp > 100 {
-            *mwp = 100;
-        }
+        self.main_width_percentage = (self.main_width_percentage as i8 + delta)
+            .max(0) // not smaller than 0
+            .min(100) as u8; // not larger than 100
     }
```
AethanFoot, line 257: This is so nice

AethanFoot (approved):
This looks great, there seems to have been a good amount of testing so this should be ready. Unless we want another day of more people testings.

hertg:
I fixed some small stuff in the docs/comments and added tests in `tag.rs`. Haven't changed anything in the actual source code though. Can probably be merged if nothing else comes up :)

hertg:
There's one question that came up after writing the tests: Should `Tag.label` really be a `String` or would it be better if it was `&str`? Does using `String` for the label provide any benefits that I'm not seeing?

The `Tag.id` was `String` previously, that's why I also used `String`.

VuiMuich:
I believe, if we eventually would want to allow fully dynamic labels `String` would be easier to mutate, right?

If you guys agree, I think we can merge.

AethanFoot:
Ye I think String makes more sense as it is an owned value (also clippy might moan about &str).

AethanFoot:
I have noticed the tag doesn't seem to restore its settings, nothing major.

hertg:
> I have noticed the tag doesn't seem to restore its settings, nothing major.

Can you elaborate? I'm not sure I understand what you mean by that :P

AethanFoot:
After a soft reload the prevois layout/main width doesn't seem to restore.

hertg:
I think I found the problem, I'll submit a fix soon

PR #563, opened 2021-11-05
fix #549

### LINKED ISSUE #549: keystrokes are sometimes grabbed by an unfocused window
**Describe the bug**
Sometimes when a new window opens or when the focus changes, the keystrokes aren't grabbed by the newly focused window but rather the previously focused one. Occurs when using focus behavior `Sloppy`.

**To Reproduce**
I don't have a reliable way to reproduce that issue yet.

**Expected behavior**
The focused window should always be the one grabbing the keystrokes.

**Additional context**
I'm on the latest `master`. While I don't know how to reproduce it right now, the issue was also confirmed by @mautamu in the `#dev` channel on Discord. I didn't have that issue for long, so it is probably a regression of a recent commit (last 1-2 weeks?).

It happens around 1-2 times a day on my machine, and is mostly just an annoyance. I don't look forward to the day where I'm focused on a password prompt and send my password in a Discord channel though :wink:

AethanFoot:
Fixes #549, at least in my testing. Also doesn't seem to break anything (from what I can tell).
Thanks.

### DIFF
--- a/leftwm-core/src/display_servers/xlib_display_server/xwrap/window.rs
+++ b/leftwm-core/src/display_servers/xlib_display_server/xwrap/window.rs
@@ -166,10 +166,7 @@ impl XWrap {
         if let WindowHandle::XlibHandle(handle) = window.handle {
             self.grab_mouse_clicks(handle);
 
-            if window.never_focus {
-                // Tell the window to take focus
-                self.send_xevent_atom(handle, self.atoms.WMTakeFocus);
-            } else {
+            if !window.never_focus {
                 // Mark this window as the `_NET_ACTIVE_WINDOW`
                 unsafe {
                     (self.xlib.XSetInputFocus)(
@@ -187,6 +184,10 @@ impl XWrap {
                     );
                     std::mem::forget(list);
                 }
+            }
+            if self.focus_behaviour != FocusBehaviour::Sloppy {
+                // Tell the window to take focus
+                self.send_xevent_atom(handle, self.atoms.WMTakeFocus);
             }
         }
     }
--- a/leftwm-core/src/display_servers/xlib_display_server/xwrap/window.rs
+++ b/leftwm-core/src/display_servers/xlib_display_server/xwrap/window.rs
@@ -185,7 +185,7 @@ impl XWrap {
                     std::mem::forget(list);
                 }
             }
-            // This is fixes windows that process the `WMTakeFocus` event too slow.
+            // This fixes windows that process the `WMTakeFocus` event too slow.
             // See: https://github.com/leftwm/leftwm/pull/563
             if self.focus_behaviour != FocusBehaviour::Sloppy {
                 // Tell the window to take focus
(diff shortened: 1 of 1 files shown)


AethanFoot:
My guess is that when warping the pointer, we receive an enter event from firefox/libre office/... then receive an enter event from the other window. These events are processed normally (e.g. the active window gets set properly), the issue may come from firefox/libre office/... processing the wm_take_focus event slowly. This may mean that the "fast" app has already processed this event, giving itself focus, but the "slow" window then processes this event after, taking the focus back.

hertg:
Would be great if you could add a comment reasoning the if-statement, maybe even mention the Github issue number. Otherwise we might end up in a similar situation like recently where nobody knew what the reason for a certain if-statement was :)

hertg:
> Would be great if you could add a comment reasoning the if-statement, maybe even mention the Github issue number. Otherwise we might end up in a similar situation like recently where nobody knew what the reason for a certain if-statement was :)

I think we should start doing that more consistently to prevent regressions in future updates. It's easy to remove code that mitigates some weird errors during refactoring because it is not inherently clear why it's there. @VuiMuich @mautamu @AethanFoot

hertg:
The fix works for me. I'm not sure if it could potentially become problematic though. 
The Method is named `window_take_focus(window)`. Isn't this fix essentially saying, "i don't care whoever calls this method, for whatever reason, don't make the window take focus if the focus-behavior is set to sloppy". This seems kinda wrong? It's not like you'd never allow a window to take focus in the `Sloppy` behavior. It's just that it can be problematic in combination with the automatic mouse move of the `Sloppy` behavior.

AethanFoot:
The XSetInputFocus takes the focus as well, this is more of a just in case. It probably needed in say driven as the window won't now it is taking focus, as the mouse isn't moved to it.

hertg:
Ok, thanks for the insight. Are there any applications that may depend on receiving the `WMTakeFocus` for the focus change to actually work? Maybe because they ignore other focus-events and only implemented this one? If so, this could potentionally break those from working in the `Sloppy` behavior, but it may also be that an application implementing just the `WMTakeFocus` would have implemented some spec wrong? X11 and its API is still a black box to me, lol.

hertg (approved):

VuiMuich:
`i_dont_care_whoever_calls_this_method_for_whatever_reason_dont_make_the_window_take_focus_if_the_focus-behavior_is_set_to_sloppy(window)` :+1: function name! ... jk

AethanFoot:
The hard thing about x11 is you have overlapping specs with various levels of implementations. We may just need to play it by ear and implement exceptions for certain window classes. But I doubt this will be an issue as most users of these applications will use a mouse, and every different window manager being required to tell the app to take focus isn't going to be reliable.

VuiMuich:
Can confirm, this fixes #549.

Tested with blender and libreoffice writer vs alacritty

hertg:
Seems like there is no veto, I think this can be merged?

PR #565, opened 2021-11-06
relative_find helper: fix bug, add tests, add docs, remove duplicate code

hertg:
I saw that the `relative_find` logic was implemented redundantly in `focus_tag_change`. The method now also uses the pre-existing helper method.

After looking at the code, I realized that it had some bugs, for example: on a list of length `1` the only element in the list was always returned, even if `should_loop` was set to false. I've re-implemented the method and added tests to verify the implementation. Please let me know if you think a test-case is missing.


AethanFoot:
Great work as always!

Issue #558, opened 2021-10-31
Help wanted to improve documentation.


AethanFoot:
The is a call-to-arms of anyone who wants to help out. The task at hand is improving the code documentation and the information in the wiki.

For the code documentation follow the [guides provided by rust](https://doc.rust-lang.org/rust-by-example/hello/comment.html). The xwrap folder (leftwm-core/src/display_servers/xlib_display_server/xwrap) have been documented, with helper comments for new contributors. The helper comments should only need to detail anything not clear from the code, or if you have a link to something explaining stuff in the code is great.

The wiki is open to be added to. We appreciate it if you add something not already mentioned, want to improve the already documented stuff, or just want to add some tips & tricks/trouble shooting points.

When I have some more time I will add a list of code areas that need documenting, unless someone else wants to attach one first.

Many Thanks!

NicTanghe:
what whould be required and if i don`t understand a line of could can i ask for help for what it means ?
I`ve been wanting to contribute for a while and   have been reluctant to try and figure out where to start and how everything is connected.

PR #566, opened 2021-11-06
fix #552

### LINKED ISSUE #552: Menu/History... in Brave disappear when is cover second window
**Describe the bug**
Menu/History... in Brave disappear when is a cover second window

**To Reproduce**
Open two brave's windows on the same screen
On the left window click "Customize and control Brave"
After the menu appear click "History|
Move the mouse to some item on the list which is covered right window

**Expected behavior**
Should be still selected left window and possibility item from the menu

**Screenshots**
https://drive.google.com/file/d/1PDYBqJR-6Nsftr2hN94JAaVpdA-T5yx2/view?usp=sharing

**Environment:**
 - OS/Distro: Arch
 - Output of `leftwm-check`
`:: LeftWM version: 0.2.9
:: LeftWM git hash: a6326ae
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK
`
**Additional context**

AethanFoot:
Fixes #552. Instead of using the point we just use the window the cursor is reporting. Also changed it so the notify mode doesn't have to be normal (so it can be grab and ungrab), this eliminates half the reason window focus events get missed.

To add the reason I have found where focus events are missed are when a mouse button is pushed we don't receive any motion or enter notifies. So when you move the mouse to a window when holding a button the focus didnt change (this is without the validate focus), as the notify mode is classed as a grab. So allowing these events means the focus would update once the button is released (However this is probably unneeded with the updated validate focus)

Thanks!
ps I gonna add some comments tomorrow, have cant let hertg show me up :stuck_out_tongue_winking_eye:

### DIFF
--- a/leftwm-core/src/display_event.rs
+++ b/leftwm-core/src/display_event.rs
@@ -9,14 +9,14 @@ use crate::Command;
 pub enum DisplayEvent {
     Movement(WindowHandle, i32, i32),
     KeyCombo(ModMask, XKeysym),
-    KeyGrabReload, // Reloads keys for when keyboard changes
+    KeyGrabReload, // Reloads keys for when keyboard changes.
     MouseCombo(ModMask, Button, WindowHandle),
     WindowCreate(Window, i32, i32),
     WindowChange(WindowChange),
     WindowDestroy(WindowHandle),
     MouseEnteredWindow(WindowHandle),
-    VerifyFocusedAt(WindowHandle), //Request focus validation at this point
-    MoveFocusTo(i32, i32),         //Focus the nearest window to this point
+    VerifyFocusedAt(WindowHandle), // Request focus validation for this window.
+    MoveFocusTo(i32, i32),         // Focus the nearest window to this point.
     MoveWindow(WindowHandle, c_ulong, i32, i32),
     ResizeWindow(WindowHandle, c_ulong, i32, i32),
     ScreenCreate(Screen),
--- a/leftwm-core/src/display_event.rs
+++ b/leftwm-core/src/display_event.rs
@@ -15,8 +15,8 @@ pub enum DisplayEvent {
     WindowChange(WindowChange),
     WindowDestroy(WindowHandle),
     MouseEnteredWindow(WindowHandle),
-    VerifyFocusedAt(i32, i32), //Request focus validation at this point
-    MoveFocusTo(i32, i32),     //Focus the nearest window to this point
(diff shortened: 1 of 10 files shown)


hertg (approved):
I love all those patches coming in from you! :smile: 

I just took a quick glance at the code and it all looks great. I'm still lacking experience with the whole X11 API for my review to mean much, but your description is great and explains everything clearly. For once, I actually understood what's going on :laughing:

hertg:
This isn't related to the PR directly: 

What does the validate focus logic do exactly, you mean the one called periodically in the event-loop, right? If this PR fixes the problem that we didn't receive some events while a button is pressed, are there still other events that get missed even with this patch?

The reason I ask is because there are some "bandaid" lines I occasionally stumble across when browsing the code (Or at least I think they are, there's always the possibility that I'm just too dumb to understand why they are there :laughing:). As inconsistencies and bugs get patched out more and more with PRs like this, we should keep an eye on the bandaid solutions and continuously remove them if they become obsolete.

It's a pity that we can't really write tests for stuff like the issue you described here.

AethanFoot:
I will probably update the comments with something like this.  The events arent exactly missed but blocked, which would cause delay in the focusing (until the mouse button is released). The bit in the event loop will basically update the focus if the EnterNotify has been delayed for too long (Im gonna try and see if I can devise a better way of calling this smater rather than a straight delay).

VuiMuich:
> This isn't related to the PR directly: 
> 
> 
> 
> What does the validate focus logic do exactly, you mean the one called periodically in the event-loop, right? If this PR fixes the problem that we didn't receive some events while a button is pressed, are there still other events that get missed even with this patch?
> 
> 
> 
> The reason I ask is because there are some "bandaid" lines I occasionally stumble across when browsing the code (Or at least I think they are, there's always the possibility that I'm just too dumb to understand why they are there :laughing:). As inconsistencies and bugs get patched out more and more with PRs like this, we should keep an eye on the bandaid solutions and continuously remove them if they become obsolete.
> 
> 
> 
> It's a pity that we can't really write tests for stuff like the issue you described here.
> 
> 

Probably an easy way would be to add `//🤕` or `// bandaid` along these fixes and mark all suspicious lines as well, when we will attend #558

For a moment I thought about a feature-gate in cargo.toml but that would not work as intended I guess.

I find it hard to find proper tests in general! The tests I came up, when trying either are so trivial that I doubt they are of any use, or they don't work at all 😂

PR #567, opened 2021-11-06
fix regression from #557, restore tag state on reload

hertg:
Fixes a regression reported by aethan in #557.

The `all()` methods returned a clone of the tag-list, mutating it did therefore not affect the actual state.
I rewrote so `all()` returns actual borrowed values, rather than cloned ones and introduced `all_mut()` in case the tags should be mutable.

I'm not a 100% sure if that's the best way to do it, please let me know if you have any opinions on that.


mautamu:
Looks reasonable to me

PR #533, opened 2021-10-17
added makefile for easier install, uninstall and development

hertg:
This addresses #139 and #471. This allows for easier installation / uninstallation and also eases the development setup, the following make rules are provided.

`make`\
Runs the `build` and `test` rules described below

`make test`\
Runs tests and linters (test, fmt, clippy)

`make build`\
Build the project with `cargo build --release --features=journald`

`make clean`\
Cleans the `**/target` directories

`make install`
Builds the project and installs the binaries + `.desktop` file in `/usr/bin` and `/usr/share/xsessions` respectively.

`make install-dev`
Builds the projects and **links** the binaries to `/usr/bin` and installs the `.desktop` file to `/usr/share/xsessions`.

`make uninstall`
Uninstalls leftwm from the system by removing binaries from `/usr/bin` and `.desktop` file from `/usr/share/xsessions`.

**Note**
Target directories can be overwritten via `TARGET_DIR` and `SHARE_DIR`. (eg. `make install TARGET_DIR=/usr/local/bin`)

---

@AethanFoot @VuiMuich 
Let me know if you'd like to change any of the rules or if you want to have additional ones. I can also update the commands mentioned in the README, wanted to wait on your feedback before doing that though :smile:

PR #570, opened 2021-11-08
README: Add info about Makefile

VuiMuich:
As mentioned in #533 (though adding stuff, instead of making the REAME.md lighter :joy:)﻿

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,7 @@ Mod + Shift + R
 
 ## Using the Makefile
 
-For conveniece we also have a Makefile with the following subcommands:
+For conveniece we also have a Makefile with the following rules:
 
 |make ... | info |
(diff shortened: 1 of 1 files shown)


AethanFoot:
Maybe in future (not necessarily this pr), to lighten the readme we could only provide the recommended ways of installing with a package manager and without one and place the rest in the wiki with a link pointing to them as alternatives.

VuiMuich:
> Maybe in future (not necessarily this pr), to lighten the readme we could only provide the recommended ways of installing with a package manager and without one and place the rest in the wiki with a link pointing to them as alternatives.

That was somewhat the plan when discussing this idea elsewhere (I believe discord). Maybe this something light and easy for the weekend...

PR #572, opened 2021-11-10
Fix wrong _NET_CURRENT_DESKTOP index.

VentGrey:
As discussed on Discord, commit 589dc71d2a98500c83c2a94210d4d2e80773bc7b introduced a minor regression where _NET_CURRENT_DESKTOP started at `1` instead of `0`. This PR should fix this as suggested by @hertg :)


hertg:
I think we can merge this right away :)

PR #573, opened 2021-11-10
Add manpage

VentGrey:
This is part of #558 help wanted. Some users might want to refer to offline documentation on how to configure / use leftwm properly. This commit introduces a manpage placed in `/usr/local/share/man/man1/leftwm.1`.

This manpage was written using the current documentation available on leftwm with some changes:
- The version requirements have been removed (it is assumed that people build leftwm via git for now)
- Replaced unicode / emojis with text


VuiMuich:
Looking great!
One small request though: would you also include the .md so updating the manpage can be done in there and recompiled with pandoc.

VentGrey:
Sure! I'll make sure to make the appropiate changes ASAP :)

VentGrey:
:thinking: Should I add a new section to the `README.md`  file warning users about pandoc to generate the manpage?

hertg:
Thanks so much for the PR :)

I'm uncertain how we should manage that. I think you based the manpage on the current wiki, made a few tweaks and converted it, right? I do like verbose manpages like that, but I'm a bit worried that if this is done manually, the contents of the manpage and the wiki might divert quickly, which is problematic for the users, because it is no longer clear which of them is right and which isn't.

Also, should we create multiple manpage sections to separate commands from config file formats etc? see https://man7.org/linux/man-pages/man7/man-pages.7.html#DESCRIPTION

One small sidenote: I'd personally prefer the folder to be named `docs` rather than `doc`, because it's what I see most in other repos. Obviously a minor thing, I can live with either.

VentGrey:
If the wiki docs can be extracted as markdown files or "scrapped" in some way there should be an easy path to automate that part. As of manpage sections I saw that most wm's don't really separate them. (I made this manpage based on the one that herbs(something) wm [provides](https://github.com/herbstluftwm/herbstluftwm/blob/master/doc/herbstluftwm.txt).

The folder rename is the most trivial thing so I think it would harm nobody if it is `docs/` :) let me know what you think.

VuiMuich:
> :thinking: Should I add a new section to the `README.md`  file warning users about pandoc to generate the manpage?

The README should mention the existence of the manpage.
Compiling the manpage with pandoc should be in the `contributing` wiki, I guess.

VentGrey:
> > thinking Should I add a new section to the `README.md`  file warning users about pandoc to generate the manpage?
> 
> The README should mention the existence of the manpage. Compiling the manpage with pandoc should be in the `contributing` wiki, I guess.

Right. I'll submit more commits later separating the compile step from the development build.

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,343 @@
...
+T}
+T{
+Mod + Shift + (1-9)
+T}@T{
+Switch a floating window to tiling mode
+T}
+.TE
+.SS Configuration
+.PP
+All entries require a modifier, even if blank: \f[C]modifier = []\f[R]
```
VuiMuich, line 177: It might be misleading to have this info in this place. This should be mentioned inthronisiere keybinds section. 

Or is it the `Configuration` heading two lines above that is wrong?

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,343 @@
...
+.PP
+\f[B]-v, \[en]version\f[R]: Prints version information.
+.PP
+\f[B]check\f[R]: Checks the syntax of the configuration file
+.PP
+\f[B]command\f[R]: Sends external commands to \f[C]leftwm\f[R]
+.PP
+\f[B]state\f[R]: Prints the current state of leftwm (in JSON format)
+.PP
+\f[B]theme\f[R]: Manage leftwm themes
```
VuiMuich, line 35: `leftwm-theme` is not part of the standard package, this should be mentioned

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,343 @@
...
+.SS Configuring leftwm
+.PP
+The settings file to change key bindings and the default mod key can be
```
VuiMuich, line 38: This line should also mention other config options like tags, workspaces etc.

VuiMuich (commented):
A view annotations from reading on mobile.

VentGrey:
I've made corresponding changes :) i'll be waiting for feedback to improve the current manpage text

hertg:
Would it be an option to create the manpage manually instead of converting it from markdown? I know it's a bit of a pain to get into the troff syntax, but I'd prefer handmade. Because the conversion from markdown has some downsides:
- Tracking the source **and** the generated file on Github seems weird, because they can divert if the generation doesn't happen automatically.
- The resulting manpage syntax is a bit weird in some places (like the keybinds table) and it's overall kind of unreadable because it seems to convert things into control characters where other ones would be more appropriate.

I've written a manpage for a project of mine by hand in the past, I know it's an unfair comparison because mine contains a lot less information than the one for LeftWM, but the difference in readability is definitely noticeable.
- [The automatically generated one](https://github.com/leftwm/leftwm/blob/e7732ab9ab776bb05995ed9012d25a503f5b9e62/leftwm/doc/leftwm.1)
- [The handmade one](https://github.com/hertg/egpu-switcher/blob/master/docs/egpu-switcher.1)

That's just my opinion though, I'm open to change my mind about it :)

VuiMuich:
So, I finally managed to read it actually with `man`.
I basically have these points:
- headlines didn't get converted properly, so they don't stand out against plain text
- subcommands could use a little more info imo, especially lefftm-command and leftwm-check as they are most in contact with the user directly, leftwm-state is also important, but only when tinkering with themes, so having this in the comprehensive guide in the wiki is fine
- as much as I like it in general, the 'demonstration' of the `max_window_width` feels a bit out of place here, I'd rather have this compact
- config info for workspaces is missing at all, but this is somewhat important to have in offline info IMO
- info on how the default terminal is determined should be here as well I guess
- in the floating section I am a bit uncertain about `Mod + Shift + (1-9)       Switch a floating window to tiling mode` as this feels a bit like unintentional behaviour actually. But we are missing the `FloatingToTile`, `TileToFloating` and `ToggleFloating` commands
- In Layouts please:
  ```diff
  - Leftwm supports an ever-growing amount layouts, which define the way that windows are tiled in the workspace.
  + Leftwm supports variety of layouts, which define the way that windows are tiled in the workspace.
  ```
I guess that's it for now.

I guess hertg is right, that having the markdown source here is not the best idea. And as observed with headings, it seems indeed, that the conversion works not as flawless as expected.
If you would be willing to manually tidy the troff syntax up, I guess we could loose the `.md` again..

VentGrey:
I'll begin working on a proposal. It might take a while since I have to learn the troff syntax. I'll submit another commit with explanations when done :)

@hertg woud you consider using `groff` instead? :thinking:

hertg:
> @hertg woud you consider using `groff` instead? :thinking: 

I wasn't even sure if it's troff or groff that is used by `man`? I'm fine with either.

VentGrey:
I think both are valid but groff comes with handy extensions. I'm still working on it as the syntax is kind of tricky. Feel free to add wishlist or suggestions to this manpage, I would find that particularly useful since y'all the devs and know the codebase from up close.

VuiMuich (changes requested):

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,332 @@
...
+T{
+Drag window onto a tile
+T}@T{
+Switch a floating window to tiling mode
+T}
+T{
+Mod + Shift + (1-9)
+T}@T{
+Switch a floating window to tiling mode
+T}
```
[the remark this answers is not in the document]
VentGrey: The README.md file (https://github.com/leftwm/leftwm#floating-windows) has this at the end of the floating windows table. I know it might be an accident that got into the file but I'll remove it from here

VentGrey:
Made changes requested :D

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,332 @@
...
+starts the left window manager on $DISPLAY. This also launches leftwm-worker and logs for errors to the console. The list of available subcommands available for leftwm is listed below:
+.SH SUBCOMMANDS
+.IP "-h, --help"
+Prints help information.
+.IP "-v --version"
+Prints the version information.
+.IP "check"
+This command will run several actions to ensure leftwm is configured properly, this will report the current leftwm version and git commit, it will also check if configuration is loaded correctly, check for syntax errors in your
+.I config.toml
+file, check your environment and theme files for errors as well.
```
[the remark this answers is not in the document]
mautamu: Perhaps
"Themes are checked by this command for very basic and common mistakes, e.g. improper permissions or missing files, but are not checked for accuracy."

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,339 @@
...
+.IP "command"
+This subcommand sends commands directly to leftwm. It can also be used to concat commands for a keybind in
+.I config.toml
+For a list of a list of available commands use the '-l|list' flag.
```
mautamu, line 27: "For a list of available commands"

mautamu (commented):
Good job!

leftwm/doc/leftwm.1
VentGrey: Is there a typo somewhere :scream: ?

VuiMuich: Double appearance of "a list"

VentGrey: Solved xD

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,339 @@
...
+Mod + Shift + R
+T}@T{
+Reload LeftWM and its config
+T}
+T{
+Mod + p
+T}@T{
+Use dmenu to start application
+T}
+.TE
```
hertg, line 130: This table markup scares me. Could you change that to something like the following?
Maybe use a simpler markup like that on the other tables too?

```groff
.TS
tab(;);
l l.
Keybinding;Description
_
Mod + (1-9);Switch to a desktop/tag
Mod + Shift + (1-9);Move the focused window to desktop/tag
Mod + W;Switch the desktops for each screen. Desktops [1][2] changes to [2][1]
Mod + Shift + W;Move window to the other desktop
Mod + (up or down);Focus on the different windows in the current workspace
Mod + Shift + (up or down);Move the different windows in the current workspace
Mod + Enter;Move selected window to the top of the stack in the current workspace
Mod + Ctrl + (up or down);Switch between different layouts
Mod + Shift + (left or right);Switch between different workspaces
Mod + Shift + Enter;Open a terminal
Mod + Ctrl + L;Lock the screen
Mod + Shift + X;Exit LeftWM
Mod + Shift + Q;Close the current window
Mod + Shift + R;Reload LeftWM and its config
Mod + p;Use dmenu to start application
.TE
```

hertg (changes requested):

leftwm/doc/leftwm.1
VentGrey: I'm no groff / troff expert, will that render the same on terminal? :o

hertg: haha, me neither :)
Yeah it should, would probably be good if you could do a quick comparison to make sure.

hertg: Here's a quick roundup what the markup means as I understand it, but correct me if I'm wrong because I learned that by doing, not by reading the spec :stuck_out_tongue: 

I interpret `.TS` as Table-Start and `.TE` as Table-End.
At the top is `tab(;);` which means: The columns will be separated by a semicolon in the markup. Notice how in the original it says `tab(@);`, that's why the columns are separated by `@` originally.
Then the `l l.` means both column have left-aligned text.
The `_` just adds the separating line below the header row
The rest is pretty self-explanatory.

I'm not sure what the 
```groff
T{
T}
```

in the original is for. I'm guessing that's useful if you want to use advanced markup in the cell contents themselves? But since you aren't doing that, it seems unnecessary.

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,339 @@
...
+This behaviour snaps the currenly focused window into the workspace below.
+.SS TileToFloating
+This behaviour switches the currently focused window to floating mode when it's tiled.
+.SS ToggleFloating
+This behaviour switches the currenly focused window between floating and tiled mode.
+
+.\" Configuration section\"
+.SH CONFIGURATION
+.B IMPORTANT:
+You will need to You will need to reload in order to apply any changes you write to
```
hertg, line 214: ~You will need to~ You will need to

hertg (changes requested):

leftwm/doc/leftwm.1
VentGrey: I just realized I was recording a kbd macro in vim when writing (pressed q twice xD) that's why some text is duplicated.

VentGrey:
Fixed the table markup syntax and removed the duplicate.

VuiMuich:
> Fixed the table markup syntax and removed the duplicate.

Nice this is way better for markdown readability! Especially as in line 121-123 where you have space around the semicolons.

VentGrey:
Glad that you like it :D, BTW thanks for giving me the new ability of writing groff. This will prove useful in the future.

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,260 @@
...
+You can optionally switch between tiling or floating mode for any
+window.
+.PP
+.TS
+tab(;);
+l l.
+
+Keybinding;Description
+_
+
```
hertg, line 120: Remove the empty line
```suggestion
```

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,260 @@
...
+Mod + MouseDrag ; Switch a tiled window to floating mode
+Mod + RightMouseDrag ; Resize a window
+Drag window onto a tile ; Switch a floating window to tiling mode
```
hertg, line 123: Shouldn't this say something like "Drag window to workspace edge" with its current behavior?

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,260 @@
...
+Focus follows the mouse, hovering over a window brings it to focus.
+.IP "2." 3
+Click-to-Focus.
+Focus follows the mouse, but only clicks change focus.
+.IP "3." 3
+Driven Focus.
+Focus disregards the mouse, only keyboard actions drive the focus.
+.IP "4." 3
+Event Focus.
+Focuses when requested by the window/new windows.
```
hertg, line 168: I don't think "Event" is a thing (anymore)?
https://github.com/leftwm/leftwm/blob/589dc71d2a98500c83c2a94210d4d2e80773bc7b/leftwm-core/src/models/focus_manager.rs#L10-L14

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,260 @@
...
+workspaces = [
+    { y = 0, x = 0, height = 1440, width = 1720 },
+    { y = 0, x = 1720, height = 1440, width = 1720 },
+]
+\f[R]
+.fi
+
+.SS Tags
+.PP
+Tags are the names of the virtual desktops were windows live.
```
hertg, line 246: typo
```suggestion
Tags are the names of the virtual desktops where windows live.
```

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,260 @@
...
+In other window managers these are sometimes just called desktops.
+You can rename them to any unicode string including symbols/icons from
+popular icon libraries such as font-awesome.
+.PP
+Default:
+\f[C]tags = [\[dq]1\[dq], \[dq]2\[dq], \[dq]3\[dq], \[dq]4\[dq], \[dq]5\[dq], \[dq]6\[dq], \[dq]7\[dq], \[dq]8\[dq], \[dq]9\[dq]]\f[R]
+
+
+.SH BUGS
+If you find any bugs or functionality issues please report them to the following URL: https://github.com/leftwm/leftwm/issues
```
hertg, line 256:
```suggestion
If you find any bugs or functionality issues please report them on Github: https://github.com/leftwm/leftwm/issues
```

hertg (changes requested):

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,260 @@
+.\" Process this file with
+.\" File taken from BSD mandoc template.
+.\"
+.TH leftwm 1 "November 2021" Gnu/Linux "User Manuals"
```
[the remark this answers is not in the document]
VentGrey: I thnk that is the date when it was written, not updated :sweat_smile:  I can remove it if needed :)

VentGrey:
Added corrections as requested :)

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,256 @@
...
+This value can be used to determine which key, when held, can assist a
+mouse drag in resizing or moving a floating window or making a window
+float or tile.
+.PP
+Default: \f[C]mousekey = \[dq]Mod4\[dq]\f[R] (windows key)
+.SS Focus Behaviour
+.PP
+LeftWM has 3 focusing behaviours (Sloppy, ClickTo, and Driven) and
+one option (focus_new_windows), which alter the way focus is handled.
+These encompass 4 different patterns:
```
hertg, line 155:
```suggestion
These encompass 3 different patterns:
```

hertg (approved):
I hinted just one more typo, the rest looks great.
Thanks for sticking to it despite all the nitpicks :)

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,256 @@
...
+.IP "check"
+This command will run several actions to ensure leftwm is configured properly, this will report the current leftwm version and git commit, it will also check if configuration is loaded correctly, check for syntax errors in your
+.I config.toml
+file, check your environment and theme files for errors as well.
+.B "NOTE:"
+This subcommand will check for basic / common mistakes made in the directory structure or permissions, however it will not check or report if any mistakes were made in any extra scripts, bar configurations, liquid templates and anything that isn't the up or down executables.
+.IP "command"
+This subcommand sends commands directly to leftwm. It can also be used to concat commands for a keybind in
+.I config.toml
+For a list of available commands use the '-l|list' flag.
```
VuiMuich, line 29:
```suggestion
For a list of available commands use the '-l | --list' flag.
```

leftwm/doc/leftwm.1
```diff
@@ -0,0 +1,256 @@
...
+Mod + Shift + (up or down);Move the different windows in the current workspace
+Mod + Enter;Move selected window to the top of the stack in the current workspace
+Mod + Ctrl + (up or down);Switch between different layouts
+Mod + Shift + (left or right);Switch between different workspaces
+Mod + Shift + Enter;Open a terminal
+Mod + Ctrl + L;Lock the screen
+Mod + Shift + X;Exit LeftWM
+Mod + Shift + Q;Close the current window
+Mod + Shift + R;Reload LeftWM and its config
+Mod + p;Use dmenu to start application
```
VuiMuich, line 64:
```suggestion
Mod + (1-9) ; Switch to a desktop/tag
Mod + Shift + (1-9) ; Move the focused window to desktop/tag
Mod + W ; Switch the desktops for each screen. Desktops [1][2] changes to [2][1]
Mod + Shift + W ; Move window to the other desktop
Mod + (up or down) ; Focus on the different windows in the current workspace
Mod + Shift + (up or down) ; Move the different windows in the current workspace
Mod + Enter ; Move selected window to the top of the stack in the current workspace
Mod + Ctrl + (up or down) ; Switch between different layouts
Mod + Shift + (left or right) ; Switch between different workspaces
Mod + Shift + Enter ; Open a terminal
Mod + Ctrl + L ; Lock the screen
Mod + Shift + X ; Exit LeftWM
Mod + Shift + Q ; Close the current window
Mod + Shift + R ; Reload LeftWM and its config
Mod + p ; Use dmenu to start application
```

VuiMuich (changes requested):

VuiMuich (approved):

mautamu (approved):
Looks good, thank you!

VuiMuich:
Thanks @VentGrey this is a really great contribution!

VentGrey:
Glad to help this superb window manager! :heart:

Issue #576, opened 2021-11-16
leftwm-check did not complain for lack of permissions


Qark-dev:
**Describe the bug**
apparently, leftwm-check is supposed to complain whether up/down scripts for themes are lacking the permissions to run

**To Reproduce**
I was on gentoo and simply cloned the Red Moon theme, symlinked to ~/.config/leftwm/themes/current.

**Expected behavior**
leftwm-check is supposed to tell me that up/down scripts are lacking permissions.

**Screenshots**
![image](https://user-images.githubusercontent.com/80141845/142059036-ec9c4597-3e83-49a5-9bae-ddb8fe26fbc9.png)

**Environment:**
 - OS/Distro: Gentoo
 - Output of `leftwm-check`
 ```sh
:: LeftWM version: 0.2.9
:: LeftWM git hash: NONE
:: Loading configuration . . .
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK
```

**Additional context**
Users would be left wondering why leftwm wasn't loading their themes and this `leftwm-check` discrepancy fails to show the reason

VuiMuich:
Can confirm this with LeftWM git hash ec7ebca

PR #583, opened 2021-11-22
Fixed leftwm-check not running essential checks if verbose arg is false (#576)

Mik-Grzeg:
It was meant to fix bug #576, however the function which runs checks on `theme.toml` had the same issue. Hopefully I fixed that as well.

leftwm-check.rs
* if the verbose argument was false, `check_permissions` function was always returning Ok(filepath) 
* same thing with `check_theme_toml`, all the essential checks were inside `if verbose` block

### DIFF
--- a/leftwm/src/bin/leftwm-check.rs
+++ b/leftwm/src/bin/leftwm-check.rs
@@ -303,42 +303,43 @@ fn check_current_theme_set(filepath: &Option<PathBuf>, verbose: bool) -> Result<
 fn check_permissions(filepath: PathBuf, verbose: bool) -> Result<PathBuf> {
     let metadata = fs::metadata(&filepath)?;
     let permissions = metadata.permissions();
-    if verbose {
-        if metadata.is_file() && (permissions.mode() & 0o111 != 0) {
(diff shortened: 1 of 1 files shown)


VuiMuich:
Welcome Adventurer!
Thanks a lot for solving this.
May I request one little improvement, while you are already at this:
Now the check stops at the first error, but it would be great, if it would collect all mistakes and report them at once. That would be a superb topping or an already great first contribution.
Thanks.

Mik-Grzeg:
Hi there
Appreciate the nice greeting and I am happy for the opportunity to contribute.

Sure, I can handle this, it doesn't seem complicated.

Mik-Grzeg:
I managed to change it a bit, now it collects all the errors/mistakes that can occur in `check_theme`. I tried to make it similar to the functions that already do that, meaning `check_keybinds` and `check_workspace_ids`. 

There are few errors that won't allow further execution of the checks, but imo it is expected behaviour e.g. if `leftwm/themes` directory is missing, it won't run any checks on files that should be inside. 

Error messages are displayed like **`ERROR: <error message>`**, whereas ERROR is red.

Issue #560, opened 2021-11-01
[bug] focus gets lost in focus mode `Driven`


VuiMuich:
**Describe the bug**
I reported this already when reviewing #538, but the details are still a bit obscure.
In some situations when switching tag or when hiding a scratchpad window focus does not follow along.

I have seen this on both `Monocle` and `MainAndDeck` layouts with focus mode `Driven` I switched now to `ClickTo` to see if this happens there and with other layouts as well.

One suspicion is, that this happens when a scratchpad is still open on a hidden tag, but I can't confirm that yet.

Most of the time it is recoverable by un-hiding a scratchpad and moving focus away from it, then hiding it again, but sometimes this get the windows completely detached from the manager and is only recoverable by `SoftReload`.

So far my findings, will update as I gain more insight.

**To Reproduce**
still unclear, if I find reprocuible steps I'll update.

**Expected behavior**
Focus should never be lost when switching tags or un-/hiding scratchpad windows

**Environment:**
 - OS/Distro: Manjaro (but I believe this is issue is disro agnostic)
 - Output of `leftwm-check` should be irrelevant

hertg:
I've fixed some focus issues in #568. This issue might be related. Can you keep an eye on it and see if this issue here still occurs?

VuiMuich:
I believe I have steps to reproduce now:

1. configrue focus mode to `Driven`
2. Open some window
3. Open a scratchpad
4. go to another tag
5. call scratchpad to this tag by using the `ToggleScratchPad` command
6. got to the previous tag

Now the window of the previous tag is unfocused. 

7. Creating a new window and moving focus brings this window back
8. as well as calling the scratchpad back again
9. in some situations, still unclear on the condition, the window gets completely detached and only a `SoftReload` brings it back.

brownjohnf:
I'm seeing what seems like very similar behavior to this when using sticky windows. When switching between tags, focus just seems to go... nowhere. Launching a new window will restore focus to the new window, and then focus can be shifted around normally. Switching tags causes the same problem again. It doesn't seem to matter whether the layout is filled up, or there's only a single window.

I've also got 

```
focus_behaviour = 'ClickTo'
focus_new_windows = true
```

set in my config, but I'm unable to click-to-focus on a window. The behavior seems more like what's described by `Driven`, other than the issue above. I'm not sure if the 2 are related or not. I'm happy to file a new issue, if that makes more sense.

I'm on Arch linux, leftwm 0.3.0 from the AUR.

Issue #517, opened 2021-10-10
[bug] focus behaviour 'ClickTo' and floating windows don't mix well


hertg:
**Describe the bug**
I recently changed `focus_behaviour` to `ClickTo`, now I get weird freezes occasionally, which can only be resolved by switching tags via keybinding.

**To Reproduce**
I can't provide a 100% working step-by-step to reproduce it, because I still haven't found out which of my actions actually cause the bug.

Here is how I was able to reproduce it on at least two different computers within a few minutes of use:
1. Change `focus_behaviour` to `ClickTo`
2. Toggle one or more windows to be floating
3. Move floating windows around and resize them
4. The issue will eventually happen, sometimes right after toggling a window to be floating, sometimes after 20-30 seconds of moving stuff around

Something I observed when the issue occurs:
I use [unclutter](https://wiki.archlinux.org/title/Unclutter), which hides my mouse cursor after I haven't moved it around for some time. Funny enough, as soon as the issue occurs, I can still move the mouse cursor around and see it moving, but it will suddenly disappear because unclutter thinks it isn't moving.. Maybe that helps in tracking down the issue.

**Expected behavior**
No freezes :)

**Screenshots**
*None*

**Environment:**
 - OS/Distro: Arch
 - Output of `leftwm-check` (I'm on latest master)
   ```
   :: LeftWM version: 0.2.9
   :: LeftWM git hash: b4a0686
   :: Loading configuration . . .
       -> Configuration loaded OK
   :: Checking keybinds . . .
       -> All keybinds OK
   :: Checking environment . . .
       -> Environment OK
   :: Checking theme . . .
       -> Theme OK
   ```

**Additional context**
*None*

hertg:
I've also experienced this issue in `Sloppy` in the last few days/weeks. When moving/resizing floating windows (mostly popups) leftwm freezes occasionally, most of the time it catches up and applies the move/resize requests after a few seconds and continues normally, very rarely I need to SoftReload to unfreeze it though.

VuiMuich:
So I guess #585 might be related, right?

hertg:
> So I guess #585 might be related, right?

I don't think so. This one here happens with all kinds of windows. The comment about it being "mostly popups" is probably misleading because popups and my alacritty scratchpad are basically the only floating windows I use.

AethanFoot:
I have found one repeatable method demonstrated in the video:

[...]

PR #587, opened 2021-11-25
Various fixes

### LINKED ISSUE #225: XTerm takes a while to start up
An odd little issue I've run into with LeftWM.
For some reason, XTerm can take a while to start up, up to as long as 8 seconds in some cases.

This doesn't happen with anything else, just that.
For now I've been using Alacritty, but XTerm is my preferred terminal and I'd love to figure out what the issue is.

### LINKED ISSUE #585: IntelliJ autocomplete loses focus (Sloppy)
The autocomplete popup in IntelliJ makes the whole IDE lose focus if the mouse is hovered over the popup and then back out again.
autocomplete1

It's especially problematic when the mouse is already positioned where the popup will be rendered, because then it can just lose focus mid typing.

autocomplete2

**More information**
- If the mouse cursor never touches the autocomplete popup everything works as expected, accepting an option with <kbd>Enter</kbd> or <kbd>Tab</kbd> doesn't make the IDE lose focus.
- Everything works fine in the `ClickTo` focus behavior.

*I'm not sure how to get the xprops of the autocomplete popup, because if I change focus it gets destroyed* :upside_down_face:

### LINKED ISSUE #592: [feature] `SoftReload` should resore 'last focused workspace'
This is merely a tracking issue as this is created post fix simply to include it easily in the changelog for bump to `0.2.10`

AethanFoot:
Fixes #585, fixes #225 (happy accident), fixes #592, partial #560 (At least the steps found), partial #517. With #517 I feel that the issue with clickto specifically is (mostly) fixed, there are still times that we can freeze with the motion/resize and spam clicking and such. However if think these issues can be reproduced in other focusbehaviours with enough effort. In future we will want to optimise the events we read will in these modes to prevent them being "held up" by events less important to the user (e.g. visual responsiveness events), similar to what dwm does in their move/resize functions. This will be in the future though (especially as I currently don't have an idea of how to integrate it nicely into our structure).
Many Thanks!

### DIFF
--- a/leftwm-core/src/display_action.rs
+++ b/leftwm-core/src/display_action.rs
@@ -38,7 +38,7 @@ pub enum DisplayAction {
     DestroyedWindow(WindowHandle),
 
     /// Tell a window that it is to become focused.
-    WindowTakeFocus(Window),
+    WindowTakeFocus(Window, Option<WindowHandle>),
 
     /// Remove focus on any visible window by focusing the root window.
     Unfocus,
--- a/leftwm-core/src/display_action.rs
+++ b/leftwm-core/src/display_action.rs
@@ -38,7 +38,10 @@ pub enum DisplayAction {
     DestroyedWindow(WindowHandle),
 
     /// Tell a window that it is to become focused.
-    WindowTakeFocus(Window, Option<WindowHandle>),
+    WindowTakeFocus {
+        window: Window,
+        previous_handle: WindowHandle,
+    },
 
     /// Remove focus on any visible window by focusing the root window.
     Unfocus,
--- a/leftwm-core/src/display_action.rs
+++ b/leftwm-core/src/display_action.rs
@@ -40,7 +40,7 @@ pub enum DisplayAction {
     /// Tell a window that it is to become focused.
     WindowTakeFocus {
         window: Window,
-        previous_handle: WindowHandle,
+        previous_handle: Option<WindowHandle>,
     },
 
     /// Remove focus on any visible window by focusing the root window.
--- a/leftwm-core/src/display_action.rs
+++ b/leftwm-core/src/display_action.rs
@@ -44,7 +44,7 @@ pub enum DisplayAction {
     },
 
     /// Remove focus on any visible window by focusing the root window.
-    Unfocus,
+    Unfocus(Option<WindowHandle>),
 
     /// To the window under the cursor to take the focus.
     FocusWindowUnderCursor,
--- a/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
+++ b/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
@@ -49,6 +49,7 @@ impl<'a> From<XEvent<'a>> for Option<DisplayEvent> {
             }
             xlib::ButtonRelease => {
                 if xw.mode == models::Mode::Normal {
+                    xw.click_event = None;
                     return None;
                 }
                 Some(DisplayEvent::ChangeToNormalMode)
--- a/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
+++ b/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
@@ -47,13 +47,10 @@ impl<'a> From<XEvent<'a>> for Option<DisplayEvent> {
                 mod_mask &= !(xlib::Mod2Mask | xlib::LockMask);
                 Some(DisplayEvent::MouseCombo(mod_mask, event.button, h))
             }
-            xlib::ButtonRelease => {
-                if xw.mode == models::Mode::Normal {
-                    xw.click_event = None;
-                    return None;
-                }
-                Some(DisplayEvent::ChangeToNormalMode)
-            }
+            xlib::ButtonRelease => match xw.mode {
+                models::Mode::Normal => None,
+                _ => Some(DisplayEvent::ChangeToNormalMode),
+            },
 
             xlib::EnterNotify => from_enter_notify(xw, raw_event),
 
--- a/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
+++ b/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
@@ -18,7 +18,7 @@ pub struct XEvent<'a>(pub &'a mut XWrap, pub xlib::XEvent);
 impl<'a> From<XEvent<'a>> for Option<DisplayEvent> {
     fn from(x_event: XEvent) -> Self {
         let xw = x_event.0;
-        let mut raw_event = x_event.1;
+        let raw_event = x_event.1;
 
         match raw_event.get_type() {
             // new window is created
@@ -40,6 +40,7 @@ impl<'a> From<XEvent<'a>> for Option<DisplayEvent> {
             }
 
             xlib::ButtonPress => {
+                xw.click_event = Some(raw_event);
                 let event = xlib::XButtonPressedEvent::from(raw_event);
                 let h = WindowHandle::XlibHandle(event.window);
                 let mut mod_mask = event.state;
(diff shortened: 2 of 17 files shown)


hertg:
I can confirm that this fixes #585.

VuiMuich:
I think I can confirm fix for #560
not 100% sure though, as this one was kind of slippery while tracking down :wink:

cecton:
@AethanFoot I love how you "accidentally" fixed #225 which is a huge deal for me :rofl: 

So I tested and did you fix it? My answer is: yes and no.

The very first run of xterm in the session takes a few seconds but clearly much less than the 6 seconds of the socket timeout there is without your fixes. The next instances of xterm in the session are instantaneous. If I quit all the active xterm in the session and start a new one, it is instantaneous too.

For me it's good enough we can close #225 with this fix. Thanks a LOT @AethanFoot I will finally be able to start using leftwm actively.

I understand some are curious about why am I so insistent on xterm. I don't like to talk about it because it's such an unpopular opinion I have that people are making fun of it. Though you've been all very nice and I feel okay to share this.

I know this sounds weird because there are better alternatives to xterm out there but I have a preference for xterm because of its standardiness, it's use of xfont without antialiasing and its absence of GPU acceleration. All the modern terminals out there focus on GPU performance and beautifulness. Unfortunately for me, unless I make one myself, there will be no other xterm.

AethanFoot:
I found myself prefering st for its fast launching (compared to alacritty and such), and haven't found where it is noticeably slower in general use (also not fond of the idea of running a daemon constantly). And I can get it to look pretty (eventhough I had to fight it to get unicode and emojis to work completely). Im very into minimising without comprimisings looks (which is why I still use picom and lightdm, still 7 ish second system boot time).

AethanFoot:
The soft locks I could produce was due to the pointer grabbing be sync rather than async with grab_buttons. The reason it was sync was to allow replay_click to be simple, with async the ReplayPointer doesn't work. This means a less elegant way has to be used which I currently don't know if it causes issues.

hertg:
Just for my understanding: What does the replay-click do, why is it needed?

AethanFoot:
replay_click is so the click continues to the application, e.g. if you click on an unfocused app the click will not just disappear (will do the action expected).

leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
```diff
@@ -100,6 +108,17 @@ fn from_map_request(raw_event: xlib::XEvent, xw: &mut XWrap) -> Option<DisplayEv
     if let Some(trans) = trans {
         w.transient = Some(WindowHandle::XlibHandle(trans));
     }
+    // Initialise the windows floating we the pre-mapped settings.
```
cecton, line 111: typo?
```suggestion
    // Initialise the windows floating with the pre-mapped settings.
```

leftwm-core/src/display_servers/xlib_display_server/xwrap/getters.rs
```diff
@@ -160,6 +160,9 @@ impl XWrap {
             // Make sure that width and height are not smaller than the min values.
             xyhw.w = std::cmp::max(xyhw.w, xyhw.minw);
             xyhw.h = std::cmp::max(xyhw.h, xyhw.minh);
+            // Ignore the sizing if the sizing is set to 0.
+            xyhw.w = if xyhw.w == Some(0) { None } else { xyhw.w };
+            xyhw.h = if xyhw.h == Some(0) { None } else { xyhw.h };
```
cecton, line 165:
```suggestion
            xyhw.w = xyhw.w.filter(|w| w != 0);
            xyhw.h = xyhw.h.filter(|h| h != 0);
```

leftwm-core/src/display_action.rs
```diff
@@ -38,7 +38,7 @@ pub enum DisplayAction {
     DestroyedWindow(WindowHandle),
 
     /// Tell a window that it is to become focused.
-    WindowTakeFocus(Window),
+    WindowTakeFocus(Window, Option<WindowHandle>),
```
cecton, line 41: Not particularly fan of the tuple variant for this variant in particular because it is not obvious at all to understand what are the fields. I'd suggest converting to a struct variant:

```suggestion
    WindowTakeFocus {
        window: Window,
        previous_handle: WindowHandle,
    },
```

AethanFoot:
Will intergrate those changes properly when back on my desktop.

VuiMuich:
I just found a tiny issue:
when you issue the `SoftReload` from any tag other then the first, the windows on the first tag have no focus (focus mode `Driven, assume `ClickTo` might be affected as well).
Any action to enforce a focus on this tag like summoning a scratchpad or opening a new window, helps to catch these windows.
On the other hand, attempts to move focus or windows, detaches the windows from leftwm completely.

My guess is, that restoring the previous state somehow forces focus to the previously focused window on the focused tag, so either of these solutions should work:
- enforce focus loss and proper refocus in `SoftReload`
- restore focused tag in `SoftReload` (this one would probably nicer for the user eventually)

leftwm-core/src/display_servers/xlib_display_server/xwrap/mouse.rs
```diff
@@ -1,25 +1,52 @@
...
     /// Grabs the mouse clicks of a window.
-    pub fn grab_mouse_clicks(&self, handle: xlib::Window) {
+    pub fn grab_mouse_clicks(&self, handle: xlib::Window, is_focused: bool) {
         self.ungrab_buttons(handle);
-        self.grab_buttons(handle, xlib::Button1, self.mouse_key_mask);
-        self.grab_buttons(handle, xlib::Button1, self.mouse_key_mask | xlib::ShiftMask);
-        self.grab_buttons(handle, xlib::Button3, self.mouse_key_mask);
-        self.grab_buttons(handle, xlib::Button3, self.mouse_key_mask | xlib::ShiftMask);
+        if !is_focused {
+            self.grab_buttons(handle, xlib::Button1, xlib::AnyModifier, xlib::GrabModeSync);
```
hertg, line 13: i love this, now I can resize unfocused windows in `ClickTo` without left-clicking them first :)

leftwm-core/src/display_servers/xlib_display_server/xwrap/window.rs
```diff
@@ -63,8 +63,9 @@ impl XWrap {
                     let _ = self.move_cursor_to_window(handle);
                 }
                 if self.focus_behaviour == FocusBehaviour::ClickTo {
-                    self.ungrab_buttons(handle);
-                    self.grab_buttons(handle, xlib::Button1, xlib::AnyModifier);
+                    // self.ungrab_buttons(handle);
+                    // self.grab_buttons(handle, xlib::Button1, xlib::AnyModifier);
```
hertg, line 67:
```suggestion

```

hertg (approved):

hertg:
nice, this fixes the sizing issue I had with dialog boxes in NetExtender :)

AethanFoot:
SoftReload now restores the last tag on a workspace (if it still exists) then focuses the last focused window on that tag.
Thanks!

VuiMuich:
fixes #592 :grin:

AethanFoot:
Ok I feel I have gone a far as I can in this pr. If you treat it nicely it should not break :worried:. 
Thanks.

VuiMuich:
Aethan SneakyPete Rocket trying to silently adding WM_CLASS, just after stating 'this pr is done'. 
But I caught you red handed!

Nicely done, though!!! 🥳

hertg:
I think we can merge this. Or is there anything missing? 
I'm looking forward to having this in the main branch, and to finally get my dev-branch to include those fixes :stuck_out_tongue_closed_eyes:

PR #594, opened 2021-12-01
update edition to 2021, update dependencies

hertg:
Just a small maintenance PR.
Updated the Rust edition to `2021` and updated some dependencies where upgrades were available.

We might want to keep `chrono` on our watchlist, because `cargo audit` reports a vulnerability in the version of `time` that `chrono` is currently using. We should update dependencies using the vulnerable crate as soon as patches are available.

Links of interest:
- https://github.com/chronotope/chrono/issues/553
- https://github.com/chronotope/chrono/pull/578

---

*Output of `cargo audit`*:
```
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 376 security advisories (from /home/michael/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (132 crate dependencies)
Crate:         chrono
Version:       0.4.19
Title:         Potential segfault in `localtime_r` invocations
Date:          2020-11-10
ID:            RUSTSEC-2020-0159
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0159
Solution:      No safe upgrade is available!
Dependency tree: 
chrono 0.4.19
├── slog-term 2.8.0
│   ├── slog-envlogger 2.2.0
│   │   └── leftwm 0.2.10
│   └── leftwm 0.2.10
├── liquid-lib 0.23.0
│   └── liquid 0.23.0
│       └── leftwm 0.2.10
├── liquid-core 0.23.0
│   ├── liquid-lib 0.23.0
│   └── liquid 0.23.0
└── leftwm 0.2.10

Crate:         time
Version:       0.1.44
Title:         Potential segfault in the time crate
Date:          2020-11-18
ID:            RUSTSEC-2020-0071
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:      Upgrade to >=0.2.23
Dependency tree: 
time 0.1.44
└── chrono 0.4.19
    ├── slog-term 2.8.0
    │   ├── slog-envlogger 2.2.0
    │   │   └── leftwm 0.2.10
    │   └── leftwm 0.2.10
    ├── liquid-lib 0.23.0
    │   └── liquid 0.23.0
    │       └── leftwm 0.2.10
    ├── liquid-core 0.23.0
    │   ├── liquid-lib 0.23.0
    │   └── liquid 0.23.0
    └── leftwm 0.2.10

error: 2 vulnerabilities found!
```


AethanFoot (approved):
Ah nice, microsoft have now updated their rust version (when I initially pushed lefthk with 2021 clippy failed on their servers).

VuiMuich:
I guess we'll see, if the merge the bump on chrono until #587 is merged and either include this bump here, or bump chrono for 0.2.11

AethanFoot:
The hard thing is it seems to be a dependency of liquid-core, liquid-lib, and slog-term.

mautamu:
Here's one possibility if we're concerned about it:

[possibility.patch.txt](https://github.com/leftwm/leftwm/files/7637390/possibility.patch.txt)

mautamu:
Slog-term has updated to reduce the dependency on vulnerability, but have not yet published it: [79bfa8cf17d8f99600e9e7370d5b7f0d30f554bf](https://github.com/slog-rs/term/commit/79bfa8cf17d8f99600e9e7370d5b7f0d30f554bf)

Tried digging into liquid but got distracted.

AethanFoot:
Looking into this a bit more it look like slog-term has merged a pr that disables the problem part of chrono on latest commit. And there is an issue on the liquid-rs page starting a convo on it. 

@mautamu We seem to be doing the same thing :laughing:

0323pin:
Hi,
Nothing wrong with bumping to edition 2021 but, in that case the README will need to be updated again.
AFAIK, edition 2021 automatically requires rust >= 1.56

VuiMuich:
Haha, just saw your comment on the Void package PR and commented we might considering holding this PR back for a while if chances are that Void bumps in reasonable close future.

Also thanks for the heads up on the Readme mention!

0323pin:
@VuiMuich NetBSD has of now, rust-1.56.1, so I'm fine with the edition bump 😄

hertg:
Just merged latest `main` into my `update` branch.
Also ran another `cargo update` to update dependencies again, since the PR is already a bit old.

**Updates**
- Now fixes [RUSTSEC-2022-0006](https://rustsec.org/advisories/RUSTSEC-2022-0006) by updating `thread_local` to `>=1.1.4`
- Updated toolchain in `msrv` Github Action to `1.56.0`

**To-Do**
The following are still unpatched, any news on how we could address them? We can also do that in a separate PR if it requires code changes.
- [RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071)
- [RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159)

AethanFoot:
Could you change it to 1.56.1 as it fixed some security issues with 1.56.0

VuiMuich:
As far as I can tell [RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071) is fixed in the latest version of the `time` crate and the `chrono` crate appears to be a bit stalled currently.

We are currently using the `chrono` crate for exactly two things:
- setting the timestamp for the logfile name, when logging to `/tmp/leftwm/leftwm-<YYYYMMDDHHMM>.log`
- in `leftwm-state` when `{{ localtime }}` is parsed from the `liquid` template. This occurs exactly once across all themes in `leftwm-community-themes` and the example themes in the `basic_lemonbar` example. Tagging @arcolinuxz: does any of the arcolinux themes use this feature?

So imo we could deprecata the latter and I could try to switch to the `time` crate with #508.

0323pin:
> So imo we could deprecate the latter and I could try to switch to the time crate with https://github.com/leftwm/leftwm/pull/508.

As long as it doesn't break `lemonbar`.
`polybar` is full of linuxisms and I'd rather avoid it.

VuiMuich:
> As long as it doesn't break `lemonbar`. `polybar` is full of linuxisms and I'd rather avoid it.

So you actually use `leftwm-state` in with your `lemonbar` to display time? That would be the only change for you then and I guess [something like this](https://github.com/VentGrey/Epitaph/blob/90a4a799001d8896beff5226143dddaf34e9b3f4/scripts/lemonbar/time.c) should work for you too, right?

0323pin:
@VuiMuich 

> So you actually use leftwm-state in with your lemonbar to display time?

Yes, I do. Don't let that stop anything, I can go back to `citron` + `wired-notify` if needed. It was nice to have a bar for sometime, though.

> I guess [something like this](https://github.com/VentGrey/Epitaph/blob/90a4a799001d8896beff5226143dddaf34e9b3f4/scripts/lemonbar/time.c) should work for you too, right?

Guess so, I'll have a look.

mautamu:
Please use 1.56.1; see https://blog.rust-lang.org/2021/11/01/cve-2021-42574.html

mautamu:
> Could you change it to 1.56.1 as it fixed some security issues with 1.56.0

Thanks for catching this! I need to learn how to read :p

0323pin:
@VuiMuich using your suggestion above would mean I need to do a while loop with constant calls to keep it refreshing the time. Of course I could change from `{{ localtime }}` to the previously suggested `{{ "now" | date: "%Y-%m-%d %H:%M" }}` but, then I'm stuck with UTC.

Which brings me back to a notification-daemon as the best solution for the time issue.

VuiMuich:
> @VuiMuich using your suggestion above would mean I need to do a while loop with constant calls to keep it refreshing the time. Of course I could change from `{{ localtime }}` to the previously suggested `{{ "now" | date: "%Y-%m-%d %H:%M" }}` but, then I'm stuck with UTC.
> 
> Which brings me back to a notification-daemon as the best solution for the time issue.

Well, the leftwm-state based time only updates, when you have wm-interaction, so, when you stay consistently on a single window like browser or editor/ide etc. it can get out of syn, easily.
I'll have a look how to get US time formatting with the `time` crate and consider adding a dedicated `--time` flag for `leftwm-state`..

0323pin:
> I'll have a look how to get US time formatting with the time crate and consider adding a dedicated --time flag for leftwm-state..

Thanks for looking into it but, as said, don't worry too much if it turns out to be a PITA, after all I have time and date stamps on my shell prompt.
Also, I doesn't need to be US format 😄 
2022-02-06 would be better ;)

0323pin:
> Personally I am using a little utility program I call uhrwerk I hacked together in Rust

I remember you talked about it before, have a link to it? Guess it uses chrono and that's what you guys are trying to avoid, if I got it right.

VuiMuich:
> > I'll have a look how to get US time formatting with the time crate and consider adding a dedicated --time flag for leftwm-state..
> 
> Thanks for looking into it but, as said, don't worry too much if it turns out to be a PITA, after all I have time and date stamps on my shell prompt. Also, I doesn't need to be US format smile 2022-02-06 would be better ;)

Just thought if we keep it, we can as well make it right and allow tho user to setup the exact formatting..

VuiMuich:
@hertg I believe this would be a good place to bump `futures = "0.3.21"` as the `0.3.18` version we are using got *yanked*.

hertg:
> @hertg I believe this would be a good place to bump `futures = "0.3.21"` as the `0.3.18` version we are using got _yanked_.

Hmm, I could also change it to `0.3` so it would always be on the latest patch, right? Or do we have reasons not to do that with the `futures` crate?

VuiMuich:
> > @hertg I believe this would be a good place to bump `futures = "0.3.21"` as the `0.3.18` version we are using got _yanked_.
> 
> 
> 
> Hmm, I could also change it to `0.3` so it would always be on the latest patch, right? Or do we have reasons not to do that with the `futures` crate?

That's a good idea and I am not aware of a reason to hold a specific reason.
Speaking of that, if some package needs to stay on a specific version I'd say this should be noted in comments in the `Cargo.toml`, right?

carrascomj:
> Speaking of that, if some package needs to stay on a specific version I'd say this should be noted in comments in the `Cargo.toml`, right?

[Cargo deps are resolved  via semantic versioning](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility).

A crate that is specified as `dep="0.13.0"` will pull a version that is "semver-compatible" (0.13.* version). If you want to pin it to a specific version you can do it like `dep="=0.13.0"` and it will always pull that one; or `dep=">=0.14.0"`, etc. So more that comments (which is always nice), it is better to actually rely on Cargo for that.

hertg:
> > Speaking of that, if some package needs to stay on a specific version I'd say this should be noted in comments in the `Cargo.toml`, right?
> 
> [Cargo deps are resolved via semantic versioning](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility).
> 
> A crate that is specified as `dep="0.13.0"` will pull a version that is "semver-compatible" (0.13.* version). If you want to pin it to a specific version you can do it like `dep="=0.13.0"` and it will always pull that one; or `dep=">=0.14.0"`, etc. So more that comments (which is always nice), it is better to actually rely on Cargo for that.

Good point, thanks for noting. I see that my proposal of changing it to `0.3` doesn't make much sense then. I wasn't aware that `0.3.12` is the equivalent of `^0.3.12`. 

Btw. does cargo treat `0.x` the same? Because `0.x` can be a bit wild-west and [has undefined behavior in semver](https://semver.org/#spec-item-4). Given that most rust libs are in `0.x`, I'm expecting that it just treats it like `1.x+` :smile: 

@VuiMuich I'll just bump it to `0.3.21` for good measure, and then we can leave it at that.

VuiMuich:
I would like to merge this as soon as #723 is in main.
But I was thinking if we want to even bump to rust `1.59.0` then.

Any second thought, vetos etc.?

VuiMuich:
> > But I was thinking if we want to even bump to rust 1.59.0 then.
> 
> Please don't. I'm using Rust-1.59.0 but, I'm usually one version ahead of NetBSD main, which tends to keep Rust one version behind upstream to avoid unexpected breakage.
> 
> Of course, if you are not planning on releasing leftwm-0.3.0 within the coming 6 weeks, it's fine to bump to 1.59.0

Sure lets wait till NetBSD is on `1.59.0` then.
Will merge right after #723

0323pin:
@VuiMuich Awesome, just waiting for `lefthk` also and I'll re-build my install.

0323pin:
@VuiMuich Please feel free to bump rust to 1.59.0 when you want, it has been merged on NetBSD yesterday. I've moved my own system to 1.60.0 now.

PR #602, opened 2021-12-04
Allow scratchpad sizing at 100% (aka "Q3 console")

cecton:
I have no idea why those have been limited to 90% but I can assure you it works fine. I made a scratchpad with xterm at 0,0 of size 100%x50% and it just looks like quake3

:thinking: .... I wonder if this could be related to window border and gaps... In my config I have 0 border and 0 gap so I can actually use 100% safely.

cc @tramhao @VuiMuich @hertg @AethanFoot 

Related to #493

### DIFF
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -440,7 +440,7 @@ fn sane_dimension(config_value: Option<Size>, default_percent: f32, max_pixel: i
     match config_value {
         Some(size) => match size {
             Size::Percentage(percentage) if (0.0..=1.0).contains(&percentage) => {
-                size.into_absolute(100.0) as i32 * max_pixel / 100
+                size.into_absolute(max_pixel as f32) as i32
             }
             Size::Pixel(pixel) if (0..=max_pixel).contains(&pixel) => pixel,
             _ => (default_percent * max_pixel as f32) as i32,
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -439,10 +439,10 @@ pub fn scratchpad_xyhw(xyhw: &Xyhw, scratch_pad: &ScratchPad) -> Xyhw {
 fn sane_dimension(config_value: Option<Size>, default_percent: f32, max_pixel: i32) -> i32 {
     match config_value {
         Some(size) => match size {
-            Size::Percentage(percentage) if (0.0..0.9).contains(&percentage) => {
+            Size::Percentage(percentage) if (0.0..=1.0).contains(&percentage) => {
                 size.into_absolute(100.0) as i32 * max_pixel / 100
             }
-            Size::Pixel(pixel) if (0..(max_pixel as f32 * 0.9) as i32).contains(&pixel) => pixel,
+            Size::Pixel(pixel) if (0..=max_pixel).contains(&pixel) => pixel,
             _ => (default_percent * max_pixel as f32) as i32,
         },
         _ => (default_percent * max_pixel as f32) as i32,
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -438,13 +438,10 @@ pub fn scratchpad_xyhw(xyhw: &Xyhw, scratch_pad: &ScratchPad) -> Xyhw {
 
 fn sane_dimension(config_value: Option<Size>, default_percent: f32, max_pixel: i32) -> i32 {
     match config_value {
-        Some(size) => match size {
-            Size::Percentage(percentage) if (0.0..=1.0).contains(&percentage) => {
-                size.into_absolute(max_pixel as f32) as i32
-            }
-            Size::Pixel(pixel) if (0..=max_pixel).contains(&pixel) => pixel,
-            _ => (default_percent * max_pixel as f32) as i32,
-        },
+        Some(size @ Size::Percentage(percentage)) if (0.0..=1.0).contains(&percentage) => {
+            size.into_absolute(max_pixel as f32) as i32
+        }
+        Some(Size::Pixel(pixel)) if (0..=max_pixel).contains(&pixel) => pixel,
         _ => (default_percent * max_pixel as f32) as i32,
     }
 }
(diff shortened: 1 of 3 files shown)


leftwm-core/src/handlers/window_handler.rs
```diff
@@ -439,10 +439,10 @@ pub fn scratchpad_xyhw(xyhw: &Xyhw, scratch_pad: &ScratchPad) -> Xyhw {
 fn sane_dimension(config_value: Option<Size>, default_percent: f32, max_pixel: i32) -> i32 {
     match config_value {
         Some(size) => match size {
-            Size::Percentage(percentage) if (0.0..0.9).contains(&percentage) => {
+            Size::Percentage(percentage) if (0.0..=1.0).contains(&percentage) => {
                 size.into_absolute(100.0) as i32 * max_pixel / 100
```
hertg, line 443: Unless I'm misunderstanding the code, I think this can be changed as follows:
```suggestion
                size.into_absolute(max_pixel as f32) as i32
```

I know you didn't add this line, but I just saw that. :)

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -439,10 +439,10 @@ pub fn scratchpad_xyhw(xyhw: &Xyhw, scratch_pad: &ScratchPad) -> Xyhw {
...
             }
-            Size::Pixel(pixel) if (0..(max_pixel as f32 * 0.9) as i32).contains(&pixel) => pixel,
+            Size::Pixel(pixel) if (0..=max_pixel).contains(&pixel) => pixel,
             _ => (default_percent * max_pixel as f32) as i32,
         },
         _ => (default_percent * max_pixel as f32) as i32,
```
hertg, line 448: And again... sorry to point out lines this PR didn't even touch.
Is there an easy way to prevent the duplicate code on line 446 and 448?

hertg (commented):
This change makes sense. I have no strong opinion about it, because I don't know why this limit is there. I think @AethanFoot or @VuiMuich probably know more about that.

I pointed out two things I just saw that aren't even related to your code changes, but they stood out to me. Sorry about that :smile:

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -439,10 +439,10 @@ pub fn scratchpad_xyhw(xyhw: &Xyhw, scratch_pad: &ScratchPad) -> Xyhw {
 fn sane_dimension(config_value: Option<Size>, default_percent: f32, max_pixel: i32) -> i32 {
     match config_value {
         Some(size) => match size {
-            Size::Percentage(percentage) if (0.0..0.9).contains(&percentage) => {
+            Size::Percentage(percentage) if (0.0..=1.0).contains(&percentage) => {
```
AethanFoot, line 442: Does this:
```suggestion
            Size::Percentage(percentage) @ 0.0..=1.0 => {
```
or this:
```suggestion
            Size::Percentage(percentage @ 0.0..=1.0) => {
```
Work?

leftwm-core/src/handlers/window_handler.rs
AethanFoot: I don't think so, unless it is refactored into if let... maybe.

AethanFoot:
I'm not sure why there was a limit in place, it may have just made sense at the time.

leftwm-core/src/handlers/window_handler.rs
cecton: Good catch

VuiMuich:
I think it was simply assumed, that a floating window with the size of the whole workspace is undesirable.

cecton:
> I think it was simply assumed, that a floating window with the size of the whole workspace is undesirable.

wat! i want scratchpad only to make quake3 console 😁

leftwm-core/src/handlers/window_handler.rs
cecton: Wow I didn't know that syntax could be used like that. Really cool!!

Unfortunately it doesn't seem to work because of the floats... it would have work for integers though!

```
warning: floating-point types cannot be used in patterns
   --> leftwm-core/src/handlers/window_handler.rs:442:36
    |
442 |             Size::Percentage(0.0..=1.0) => {
    |                                    ^^^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #41620 <https://github.com/rust-lang/rust/issues/41620>
```

The one for Pixel doesn't work because there is a variable inside:

```
error[E0080]: runtime values cannot be referenced in patterns
   --> leftwm-core/src/handlers/window_handler.rs:445:37
    |
445 |             Size::Pixel(pixel @ 0..=max_pixel) => pixel,
    |  
```

I would gladly replace by `if pixel <= max_pixel` but this won't work either because max_pixel is i32.

leftwm-core/src/handlers/window_handler.rs
cecton: Fixed in 588faba

AethanFoot (approved):
Looks great!

cecton:
Wait a sec

VuiMuich:
Ahh, now I have to install `openarena` for nostalgia 😆

cecton:
Done, ready for review

AethanFoot (approved):

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -436,15 +436,12 @@ pub fn scratchpad_xyhw(xyhw: &Xyhw, scratch_pad: &ScratchPad) -> Xyhw {
...
     match config_value {
-        Some(size) => match size {
-            Size::Percentage(percentage) if (0.0..0.9).contains(&percentage) => {
-                size.into_absolute(100.0) as i32 * max_pixel / 100
-            }
-            Size::Pixel(pixel) if (0..(max_pixel as f32 * 0.9) as i32).contains(&pixel) => pixel,
-            _ => (default_percent * max_pixel as f32) as i32,
-        },
-        _ => (default_percent * max_pixel as f32) as i32,
+        Some(size @ Size::Ratio(percentage)) if (0.0..=1.0).contains(&percentage) => {
```
cecton, line 441: oops, forgot to rename that one

AethanFoot: If you mean the percentage we can change it in another pr as a passing thing.

cecton: thanks ^_^

PR #605, opened 2021-12-05
feature: implement FocusWindowTop command

kid:
This is a simple thing I was missing coming from XMonad, so figured I
would add it as it turned out to be fairly simple to do.

I'm not super fan of the implementation though, maybe the helper method
should be renamed to something else, or a new method added in
`command_handler`, e.g. `focus_window_top`.

Ideally, I would like to add the ability to go back to the previously
focused windowed by pressing the key again, just not sure how to do that
yet.

### DIFF
--- a/leftwm-core/src/config/mod.rs
+++ b/leftwm-core/src/config/mod.rs
@@ -82,7 +82,7 @@ impl Config for TestConfig {
         unimplemented!()
     }
     fn focus_behaviour(&self) -> FocusBehaviour {
-        FocusBehaviour::Sloppy
+        FocusBehaviour::ClickTo
     }
     fn mousekey(&self) -> String {
         "Mod4".to_string()
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -575,7 +575,7 @@ fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
     let next = state
         .windows
         .iter()
-        .find(|x| x.tags.contains(&tag) && !x.is_unmanaged())
+        .find(|x| x.tags.contains(&tag) && !x.floating() && !x.is_unmanaged())
         .map(|w| w.handle);
 
     match (next, cur, prev) {
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -66,7 +66,7 @@ fn process_internal<C: Config, SERVER: DisplayServer>(
         Command::FocusPreviousTag => focus_tag_change(state, -1),
         Command::FocusWindowUp => move_focus_common_vars(focus_window_change, state, -1),
         Command::FocusWindowDown => move_focus_common_vars(focus_window_change, state, 1),
-        Command::FocusWindowTop => move_focus_common_vars(focus_window_top, state, 0),
+        Command::FocusWindowTop(toggle) => focus_window_top(state, *toggle),
         Command::FocusWorkspaceNext => focus_workspace_change(state, 1),
         Command::FocusWorkspacePrevious => focus_workspace_change(state, -1),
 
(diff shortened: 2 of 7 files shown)


leftwm-core/src/utils/helpers.rs
```diff
@@ -110,6 +110,10 @@ pub fn relative_find<T, F>(
 where
     F: Fn(&T) -> bool,
 {
+    if shift == 0 {
+        return list.get(0);
+    }
+
```
AethanFoot, line 116: I would avoid adding specific code to generic helpers. This may best be placed at the top of focus_window_change, where you focus the first non floating window in to_reorder (make sure if you exit early that to_reorder is appended back into state.windows, otherwise they will get lost and leftwm will crash).

AethanFoot (commented):
Nice feature, for the switching back you should be able to check if the top window is focused and then focus the prior focused window (stored in the focus manager) if that window is on that tag (otherwise if you click it accidently you may switch to another workspace or lose focus). Sorry if the writing is clumbsy, currently on my phone.
Thanks!

leftwm-core/src/config/mod.rs
```diff
@@ -82,7 +82,7 @@ impl Config for TestConfig {
         unimplemented!()
     }
     fn focus_behaviour(&self) -> FocusBehaviour {
-        FocusBehaviour::Sloppy
+        FocusBehaviour::ClickTo
```
kid, line 85: Had to change this, otherwise the focus is not immediately changed in the tests

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -567,6 +568,28 @@ fn focus_window_change(
...
+        .iter()
+        .find(|x| x.tags.contains(&tag) && !x.is_unmanaged())
+        .map(|w| w.handle);
+
+    match (next, cur, prev) {
+        (Some(next), Some(cur), Some(prev)) if next == cur && toggle => {
+            Some(handle_focus(state, prev))
+        }
+        (Some(next), Some(cur), _) if next != cur => {
+            state.focus_manager.tags_last_window.insert(tag, cur);
```
kid, line 586: Shouldn't this be done by `handle_focus`?

kid:
Alright, this now ready for review.

I'moved away form using `move_focus_command_vars` as it wasn't working with the toggle flag, and it caused me multiple issues where the state would end up broken because a `?` would return `None` / `Err()` early before re-appending all windows into the state, which didn't make sense anyway when just changing the focus

leftwm-core/src/handlers/command_handler.rs
AethanFoot: Yes this would be handled by focus_window which is called by handle_focus.

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -567,6 +568,28 @@ fn focus_window_change(
...
+fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
+    let tag = state.focus_manager.tag(0)?;
+    let cur = state.focus_manager.window(&state.windows).map(|w| w.handle);
+    let prev = state.focus_manager.tags_last_window.get(&tag).copied();
+    let next = state
+        .windows
+        .iter()
+        .find(|x| x.tags.contains(&tag) && !x.is_unmanaged())
+        .map(|w| w.handle);
+
```
AethanFoot, line 580: You may want to do && !x.floating(), otherwise it will move focus to the top floating window if there are any (unless that is desired?).

leftwm-core/src/handlers/command_handler.rs
kid: It does not seems to be the case, `focus_window` only add it to `window_history` and not `tags_last_windows`, and the test fails If I remove this

AethanFoot: Oh yeah sorry misread it. At some point we should probably add it there, as currently we save of only when changing tag or reloading.

leftwm-core/src/handlers/command_handler.rs
kid: Fixed that, as that is definitely not the desired effect of this feature imo

AethanFoot:
Would you be okay to add an external command as well. You should just need to add a bit to command_pipe.rs under utils.

Issue #614, opened 2021-12-07
Popup window - pamac-all - mint-stick-git


arcolinuxz:
**Describe the bug**
Going to preferences in Pamac gives us  a very small window - was unable to resize it.
Solved it by going super + F = fullscreen

**To Reproduce**
Arch Linux or ArcoLinux and install pamac-all - open up preferences

**Expected behavior**
a bit bigger please

**Screenshots**
![Screenshot from 2021-12-07 20-57-06](https://user-images.githubusercontent.com/36679406/145097696-7e827b18-94b3-4759-80da-18bb94c39754.png)

**Environment:**
 - OS/Distro: ArcoLinux
 - Output of `leftwm-check`
 `[erik@ArcoLinux-IsoBuilder ~]$ leftwm-check
:: LeftWM version: 0.2.10
:: LeftWM git hash: 30b331d
:: Loading configuration . . .
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK
[erik@ArcoLinux-IsoBuilder ~]$`

**Additional context**
new package build today on Arch Linux - leftwm - 1:0.2.10-1 based on your tags

arcolinuxz:
Had some issues with mintstick as well.

The problem there is the popup to get root permissions.

Sometimes the input for the root password is not available.

Mintstick - r221-e7b1893-1

I have seen you guys have been working on a solution for these windows.
I though I report already this.

![Screenshot from 2021-12-07 21-02-33](https://user-images.githubusercontent.com/36679406/145098176-dd19f90f-d331-4abe-9259-4a2e9b8d9af1.png)

AethanFoot:
The one with pamac is something we are looking to fix, if the dialog seems too small we can make it bigger. The authenticate one from a small testing seems the polkit gui seems to give a smaller size with a configure request which then gets resized once an update loop is called to the min size which allows the bar to appear. A lot of these issues are because the sizes and information we are sent isn't ideal, these are generally going to be easy to fix. 

The mintstick one with the main window be small and non-resizable should be raised with them as they are setting the min and max size to the same (and too small and continuously changing) which [indicates the window should be non-resizable](https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#NORESIZE). We can address this but means we would need to ignore all of their property notifies as the "fixed size" is continuously changing depending on the window state.

[...]

PR #615, opened 2021-12-08
Dialog fixes

AethanFoot:
Improvements to #614.

### DIFF
--- a/leftwm-core/src/display_action.rs
+++ b/leftwm-core/src/display_action.rs
@@ -69,4 +69,7 @@ pub enum DisplayAction {
 
     /// SoftReload keygrabs, needed when keyboard changes.
     ReloadKeyGrabs(Vec<Keybind>),
+
+    /// Configure a xlib window.
+    ConfigureXlibWindow(Window),
 }
--- a/leftwm-core/src/display_event.rs
+++ b/leftwm-core/src/display_event.rs
@@ -19,5 +19,6 @@ pub enum DisplayEvent {
     ResizeWindow(WindowHandle, i32, i32),
     ScreenCreate(Screen),
     SendCommand(Command),
+    ConfigureXlibWindow(WindowHandle),
     ChangeToNormalMode,
 }
--- a/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
+++ b/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
@@ -210,7 +210,6 @@ fn from_configure_request(xw: &XWrap, raw_event: xlib::XEvent) -> Option<Display
         Mode::Normal => {}
     };
     let event = xlib::XConfigureRequestEvent::from(raw_event);
-    log::info!("Configure Request");
     // If the window is not mapped, configure it.
     if !xw.managed_windows.contains(&event.window) {
         let window_changes = xlib::XWindowChanges {
--- a/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
+++ b/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs
@@ -100,7 +100,7 @@ fn from_map_request(raw_event: xlib::XEvent, xw: &mut XWrap) -> Option<DisplayEv
 
     // Build the new window, and fill in info about it.
     let mut w = Window::new(handle, name, pid);
-    w.r#type = r#type;
+    w.r#type = r#type.clone();
     w.set_states(states);
     if let Some(trans) = trans {
         w.transient = Some(WindowHandle::XlibHandle(trans));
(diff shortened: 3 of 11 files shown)


VuiMuich:
I beleive it is a regression of this PR, that when closing dialogs the focus moves to the main window, but should return to the parent.
I was thinking if the focus should return to the previous focused window, but I believe returning to the parent would feel more natural.

AethanFoot:
I had an idea to create a folder called patches which would store fixes like these that would be important for the previous release. Then say in the PKGBUILD we would go through the patches applying them to the version git download (like in the PKGBUILD in put in #614 expect how ever many patches there are). This would help reduce the effort needed to add important bug patches (each package being maintained would need updating in their own way) which would automatically be added.

AethanFoot:
@lex148 How easy is it to add this as a patch on crates.io, the aur is easy enough. Tagging @0323pin, @rigoletto-freebsd as I know you guys maintain some packages.

0323pin:
@AethanFoot Sorry for the delay, AFK yesterday.
It's doable but, it will take some time.

Isn't a point release a better solution?
Anyway, I can do it, just let me know if you go ahead with patching on crates.io and on Arch.

AethanFoot:
Okay I have added a branch that has 0.2.10 on which then can receive point release patches. This should hopefully make it easier for lex to create a tag with just the needed patches.

0323pin:
> Okay I have added a branch that has 0.2.10 on which then can receive point release patches. This should hopefully make it easier for lex to create a tag with just the needed patches.

So, if I understood this correctly, a v0.2.10.1 could show up soon?
Or, should I start patching _our_ v0.2.10 to include the changes in these 5 commits?

AethanFoot:
A 0.2.10.1 will hopefully become available based on the patched branch. I been thinking about this system for a while to give a more normal bug release cycle, I was just worried about putting a burden on lex but this one seems quite important.

0323pin:
> A 0.2.10.1 will hopefully become available based on the patched branch. I been thinking about this system for a while to give a more normal bug release cycle, I was just worried about putting a burden on lex but this one seems quite important.

I see, only one person is allowed to do releases.
I'll keep an eye on it, thanks.

OT, I'm considering using the development branch myself and jumping to git-HEAD.

VuiMuich:
Maybe in order to not break semantic versioning it would be better, to release 0.2.11 early and move the open milestone items to 0.2.12?

AethanFoot:
Actually looking as semantic versioning, we don't really follow it exactly otherwise it would be better without the initial 0, e.g. 2.10.0. This would make more sense here as well as #292, where the major version number change indicates a breaking change. It also says that you should consider "1.0.0" be a thing when your system is used in production (https://semver.org/).

VuiMuich:
Hmm, I thought leftwm follows semantic versioning and having the leading 0 is still considered `beta`.
I believe I somewhere read, crates.io requiringin or at least strongly recommending semantic versioning so I concluded this for leftwm..

AethanFoot:
I remember seeing that as well I think. I do feel dropping the 0 makes our releases align a lot more (especially with the need of this patch iteration), then 0.3.0 would be 3.0.0 which aligns with a breaking change, 0.2.x would be 2.x.0 which aligns with features additions, then 0.2.x.x would be 2.x.x which aligns with a needed patch. I hope my thought makes sense, it doesn't really effect much.

hertg:
I know I'm very late to the game, but wasn't able to catch up on the Github activity until now.

Just wanted to leave this comment here: The current versioning of `0.x` is perfectly fine in semver. It just means that the versioning is a bit wild-west and breaking changes are to be expected at any time.

> *Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.*
&ndash; https://semver.org/

Which I think is fair and kind of fits the current state of the WM. But as mentioned in the Discord, it's a great goal to bump to `1.0.0` soon, as the WM has increased a lot in stability, but we probably should get in the breaking changes that are on the roadmap first.

Some things need to be taken into consideration though, and this won't be easy: If `leftwm-core` should be usable as a library to allow custom implementations of it, avoiding breaking changes becomes very hard. In the context of a library, changing any structs, properties, method names, or method parameters that can be used from outside `leftwm-core` is by strict semver definition considered a breaking change. So we may need to specify what we actually mean with "breaking changes". Breaking changes from a user-standpoint (like config-option renaming, config-lang change, ..) or breaking changes from a library standpoint (like renaming a `pub struct` in leftwm-core).

Maybe this will be the point where leftwm and leftwm-core start to diverge in their versions, because a breaking change means something different in both these crates.

I still see some potential to decouple some logic in the source code, like my current attempt to extract the layout-logic from other WM-related stuff. There are probably other examples where decoupling the logic would be beneficial, but this will happen gradually and one-step at a time, and I'm not sure yet how hard that will be to achieve in strict semver.

Issue #611, opened 2021-12-06
Regression for Layout LeftWiderRightStack and RightWiderLeftStack


gitwipo:
Hello,

thanks for creating this great window manager. I use it as my daily driver for quite some while. It is nice to follow the discussions and where it is going. I do update every week regulary and I don't regret doing so.

**Describe the bug**

I do see a regression in  the layout `LeftWiderRightStack` because for me it is the same as `MainAndVertStack`.

**To Reproduce**
1. Change layout to `MainAndVertStack`
2. Now switch to `LeftWiderRightStack` -> The main stack should be wider

**Expected behavior**
The layout should change to: `LeftWiderRightStack`

**Environment:**
 - Distro: Debian Bullseye
 - Leftwm build from git
 - config: `layout_mode="Tag"`
 - Output of `leftwm-check`:
```
leftwm-check
:: LeftWM version: 0.2.10
:: LeftWM git hash: 5fc861e
:: Loading configuration . . .
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
ERROR: No theme folder or symlink `current` found.
```

**Additional context**
The layout was no longer working around end of October. If needed I can check which weekly build still worked for me.
I'm using leftwm without any bar or theme.

VuiMuich:
Welcome Adventurer!

may I ask how many windows you have open, in your "steps to reproduce" case?
You need at least three to make out the difference.

I give you that though, that `LeftWiderRightStack` should have a different `main_width` then the 50% it has now. As this makes it the same as `MainAndVertStack`.

gitwipo:
Hello, 

thanks for the fast feedback. 

I did try to switch to the layout with two or more windows open on the tag. I did see the difference already with two windows. 

Kind regards,

hertg:
@gitwipo Just a quick question, are you aware that you can change the main window width on the fly with `IncreaseMainWidth` and `DecreaseMainWidth`?

I configured a keybind in my config as shown below. I use this feature pretty extensively and the `_Wider_` and `_Balanced_` layouts became pretty much obsolete to me.

```toml
[[keybind]]
command = "IncreaseMainWidth"
value = "5"
modifier = ["modkey"]
key = "l"

[[keybind]]
command = "DecreaseMainWidth"
value = "5"
modifier = ["modkey"]
key = "h"
```

gitwipo:
Hello, 

[...]

PR #616, opened 2021-12-09
fix fix RightWiderLeftStack and LeftWiderRightStack not having the right … 

blackdragon2447:
fixed fix RightWiderLeftStack and LeftWiderRightStack not having the right width and an issue where Leftwm would refuse to start when running under Xephyer cause of a devide by zero error...


VuiMuich:
closes #611

can merge after a bit of clean up

PR #617, opened 2021-12-11
Refactor: remove disable_current_tag_swap from Config trait

cecton:
This PR refactor slightly the Config trait by removing the method `disable_current_tag_swap`.

My reasoning is that this is the kind of parameter that should be on the command and not on the configuration. So I updated `leftwm_core::Command::GotoTag` to take the `swap` parameter in argument and I removed `disable_current_tag_swap` from the `State` and from the `Config` trait.

For the sake of retro-compatibility I kept the field `disable_current_tag_swap` on `leftwm`'s config file ~~and I added 2 new commands: `GotoTagSwap` and `GotoTagNoSwap`, those are for explicitness~~. `GotoTag` is preserved as it used to be and use the parameter `disable_current_tag_swap` from the configuration file. ~~My hope is to deprecate `GotoTag` in favor of their explicit alternatives (`GotoTagSwap` and `GotoTagNoSwap`).~~

## Changes
- Remove `disable_current_tag_swap` from `Config` trait (the point of this PR)
- Split `config.rs` to smaller files: I created a module and directory `config` and moved the `Default` impl to `config/default.rs` inside that module.
- Move configuration checks to `config/checks.rs`: this lighten `leftwm-check.rs` a bit
- Remove Option from field `value` in `Config`: there is no need to have an option wrapping a String. The empty String is the "none" case. You want an `Option<String>` only if you intend to differentiate between `None` and `Some(empty_string)`. This modification changes a bit the behavior of the config file because now `value = ""` will be considered the same as not providing value at all (which should be desirable in all cases but correct me if I'm wrong).
- Replace the `HashMap` with an `HashSet` for the check of conflicting keybinding: if you don't need the value of your hashmap, hashset is a better choice
- Breaking change: there is a piece of code here I'm not 100% sure what it is supposed to do `leftwm-core/src/display_servers/xlib_display_server/event_translate_client_message.rs`. I believe the behavior there should be "no swap" but if someone can confirm it would be nice. I added a comment at the exact spot.

[edit: remove explicit commands GoToTagSwap and GoToTagNoSwap]


leftwm-core/src/display_servers/xlib_display_server/event_translate_client_message.rs
```diff
@@ -15,7 +15,10 @@ pub fn from_event(xw: &XWrap, event: xlib::XClientMessageEvent) -> Option<Displa
         let value = event.data.get_long(0);
         match usize::try_from(value) {
             Ok(index) => {
-                return Some(DisplayEvent::SendCommand(Command::GotoTag(index + 1)));
+                return Some(DisplayEvent::SendCommand(Command::GotoTag {
+                    tag: index + 1,
+                    swap: false,
+                }));
```
cecton, line 21: I'm not 100% sure if this should be swap off or swap on.

In the current version on main of leftwm, this is currently swapon by default and dependent on the configuration field `disable_current_tag_swap`. But I have a hunch that this should actually be always swap off. If someone can confirm or infirm my hypothesis it would be great :pray:

AethanFoot: I would say it should stay swapon to prevent the majority of setups changing.

cecton: > I would say it should stay swapon to prevent the majority of setups changing.

Do you actually know when/how this code is ran? Or you're just assuming this because the swapping is the default value in config?

Either way it will change the existing setups. But I suspect this swapping here was not intentional.

But as I said, I have a good hunch the swapping here was not introduced intentionally... 

The swapping feature has been introduced in Tue Feb 23 07:23:36 2021 -0500 (a57c91a034a5649531b46bc8d192c58a3d7db277)

And that particular code dates from long time ago, at least Wed Mar 13 16:46:23 2019 -0500 (6d6ee80179c1761a7fa14b6e8536e06639d05d3d)

It's mostly triggered by something in `leftwm-core/src/display_servers/xlib_display_server/event_translate.rs`. It's hard to tell where exactly. Need to investigate more (or someone who knows exactly)

AethanFoot: > Do you actually know when/how this code is ran?

What do you mean by this? External calls from programs e.g. ```xdotool getwindowfocus windowactivate```?

AethanFoot: Oh wait I see what you mean now. This is called by external programs like ```xdotool getwindowfocus windowactivate
```. I am really bad at reading today.

Comment (AethanFoot) [reply leftwm-core/src/display_servers/xlib_display_server/event_translate_client_message.rs:21]: Which request to the window manager to swap the tags (this one brings the tag with the focused window to the screen). So maybe this should be false.

Comment (cecton) [reply leftwm-core/src/display_servers/xlib_display_server/event_translate_client_message.rs:21]: Ahh! xD I knew I had a good hunch. I have no idea what's xdotool, it sounds like something cool though

**Comment by AethanFoot** [comment]
Should this have a default flag of true? This will make less configs break. If not we might as well change ```GotoTag``` to ```GoToTag``` here if this is going to be breaking.

**Comment by cecton** [comment]
> Should this have a default flag of true? This will make less configs break. If not we might as well change `GotoTag` to `GoToTag` here if this is going to be breaking.

I already preserve compatibility for existing configs:
> For the sake of retro-compatibility I kept the field `disable_current_tag_swap` on `leftwm`'s config file and I added 2 new commands: `GotoTagSwap` and `GotoTagNoSwap`, those are for explicitness. `GotoTag` is preserved as it used to be and use the parameter `disable_current_tag_swap` from the configuration file. My hope is to deprecate `GotoTag` in favor of their explicit alternatives (`GotoTagSwap` and `GotoTagNoSwap`).

File: leftwm/src/config/mod.rs [hunk 2]
```diff
@@ -27,32 +29,47 @@ const STATE_FILE: &str = "/tmp/leftwm.state";
...
+            }
             BaseCommand::ToggleFullScreen => leftwm_core::Command::ToggleFullScreen,
             BaseCommand::ToggleSticky => leftwm_core::Command::ToggleSticky,
-            BaseCommand::GotoTag => leftwm_core::Command::GotoTag(
-                usize::from_str(&k.value.context("missing index value for GotoTag")?)
-                    .context("invalid index value for GotoTag")?,
-            ),
+            BaseCommand::GotoTag => leftwm_core::Command::GotoTag {
+                tag: usize::from_str(&self.value).context("invalid index value for GotoTag")?,
+                swap: !config.disable_current_tag_swap,
```
cecton, line 62: @AethanFoot :point_up: for compatibility for existing configs.

AethanFoot:
I am not with it today dang :facepalm:. Shall we make them GoToTagSwap/NoSwap to prevent them needing changing in the future. We can then add a depreciated tag to GotoTag in the wiki and such, and maybe add a check to leftwm-check which checks for depreciated values.

cecton:
> Shall we make them GoToTagSwap/NoSwap to prevent them needing changing in the future.

I already made those. I did not do anything to mark GotoTag deprecated though. I think we should.

AethanFoot:
Sorry I meant to capitalise the 't' in "to" e.g. ```GotoTagSwap``` -> ```GoToTagSwap```

cecton:
> Sorry I meant to capitalise the 't' in "to" e.g. `GotoTagSwap` -> `GoToTagSwap`

Ah! Yes I don't have much preference on that.... what's the common capitalization rule for that word? Maybe it's one of those word where the capitalization is done like a single word. https://en.wikipedia.org/wiki/Goto

As you prefer really.

AethanFoot:
Tbh I don't mind either but this is something that has been wanted to be changed.

cecton:
> Tbh I don't mind either but this is something that has been wanted to be changed.

Ok let's do this for the greater good then :grin: I also added the deprecation warning

AethanFoot (approved):

VuiMuich:
Sorry I am a little late to the convo, but yesterday was not my day..

I get the reasoning to lean the config further, but personally I would prefer the config parameter over the additional command. 
Maybe an optional argument to the command would be ok.
My reasoning here is besides easier backwards compatibility (speaking of, mautamu has merged a PR to fix the `T` typo on his 0.3.0 branch) config maintainability.

A third approach I could see would be to move this config parameter (and possibly a bunch of others as well) from `leftwm-core` to `leftwm`?

cecton:
> Sorry I am a little late to the convo, but yesterday was not my day..
> 
> I get the reasoning to lean the config further, but personally I would prefer the config parameter over the additional command. Maybe an optional argument to the command would be ok. My reasoning here is besides easier backwards compatibility (speaking of, mautamu has merged a PR to fix the `T` typo on his 0.3.0 branch) config maintainability.
> 
> A third approach I could see would be to move this config parameter (and possibly a bunch of others as well) from `leftwm-core` to `leftwm`?

So to be clear:
1. This PR is **entirely** backward compatible. It doesn't break anybody's config.
2. Moving the config parameter from leftwm-core to leftwm is exactly what this PR is doing

If I understand you correctly, you would prefer leftwm's config to not deprecate the command GoToTag? So would you be happy if I revert the deprecation of GoToTag, remove GoToTagSwap and GoToTagNoSwap and keep the overall changes I did on leftwm-core untouched?

>  Maybe an optional argument to the command would be ok.
Thats makes the parsing harder as "value" is a string and we need to communicated the tag index + the swappable thingy.

cecton:
> speaking of, mautamu has merged a PR to fix the T typo on his 0.3.0 branch

Cool I didn't know! I don't touch it here as I keep compatibility but I changed it in core.

VuiMuich:
Appologies, for getting this wrong, I should have read the diff before commenting :see_no_evil:
Looks like I am still not at my game yet, don't know, where I lost my brains last week :man_shrugging:

I just had an idea, how this could become really smooth, but I need more time to write this down, will come back to this after lunch.

mautamu:
Honestly I prefer https://github.com/leftwm/leftwm/pull/617/commits/d308764dc116ee3b7070b9f40bbc81197a6fba0c to #292’s GoToTag method. Especially since it’s backwards-compatible I don’t see it as being a breaking change that would need to be in #292 either. 

Up to y’all though.

VuiMuich:
Oh crap, appears I rebooted before I sent what my idea was.
Just a brief summary before I have time to rewrite it:
- Add a command to set any parameter in state. So we would not only be able to dynamically disable the swap behavior but also could eventually lean out the command list by replacing commands like `leftwm command "SetMargiMultiplier 4"` by something like `leftwm command "SetParam MarginMuliplier 4"`
(On a side note this would also allow for even easier pywal theming or changing focus behavior on the fly)
- Keybind the concatenated commands for disable swap, gototag and enable swap if you just want it temporary. _To be fair: this is the exact opposite of being maintaining friendly, apologies for possibly being delusional on this topic

- **Bonus:** the ability to concat commands directly on keybind level, without leveraging `leftwm command` would be actually nice

cecton:
> * Keybind the concatenated commands for disable swap, gototag and enable swap if you just want it temporary. _To be fair: this is the exact opposite of being maintaining friendly, apologies for possibly being delusional on this topic

Yes not great :disappointed: I'm not for this but if others want...

> * Add a command to set any parameter in state. So we would not only be able to dynamically disable the swap behavior but also could eventually lean out the command list by replacing commands like `leftwm command "SetMargiMultiplier 4"` by something like `leftwm command "SetParam MarginMuliplier 4"`

I think it's a great idea! It's not necessary to implement this in this PR though.

> * **Bonus:** the ability to concat commands directly on keybind level, without leveraging `leftwm command` would be actually nice

One keybind triggers multiple commands? I think it's doable but also a bit out of scope.

Is there anything else we need for this PR in particular?

VuiMuich:
> Yes not great 😞 I'm not for this but if others want...

Eventually I am hoping for a beybind syntax that is somewhat similar to what sxhkd has and the ability to hav aliases in the config, then this would be even more smooth.
That said, I don't want to be the roadblock, if the desire for an additional command is that big.

> I think it's a great idea! It's not necessary to implement this in this PR though.
 
> One keybind triggers multiple commands? I think it's doable but also a bit out of scope.

Thats true, those topics are for other PRs for sure.
 
> Is there anything else we need for this PR in particular?

I think 'just' a decision, if we want to add `GoToTagNoSwap` (at least until we remove it again in a bigger breaking update).
Would it make things easier if this would be a custom command like `LoadTheme`? Tbh. I still don't fully get how those custom commands are actually working..

cecton:
> Would it make things easier if this would be a custom command like `LoadTheme`? Tbh. I still don't fully get how those custom commands are actually working..

So the thing is: the important change in this PR is to make the leftwm-core command GoToTag takes a parameter for swap. Then there is the question of what to do with the current leftwm (binary) config.

In the current state of this PR: compatibility is preserved, the GoToTag command has not changed. The config parameter is there. So this can be merged as is.

Now about the configuration file of leftwm (binary) itself, it's more complicated. I'm not personally interested in that but I do think this will go with the discussion about changing the configuration file format #496. I believe at some point we will make another `config` module (in leftwm (binary)) called `config-ng` (or something like that) that will be in RON (or some other format) and the discussion of the GoToTag command will happen there. Because that's where you discuss the end-user configuration file format.

But I don't think we should block the current PR as it is not breaking any compatibility. (Slight exception for the xdotool but that seems more like a fix (if it does anything at all)). This PR is only about leftwm-core itself and impacts the public API (which is a WIP and only used by me at the moment).

Issue #612, opened 2021-12-06
Feature request - Allow to rotate GridHorizontal


gitwipo:
Hello, 

is it possible to add the rotate tag command for the Layout `GridHorizontal`.

The main window is always on the left side. Therefore it is only useful for the left workspace in a dual workspace configuration. On the right workspace the main window is on the far right of the vision.

Thanks,

VuiMuich:
Sure, this should be relatively easy. I think we simply didn't see your use case, when we decided to spare this layout from this function.

TornaxO7:
Shouldn't we close this issue?

VuiMuich:
Yeah, thanks.

PR #620, opened 2021-12-12
Rotate grid horizontal

mike-cellini:
This pull request implements #612. 

I implemented only a horizontal flip, as that seemed to be the desired use case. Adding the vertical flip would be trivial, but would require adding 2 more rotation states (which may not be desired). If the vertical flip is desired, I can implement that before completing this pull request.

### DIFF
--- a/leftwm-core/src/layouts/grid_horizontal.rs
+++ b/leftwm-core/src/layouts/grid_horizontal.rs
@@ -1,3 +1,4 @@
+use crate::models::Tag;
 use crate::models::Window;
 use crate::models::Workspace;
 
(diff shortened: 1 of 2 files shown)


VuiMuich:
Thanks!

Just having a quick look at it on my phone right now and this looks good. 

Personally I think having all 4 rotation states would be more concise. A `Flip` command could be easily implemented in config by concating commands like `leftwm command "RotateTag" "RotateTag"`.

mike-cellini:
I updated this to support all four rotation states.

Notably, with the current rotation vec, I don't think you can't quite build a concise command that will toggle a flip on a single axis. Two `RotateTag` commands flips both axis.

We'd need to change from:
`[(false, false), (true, false), (true, true), (false, true)].to_vec()`
to
`[(false, false), (true, false), (false, true), (true, true)].to_vec()`
which may not quite match semantically.

AethanFoot:
Looks good. I suppose there isn't much point in over complicating this as there is a current rewrite of layout system.

PR #624, opened 2021-12-13
Version Rejig

AethanFoot:
As discussed on discord, bringing the versioning more inline with semantic versioning.


mautamu:
Will #594 and/or #617 be part of this bump?

AethanFoot:
This is less of a bump, and more just inlining the versioning. The bump will be on the v0.2.x branch, where I will add the dialog fixes patch to as well. Then any major patches that are backwards compatible can be added to both branches.

PR #630, opened 2021-12-19
[FEATURE] Set windows to spawn on predefined tags given the WM_CLASS

### LINKED ISSUE #427: [Question] Always spawn apps on a specific tag
Hi, 

I'm not sure if this has been brough up yet or not. Is it possible (if not, is it planned), to spawn an app always on the same tag? For example, say I want Firefox to always open on tag "web", thunderbird, slack etc. to open on tag "comms". 

A possible workaround for now could be what is mentioned here: https://github.com/leftwm/leftwm/issues/76#issuecomment-598462605

carrascomj:
Would close #427 

> This PR is a (working) proof of concept.

### Description

Implement _a way_ to define the tag where a window, defined by its `WM_CLASS`, should spawn. This is a common feature in other tiling window managers (like xmonad). The concrete way for now is to have a section in the config in the following form:

```toml
[window_config_by_class]
Navigator=2
krita=3
blender=3
"org.inkscape.inkscape"=3
soffice=4
```

In this example, windows whose `WM_CLASS` is, for instance, "krita" will spawn in the workspace number 3 (1-indexed).

### Implementation

Added a function to the trait `leftwm_core::Config`, and a corresponding implementation for `leftmw::Config`, with its corresponding serializable struct as a [`HashMap`](https://doc.rust-lang.org/std/collections/struct.HashMap.html).

This allows users of compiled configs to set their own implementation, maybe using a [`phf::Map`](https://docs.rs/phf/latest/phf/).

### Considerations

This PR is incomplete because it does not implement the desires set of features, mainly, that the user should be able to also specify windows that should float. It also may be reasonable to wait for https://github.com/leftwm/leftwm/pull/292.

A possible complete implementation would have a config like the one proposed at https://github.com/leftwm/leftwm/issues/427#issuecomment-997421359:

```toml
[[window_config_by_class]]
wm-class = "qutebrowser"
spawn_on_tag = "www" #tag.id as in the tag-configuration array, this could optionally be an integer with the tag.index
spawn_floating = true
```

This would mainly require changing the introduced `leftwm_core::Config` trait method [`get_tag_from_wm_class`](https://github.com/carrascomj/leftwm/blob/feat-phf-wmclass/leftwm-core/src/config/mod.rs#L64) to

```rust
/// Take a window and set the tag and is_floating, possibly from `WM_CLASS`. 
/// Returns true if the window was handled
fn setup_predefined_window(&self, window: &mut Window) -> bool;
```

(and changing the corresponding `leftwm::Config`).

**Or** introducing a new trait (such as `WindowMiddleware` or `WindowHook`) that a field in `leftmw_core::Manager` implements, if separating it from `Config` is desired:
```rust
trait WindowMiddleware {
    fn setup_predefined_window(&self, window: &mut Window) -> bool;
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -400,7 +400,7 @@ checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
 
 [[package]]
 name = "leftwm"
-version = "0.2.11"
+version = "0.3.0"
 dependencies = [
  "anyhow",
  "chrono",
@@ -434,13 +434,14 @@ dependencies = [
 
 [[package]]
 name = "leftwm-core"
-version = "0.2.11"
+version = "0.3.0"
 dependencies = [
  "dirs-next",
  "futures",
  "log",
  "mio 0.8.0",
  "nix",
+ "phf",
  "serde",
  "serde_json",
  "signal-hook",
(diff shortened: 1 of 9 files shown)


VuiMuich:
First of all thanks for taking on this.

Tbh. I am having a bit of a hard time figuring out the proper toml syntax to integrate the currnetly implemented hasmap in my config.
The proposed:
```toml
[window_config_by_class]
Navigator=2
krita=3
blender=3
"org.inkscape.inkscape"=3
soffice=4
```
only works, when appended at the end of the config.
But if another field follows the parser fails and `leftwm check` throws this error
```
Configuration failed. Reason: invalid type: sequence, expected usize for key `window_config_by_class.scratchpad` at line 16 column 1
```

carrascomj:
For me, the non-compact form works everywhere in the config. Now, I had some problems with the compact table form in TOML too. Reading [TOML's spec](https://toml.io/en/v1.0.0#inline-table):

>  "No newlines are allowed between the curly braces unless they are valid within a value [in inline tables]."

We can only blame **T**om for not being so **O**bvious in the **ML**, I guess.

VuiMuich:
Yeah, I guess "blame toml" is the thing here.

I my case I had
```toml
[window_config_by_class]
...

scratchpad = [
...
]
```
It appears from what I get in their specs, that in this syntax the previous table gets not terminated properly, as 'the end of a table is just before the beginning of a new table' (at least that's true for the root table).

carrascomj:
Oh, I now get it with the non-compact form, that's even worse. That's going to cause a lot of trouble, maybe I can try something on leftwm-check or we can wait for #631 to implement everything.

carrascomj:
Closing in favor of #631

Issue #582, opened 2021-11-22
Always float specified windows


patrikha:
HI,

Is there also a possibility to setup rules for always floating some windows, like you can do in [i3](https://i3wm.org/docs/userguide.html#for_window) ?

Of course I can add a keybinding for this but it is very convenient to automatically float some pop-up windows.

VuiMuich:
This is kind of a regular request. Specifically this is in #251.

We are still in need of a implementation to make configurable use of window classes.

In the mean time I can only serve you another workaround using scratchpads.

Issue #251, opened 2021-03-21
How do you spawn floating windows?


jkimmerling:
I know I am probably missing something really easy, but I cannot figure out how to spawn specific windows in floating mode.

I want to spawn a floating terminal window, but I cannot find out how to do this. I've googled, searched here, looked through all the example themes, etc. No luck lol.

mautamu:
There isn't, to the best of my knowledge, a way to spawn something as floating (I think this depends on #165). However to make an existing window floating, modkey + right/left click-drag will turn a window into a floating window.

jkimmerling:
Well darn, I was hoping to be able to have polybar spawn floating terminals for things like wttr.in weather reports and such.

mautamu:
If you can choose what hints Polybar puts on a window, Dialog, Splash (pending #184), and Floating all float. Toolbar might too?

jkimmerling:
Could you possibly give me the syntax on how to choose what hints something gets?

And it would be a very nice feature for ephemeral windows such as weather reports or memo programs :)

jkimmerling:
I cannot find a way to call the terminal in question (kitty) as any of those "hints"

I tried setting the class, title, and name to all three of the ones you mentioned, and nothing worked.

Is there a way to call a program from the terminal and set the tag? As the polybar spawns are just shell commands

AethanFoot:
If you are creative enough you may be able to manipulate rofi to display what you want, as is can be display anywhere on the screen.

jkimmerling:
Fair enough, I do have some examples to play with for that. It would just be nice to have a flag or something that could be added to a call to make it float. But I think that would be way down on the priority list, or even not wanted, since it seems they want to keep it as simple as possible. Though it would make sense to allow themes to easily push this sort of feature.

jkimmerling:
it seems Guake does not float either. It seems I will have to do something "creative" :(

hertg:
I'm closing this because it should be resolved with window rules in #631

PR #631, opened 2021-12-22
[FEATURE] Full window hooks

### LINKED ISSUE #427: [Question] Always spawn apps on a specific tag

carrascomj:
Follow up to #630 
Close #427 #582 #251

### Description
Let the user configure _window hooks_: specifying a window by its `WM_CLASS` and/or its title, define how the window should spawn, on a tag or its floating state. The config is as defined in https://github.com/leftwm/leftwm/issues/427#issuecomment-997421359:

```toml
[[window_rules]]
window_class = "qutebrowser"
# the user can also select by window title 
# window_title="qutebrowser - settings"
spawn_on_tag = 2
spawn_floating = true
```

### Implementation
#630 implemented the spawning on tag feature. This PR adds on top of that:
* `leftwm-core`: the config now defines `setup_predefined_window`, which takes a `&mut Window` and decides if/how to modify it (instead of simply returning a tag number, if any).
* `leftwm`: define `WindowHook` (deserialized from `config.toml`) and a **scoring system** that decides the best `WindowHook` match for a `Window`. This is because multiple hooks may match a `WM_CLASS` but we want the most specific one (title >= class). See use case at https://github.com/leftwm/leftwm/issues/427#issuecomment-998587861.
* Slight **refactor** to avoid circular dependencies: 
  ```rust 
  Window::load_config(Config)
  ```
  is now
  ```rust
  Config::load_window(Window)
  ```

**TODO**

* [ ] Is floating working as desired?
* [ ] Define tag also by name.
* [x] Decide on final config nomenclature.
* [x] Decide on `WM_NAME`, `_NET_WM_NAME`.


AethanFoot:
Is there a better name we can think of, I know its strange to say but it doesn't really roll of the tongue and doesn't feel like an obvious/natural name. Other than that, great work!

VuiMuich:
Yeah, the name was my suggestion in #427 but I agree its far from great.
Considering @hertg's comment in the same issue I think `window_rules` might be better fitting, what do you think?

AethanFoot:
Yeah I think that works well, this then will still fit when more settings are exposed (for example custom settings for a window type).

carrascomj:
Now that it is not only about the class, `window_rules` sounds great.

Also, following previous art, `window_rules` matches the common dwm path [switchtotag](https://dwm.suckless.org/patches/switchtotag/). Another option is `manage_hook`, [commonly used in xmonad](https://wiki.haskell.org/Xmonad/General_xmonad.hs_config_tips#ManageHook_examples), or `window_hook`.

```toml
[[window_rules]]
wm_class = "qutebrowser"
# alternatively, the user can select by window title 
# title="qutebrowser - settings"
spawn_on_tag = "www" #tag.id as in the tag-configuration array, this could optionally be an integer with the tag.index
spawn_floating = true
```

or in compact form:

```toml
window_rules = [ 
  {wm_class = "Navigator", spawn_on_tag = 2},
  {wm_class = "krita", spawn_on_tag = 3},
  {wm_class = "blender", spawn_on_tag = 3},
  {wm_class = "Pentablet_Driver", spawn_on_tag = 9, spawn_floating = true},
  {wm_class = "org.inkscape.inkscape", spawn_on_tag = 3},
  {wm_class = "soffice", spawn_on_tag = 4},
]
```

Also, _title_ and _wm_class_ are not consistent, maybe it would be better to use _wm_name_ or _net_wm_name_ instead of title.

AethanFoot:
What it be worth it to make net_wm_name (or maybe _net_wm_name)  and wm_name interchangeable to make it more clear for newer users as there are both and it is no guarantee whether one or the other or both are there. Just I worry if it will cause confusion if we say use wm_name and the window they check doesn't have it and has _net_wm_name and vice versa.

VuiMuich:
Has the `_net` any actual implications?
Maybe we could go with `window_class` and `window_title`.

carrascomj:
I am using `_NET_WM_TITLE` because that's what was already implemented for `Window`, but not super sure about the differences.

> Would it be worth it to make net_wm_name (or maybe _net_wm_name) and wm_name.

I don't oppose to that but it's hard to say what the actual behavior should be. Should the manager set both for every window and the config consumer decide? Should it just set it to whatever it finds between the two? What should it prioritize? The advantage of `_NET_WM_TITLE` is that it will always match whatever leftwm-state says.

> Maybe we could go with window_class and window_title.

I like it because is more clear. It has the drawback that it does not map to a particular attribute from the output of `xprops`.

One point to make here is that the title (or maybe in the future the type or the instance class) is in the window rules to fine-tune behaviors for very particular cases, mainly differentiate between a main window of a program and dialogs etc. which share the `WM_CLASS`. So I guess the usual workflow for that is to rely on `xprops` to configurate those cases. It will have to be properly documented in the wiki.

Let me know what you think.

AethanFoot:
We gather the window name from both. Reading a bit closer, it seems wm_name was part of an older spec whereas _net_wm_name is part of ewmh. So I suppose either would do, I personally lean towards _net_wm_name if we are to use one of them (as it is newer and more relevant), otherwise window_title or window_name should work if we clearly document it.

Just as an extra thought, we may want to consider how we would make such features display server agnostic (this isn't important now).

carrascomj:
In terms of the `leftwm_core::Config` trait part, it is already server agnostic, since it accepts a reference to `Window` (would Wayland require anything else?). It is up to the consumer `impl Config` to decide on what to do with the window after that (also server-independent). 

The server-dependent step is when the `Window` information gets filled, and that's already isolated in the implementation of [`From<XEvent>` for `Option<DisplayEvent>`](https://github.com/carrascomj/leftwm/blob/8875c2f5c69b2eaab4f157f398555ec27cc0f4b3/leftwm-core/src/display_servers/xlib_display_server/event_translate.rs#L81) (which is very well thought, by the way).

carrascomj:
I have been using the fork for a while and it looks ready.

I would leave the feature of specifying the tag by name for another PR, since it requires non-trivial deserialization.

leftwm/src/config/mod.rs
```diff
@@ -123,6 +123,53 @@ impl Keybind {
...
+/// wm_class = "krita"
+/// spawn_on_tag = 3
+/// spawn_floating = false
+/// ```
+///
+/// windows whose WM_CLASS is "krita" will spawn on tag 3 (1-indexed) and not floating.
+#[derive(Serialize, Deserialize, Default, Debug)]
+pub struct WindowHook {
+    /// `WM_CLASS` in X11
+    pub window_class: Option<String>,
```
mautamu, line 144: Hmm, would/should this be an enum between window class and name when we move away from TOML?

mautamu (approved):
Awesome! Thank you for this!

AethanFoot (approved):
Finally have been able to test this and is working as expected.

leftwm/src/config/mod.rs
AethanFoot: It would be nicer on the code side, but how would that work in the config side? One advantage of having them separate is it allows you to be more specific, e.g. you much a terminal by class but only the ones with a certain name.

carrascomj: It would work the same with the current scoring; the `WindowHook` matching the title would be selected over the one with the class.

Having an Enum

```rust
struct WindowMatch {
    Class(String),
    Title(String),
}

struct WindowHook {
    wm_match: Option<WindowMatch>,
    // ...
}
```

would be fine. A custom implementation for `Deserialize` would be needed as of now, though.

AethanFoot:
@carrascomj Would you be able to add these bits to the wiki. 
Many Thanks!

carrascomj:
Sure!

PR #636, opened 2021-12-24
Implement layouts per workspace

### LINKED ISSUE #618: Per monitor / workspace layout set
Currently, the list of layouts is defined globally. It would be nice to be able to configure layouts per monitor / workspace (or even tag?)

The main use case for this is when mixing monitor aspect ratios and orientation.

For example, I currently use a tilling WM on theses:

- A laptop in 1080p, where I mainly use `MainAndVertStack` and `Monocle`
- My main computer which has an ultra-wide  and a vertical screen on the side which both have completely different layouts requirements.

For the laptop, I could of course use a different config file, not ideal, but can be worked around.
However the dual screen is problematic, I have to add all the layouts I would use for both monitor, and then manually change the layout on start. This could probably be automated via `leftwm-command` with `FocusWorkspaceNext` folowwed by `SetLayout`, but that feels more like a hack, and is still painful when cycling layouts.

In my xmonad config, I use the following code that selects the layout set based on the current monitor width:
https://github.com/kid/xmonad/blob/11fae48f82680864f3ff9463310ead3bec3b13b1/xmonad.hs#L106-L113

Ideally I would like to be able to do the same, but that will probably need https://github.com/leftwm/leftwm/discussions/496 or simply by building my own binary using only `leftwm-core`

kid:
This is an attempt to resolve #618, this is still WIP, but thought I would open the PR for early feedbacks before proceeding. It's still untested and I'm not sure how it will work with `layout_mode`

### DIFF
--- a/leftwm-core/src/layouts/mod.rs
+++ b/leftwm-core/src/layouts/mod.rs
@@ -54,6 +54,12 @@ impl Default for Layout {
     }
 }
 
+impl Default for &Layout {
+    fn default() -> Self {
+        &Layout::MainAndVertStack
+    }
+}
+
 // This is tedious, but simple and effective.
 impl Layout {
     pub fn update_windows(&self, workspace: &Workspace, windows: &mut Vec<&mut Window>, tag: &Tag) {
--- a/leftwm-core/src/layouts/mod.rs
+++ b/leftwm-core/src/layouts/mod.rs
@@ -54,12 +54,6 @@ impl Default for Layout {
     }
 }
 
-impl Default for &Layout {
-    fn default() -> Self {
-        &Layout::MainAndVertStack
-    }
-}
-
 // This is tedious, but simple and effective.
 impl Layout {
     pub fn update_windows(&self, workspace: &Workspace, windows: &mut Vec<&mut Window>, tag: &Tag) {
--- a/leftwm-core/src/models/layout_manager.rs
+++ b/leftwm-core/src/models/layout_manager.rs
@@ -38,15 +38,15 @@ impl LayoutManager {
         }
     }
 
-    pub fn new_layout(&self) -> Layout {
-        *self.layouts.first().unwrap_or(&Layout::default())
+    pub fn new_layout(&self, workspace_id: Option<i32>) -> Layout {
+        *self
+            .layouts(workspace_id)
+            .first()
+            .unwrap_or(&Layout::default())
     }
 
     pub fn next_layout(&self, workspace: &Workspace) -> Layout {
-        let layouts = workspace
-            .id
-            .and_then(|id| self.layouts_per_workspaces.get(&id))
-            .unwrap_or(&self.layouts);
+        let layouts = self.layouts(workspace.id);
 
         let next = match layouts.iter().position(|&x| x == workspace.layout) {
             Some(index) if index == layouts.len() - 1 => layouts.first(),
@@ -58,10 +58,7 @@ impl LayoutManager {
     }
 
     pub fn previous_layout(&self, workspace: &Workspace) -> Layout {
-        let layouts = workspace
-            .id
-            .and_then(|id| self.layouts_per_workspaces.get(&id))
-            .unwrap_or(&self.layouts);
+        let layouts = self.layouts(workspace.id);
 
         let next = match layouts.iter().position(|&x| x == workspace.layout) {
             Some(index) if index == 0 => layouts.last(),
@@ -90,6 +87,12 @@ impl LayoutManager {
             }
         }
         Some(true)
+    }
+
+    fn layouts(&self, workspace_id: Option<i32>) -> &Vec<Layout> {
+        workspace_id
+            .and_then(|id| self.layouts_per_workspaces.get(&id))
+            .unwrap_or(&self.layouts)
     }
 }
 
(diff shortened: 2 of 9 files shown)


leftwm-core/src/models/layout_manager.rs
```diff
@@ -19,40 +19,53 @@ impl Default for LayoutMode {
...
+        *workspace
+            .id
+            .and_then(|id| self.layouts_per_workspaces.get(&id))
+            .or(Some(&self.layouts))
+            .and_then(|layouts| match layouts.iter().position(|&x| x == layout) {
+                Some(index) if index == layouts.len() - 1 => layouts.first(),
+                Some(index) => layouts.get(index + 1),
+                None => None,
+            })
+            .unwrap_or_default()
```
kid, line 55: Should we return the current layout here if we can't find the next one?

AethanFoot, line 55: This is quite hard to digest, something like:
```rs
match workspace.id {
    Some(id) if self.layouts_per_workspace.contains_key(&id) => *self.layouts_per_workspaces.get(&id),
    _ => match self.layouts.iter().position(|&l| l == layout) {
        Some(index) if index == layouts.len() - 1 => *layouts.first(),
        Some(index) => *layouts.get(index + 1),
        None => layout,
    }
}
```
Feels a bit easier to read.

And yes I would return the current layout if we can't find the next.

leftwm-core/src/models/layout_manager.rs
```diff
@@ -19,40 +19,53 @@ impl Default for LayoutMode {
...
+            .id
+            .and_then(|id| self.layouts_per_workspaces.get(&id))
+            .or(Some(&self.layouts))
+            .and_then(|layouts| match layouts.iter().position(|&x| x == layout) {
+                Some(index) if index == 0 => layouts.last(),
+                Some(index) => layouts.get(index - 1),
+                None => None,
+            })
+            .unwrap_or_default()
     }
```
AethanFoot, line 70: **Same here.

leftwm-core/src/models/layout_manager.rs
```diff
@@ -19,40 +19,53 @@ impl Default for LayoutMode {
...
-    pub fn next_layout(&self, layout: Layout) -> Layout {
-        let mut index = match self.layouts.iter().position(|&x| x == layout) {
-            Some(x) => x as isize,
-            None => return Layout::default(),
-        } + 1;
-        if index >= self.layouts.len() as isize {
-            index = 0;
-        }
-        self.layouts[index as usize]
+    pub fn next_layout(&self, layout: Layout, workspace: &Workspace) -> Layout {
```
AethanFoot, line 45: Should we just pass the id e.g.:
```rs
pub fn next_layout(...., id: Option<i32>)...
```

leftwm-core/src/layouts/mod.rs
```diff
@@ -54,6 +54,12 @@ impl Default for Layout {
     }
 }
 
+impl Default for &Layout {
+    fn default() -> Self {
+        &Layout::MainAndVertStack
+    }
+}
+
```
AethanFoot, line 62: This is kinda redundant. With returning the current layout this won't be needed anyway.

AethanFoot (commented):
Nice work, just had some niggly little things.

leftwm-core/src/models/layout_manager.rs
kid: we could also pass the workspace itself, since everywhere this is called, the layout comes from the workspace anyway

leftwm-core/src/models/layout_manager.rs
kid: This does  not quiet work as the second part of the match need to apply to either the layouts from the workspace, or the global one, 

Keeping the same style however, this work and is I think quiet readable:
```rust
        let layouts = workspace
            .id
            .and_then(|id| self.layouts_per_workspaces.get(&id))
            .unwrap_or(&self.layouts);

        let next = match layouts.iter().position(|&x| x == current) {
            Some(index) if index == layouts.len() - 1 => layouts.first(),
            Some(index) => layouts.get(index + 1),
            None => None,
        };

        *next.unwrap_or(&current)
```

I've renamed `layout` to `current` for clarity

kid:
I've been using this locally for the last 3 days, and this seems to work perfectly so far

There is however one edge case which is when the display configuration changes.
Basically if you want to define layouts per screen, you're forced to define workspaces that match your screen configuration, which isn't ideal.

In my case, I have a dual monitor setup, one primary ultra-wide, one vertical on the right, because of this, the primary monitor is at (0, 480), but if I disconnect the second monitor (e.g, to have g-sync working properly...), then the primary is now positioned at (0, 0), and that screw up the layout (in this case, the layout starts a (0, 480))

AethanFoot:
Ye we need to improve the adaption to screen changes. I should have time tomorrow to test this then can merge.

AethanFoot:
This looks like it is ready, only one thing I found that in a practical case the returning the current layout if we can't find a next/previous leads to the user being stuck on a layout if they soft reload with different layouts, maybe we should return the first in the list for that workspace.

kid:
Ha yes, that makes sense, fixed that.

I also fixed another edge case where the layout list is defined, but is empty, in that case we should fallback go the global layout list

AethanFoot:
In future what we could do is do per screen configuration. This would allow a screen to to have settings like the current workspace config, but it can be defined by its port or say its orientation (that would be options for all in that orientation). The screen then can be configured with multiple workspaces under if someone is on an ultrawide. This would help as we can automatically find the dimensions if not specified and allows options like this to be applied without the sizing and such being needed. This would also hopefully allow greater options for multi-monitor setups and we could add a separate utility that helps the user setup these options.

Issue #598, opened 2021-12-03
Version bump `0.2.11` wishlist


VuiMuich:
This issue is going to be used as a wishlist for issues to be fixed for the `0.2.11` release.

Also feel free to briefly suggest wishes or ideas not yet tracked by an issue and eventually open the issue later on.

di-effe:
A function to retrieve all key bindings like Awesome do would be nice

mautamu:
> A function to retrieve all key bindings like Awesome do would be nice

In leftwm-state?

di-effe:
> 

I guess that would work too, parsing the result with a liquid template and passing the result to something like Rofi.
But tbh I was thinking at something that works out of the box, displaying a popup/overlay on screen

VuiMuich:
> > 
> 
> 
> 
> I guess that would work too, parsing the json with a liquid template and passing the result to something like Rofi.
> 
> But tbh I was thinking at something that works out of the box, displaying a popup/overlay on screen

I think this would be a nice feature for `lefthk` actually, but also only printing the keybinds. Putting this in a nice window definitely should be a task for conky or eww, just to name two possibilities.
Adding this directly in leftwm/lefthk could easily be the starting point of bloating things.

di-effe:
I get your point and make sense.

di-effe:
I am not sure if this would be technically possible or it would make any sense or use, but (<- magic word) what about managing the 4 window borders (optionally) separately? In theory it could open new possibilities (and nightmares for the eyes) for theming.

For example having a bottom border thicker and/or with a different accent color.

blackdragon2447:
the ability to set keybinds in the `theme.toml`

VuiMuich:
> the ability to set keybinds in the `theme.toml

I see your desire, though I think this would be better kept as a 'strong recommendation to use these keybinds' in the README of the theme, as I don't want the theme to interfere with my workflow. I think this is exactly the beauty of leftwm's concept that the theme is widely agnostic of the users workflow.
Maybe if we introduce a `block_theme_overrides = true`  in the config...

VuiMuich:
> I am not sure if this would be technically possible or it would make any sense or use, but (<- magic word) what about managing the 4 window borders (optionally) separately? In theory it could open new possibilities (and nightmares for the eyes) for theming.
> 
> For example having a bottom border thicker and/or with a different accent color.

[...]

PR #638, opened 2021-12-27
Layout CenterMainFluid (leftwm-core)

mautamu:
Adds CenterMainFluid, a layout with right blank space when there are two windows, to address @muemmel's concerns in #598. 

Cheers,
Mau


hertg:
I have two questions about that new layout and the use-case it tries to solve. Mainly just for my understanding on how it should behave, because I am working on refactoring the whole layout code in a separate crate (https://github.com/leftwm/leftwm-layouts).

1. Does / Should this layout respect the `max_window_width` config?
2. Shouldn't the left/right column be left empty with just one window open, similar to what happens with two windows? Otherwise this is a bit inconsistent.

@muemmel
And the last question, mainly so I can understand the request: Did you try out different values in `max_window_width`? What value are you using, and what is the width of your monitor?

I initially introduced the `max_window_width` configuration because I struggled with my ultra-ultra-wide 49' monitor. I am currently using a `max_window_width` of `2200px` with a monitor of `5120px` width, I'm using the `CenterMain` layout most of the time where it behaves as follows, and I'm pretty happy with that:
```txt
+-----------------------------------------------+
|               +---------------+               |
|               |               |               |
|               |       1       |               |  [49' monitor]
|               |               |               |
|               +---------------+               |
+-----------------------------------------------+

+-----------------------------------------------+
|        +--------------+--------------+        |
|        |              |              |        |
|        |       1      |       2      |        |  [49' monitor]
|        |              |              |        |
|        +--------------+--------------+        |
+-----------------------------------------------+

+-----------------------------------------------+
|+-------------+----------------+--------------+|
||             |                |              ||
||      2      |        1       |       3      ||  [49' monitor]
||             |                |              ||
|+-------------+----------------+--------------+|
+-----------------------------------------------+
```

Is your suggestion for the new layout to behave as follows?
```txt
+-----------------------------------------------+
|               +---------------+               |
|               |               |               |
|               |       1       |               |  [49' monitor]
|               |               |               |
|               +---------------+               |
+-----------------------------------------------+

+-----------------------------------------------+
|+--------------+---------------+               |
||              |               |               |
||      2       |       1       |               |  [49' monitor]
||              |               |               |
|+--------------+---------------+               |
+-----------------------------------------------+

+-----------------------------------------------+
|+--------------+---------------+--------------+|
||              |               |              ||
||      2       |       1       |       3      ||  [49' monitor]
||              |               |              ||
|+--------------+---------------+--------------+|
+-----------------------------------------------+
```

muemmel:
My horizontal resolution is 5120px. The max window width option definitely helps. I have it set to 60% for my workflow though. I like to have my 'main' working window (e.g. Rider) in the center, while using the side windows for utilities like browser and terminal or even Rider tool windows. I don't want them to take up that much space, since they rarely need it.

Your second diagram is exactly how I'd like the layout to behave,. Granted, it's mostly for convenience and a less jarring experience when opening or closing a third window.

mautamu:
Adjusted -fluid to match the second diagram for cases 1-3.

VuiMuich:
As mentioned in #640 it might be worth to either wait for @hertg's layout rework to get merged or to PR directly to the branch of his.

Otherwise, nice work!

hertg:
@muemmel 
Would you also prefer this behavior for other layouts than `CenterMain` or is this the exception? For example the `MainAndDeck` where the stack should be reserved instead of the main window being fullscreen?

Again, just asking so I can assess the implications of this request as I'm trying to generalize the layout behaviors in the refactored layout code.

muemmel:
First of all, thanks for the very quick implementation. I really appreciate it. 32:9 support is very limited on most WMs and this will definitely make my life easier and my work more enjoyable. While I had been playing around with virtual screens, splitting my physical monitor that way always feels somewhat hacky and unsatisfactory.

@hertg 
Having it as an option for other layouts would be ideal. Though I'm pretty sure MainAndDeck already behaves that way; i.e. Monocle-like with one window, respecting the max. width. However, I really don't want to inconvenience you or anyone else. My wants are very particular, regardless of my setup.

hertg:
> First of all, thanks for the very quick implementation. I really appreciate it. 32:9 support is very limited on most WMs and this will definitely make my life easier and my work more enjoyable. While I had been playing around with virtual screens, splitting my physical monitor that way always feels somewhat hacky and unsatisfactory.

I feel that, with LeftWM you could also manually create workspaces to split your monitor into multiple "screens". I never tried it out though, because I want my configuration to be applicable to changing environments when working on my laptop (docked and undocked).

> Having it as an option for other layouts would be ideal. Though I'm pretty sure MainAndDeck already behaves that way; i.e. Monocle-like with one window, respecting the max. width.

The `max_window_width` should be respected in most layouts, but the space of an absent stack is never reserved as you requested. 

> However, I really don't want to inconvenience you or anyone else. My wants are very particular, regardless of my setup.

Well, you are slightly inconveniencing me, but in a good way. It's better for me to know of the edge-cases and potential future feature request while reworking the layout code, so I can consider them and add the possibility to actually implement them.

One thought I had in mind, which is why I asked the question about other layouts than `CenterMain`, is that there could be a setting (LayoutModifier) to specify which behavior the user wants. Meaning, should the available screen-estate always be filled while respecting the `max_window_width` (as it does now), or should it reserve the space for columns that doesn't have any windows (yet). It helps thinking about the layouts in columns, because with my rework I want to allow having multiple *main windows*. The area where main-windows are placed is a single column. Each column can then be further split into multiple rectangles (vertically, horizontally, or in a grid).

**1 column**
- Monocle
- Grid
```
+---------+
|    S    |
+---------+
-> There is no main column
```

**2 column**
- Main and Deck
- Main and Stack
- Fibonacci
```
+----+----+
| M  | S  |
+----+----+
```

**3 column**
- Center Main
```
+--+---+--+
|S1| M |S2|
+--+---+--+
```

VuiMuich:
@mautamu may I ask for a resolve of the merge conflicts please, so we could finally merge this after dallying on this for some misconception about the work on the layout system.

mautamu:
Finally ready to merge, thanks @muemmel, @VuiMuich. Probably wise to wait until after #594

PR #639, opened 2021-12-28
feat: provide a Nix Flake for NixOS

kid:
This PR adds a flake.nix for people using NixOS. It makes it easy to install the latest version of leftwm, as well as running custom branches.

It also provides a shell with the minimal tooling required to work on leftwm. This can be called through `nix-direnv` or directly with `nix develop`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,3 +61,14 @@ jobs:
 
     - name: Display MSRV
       run: cargo -V
+
+  nix-flake:
+    name: nix-flake
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


AethanFoot:
Should we create a separate folder to contain these build files, e.g. this, the MakeFile, a PKGBUILD, etc. Then include a README.md detailing the build instructions.

kid:
`PKGBUILD` sounds like something that could live in a `contrib` folder indeed, but both `Makefile` and `nix` expect their configuration files to be  at the root by default, they would become painful to use as you would always need to either `cd` into the directory or pass extra cli flags

PR #640, opened 2021-12-28
Add LeftMain layout

kid:
This is pretty much the same as `CenterMain` except that the main column is on the left side instead of the center.


VuiMuich:
Hey, thanks for your contribution!
As @hertg currently is undertaking the layout system, I'd like to delay this PR.
Alternatively you could PR directly to his branch instead.

Issue #416, opened 2021-07-07
Click on empty screen and then spawn windows there


AeroNotix:
I am trying to click on an otherwise empty screen, then open a window there.

It seems to open the window on the screen I was previously focused on.

Is there some configuration I am missing, or is this not possible. Workaround I am using right now is to spawn the window I want on any screen which lets me, then push the window around until its on the correct screen.

VuiMuich:
Currently we don't have this behaviour, as we don't move workspace focus by mouse.
The only ways the workspace focus changes is by focusing a window on a workspace or by the [commands](https://github.com/leftwm/leftwm/wiki/Config#focusworkspacenext) `FocusWorkspaceNext`/`FocusWorkspacePrevious`.

AeroNotix:
Thanks for the information @VuiMuich

VuiMuich:
You are very welcome.

PR #642, opened 2022-01-02
fix(641): focus the last window in the workspace

### LINKED ISSUE #641: FocusWorkspace(Next|Previous) always reset the focus to the first window
**Describe the bug**

When calling FocusWorkspace(Next|Previous) the focus always set reset to the first window on the workspace

**To Reproduce**

With at least 2 workspace defined
1. Create 2 windows on the first workspace
2. Focus the second window
3. Call `FocusWorkspaceNext
4. Call `FocusWorkspacePrevious

**Expected behavior**

The second window should have the focus

**Actual behavior**

The focus get reset to the first window

**Environment:**
 - OS/Distro: NixOS
 - Output of `leftwm-check`
```
❯ leftwm-check             
:: LeftWM version: 0.3.0
:: LeftWM git hash: fd42fe8bff163aa9b383e946f2062f33ab5500be
    -> Configuration loaded OK 
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK 
:: Checking theme . . .
    -> Theme OK 
```

**Additional context**

Tested with `focus_behavior` `Driven` and `ClickTo`

kid:
Ensure `focus_workspace` saves the last window on the tag before
switching, and then restores it for the next workspace

Resolves #641

### DIFF
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -596,17 +596,12 @@ fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
 fn focus_workspace_change(state: &mut State, val: i32) -> Option<bool> {
     let current = state.focus_manager.workspace(&state.workspaces)?;
     let workspace = helpers::relative_find(&state.workspaces, |w| w == current, val, true)?.clone();
-    state.focus_workspace(&workspace);
+    let result = state.focus_workspace(&workspace);
     if state.focus_manager.behaviour == FocusBehaviour::Sloppy {
         let act = DisplayAction::MoveMouseOverPoint(workspace.xyhw.center());
         state.actions.push_back(act);
     }
-    let window = state
-        .windows
-        .iter()
-        .find(|w| workspace.is_displaying(w) && w.r#type == WindowType::Normal)?
-        .clone();
-    Some(handle_focus(state, window.handle))
+    Some(result)
 }
 
 fn rotate_tag(state: &mut State) -> Option<bool> {
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -585,10 +585,7 @@ fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
         (Some(next), Some(cur), Some(prev)) if next == cur && toggle => {
             Some(handle_focus(state, prev))
         }
-        (Some(next), Some(cur), _) if next != cur => {
-            state.focus_manager.tags_last_window.insert(tag, cur);
-            Some(handle_focus(state, next))
-        }
+        (Some(next), Some(cur), _) if next != cur => Some(handle_focus(state, next)),
         _ => None,
     }
 }
@@ -596,6 +593,7 @@ fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
 fn focus_workspace_change(state: &mut State, val: i32) -> Option<bool> {
     let current = state.focus_manager.workspace(&state.workspaces)?;
     let workspace = helpers::relative_find(&state.workspaces, |w| w == current, val, true)?.clone();
+
     if state.focus_manager.behaviour == FocusBehaviour::Sloppy {
         let action = workspace
             .tags
@@ -606,10 +604,9 @@ fn focus_workspace_change(state: &mut State, val: i32) -> Option<bool> {
                 |h| DisplayAction::MoveMouseOver(*h),
             );
         state.actions.push_back(action);
-        None
-    } else {
-        Some(state.focus_workspace(&workspace))
     }
+
+    Some(state.focus_workspace(&workspace))
 }
 
 fn rotate_tag(state: &mut State) -> Option<bool> {
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -601,8 +601,10 @@ fn focus_workspace_change(state: &mut State, val: i32) -> Option<bool> {
             .tags
             .first()
             .and_then(|tag| state.focus_manager.tags_last_window.get(tag))
-            .map(|handle| DisplayAction::MoveMouseOver(handle.clone()))
-            .unwrap_or_else(|| DisplayAction::MoveMouseOverPoint(workspace.xyhw.center()));
+            .map_or_else(
+                || DisplayAction::MoveMouseOverPoint(workspace.xyhw.center()),
+                |h| DisplayAction::MoveMouseOver(*h),
+            );
         state.actions.push_back(action);
         None
     } else {
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -596,12 +596,18 @@ fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
 fn focus_workspace_change(state: &mut State, val: i32) -> Option<bool> {
     let current = state.focus_manager.workspace(&state.workspaces)?;
     let workspace = helpers::relative_find(&state.workspaces, |w| w == current, val, true)?.clone();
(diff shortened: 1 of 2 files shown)


leftwm-core/src/handlers/focus_handler.rs
```diff
@@ -152,12 +158,22 @@ impl State {
...
+
+    // Add the currently focused window to the history
+    if let Some((tag, handle)) = state
+        .focus_manager
+        .tag(0)
+        .zip(state.focus_manager.window(&state.windows).map(|w| w.handle))
+    {
+        state.focus_manager.tags_last_window.insert(tag, handle);
+    }
+
```
AethanFoot, line 176: I think this being in focus_tag_work would make more sense.

AethanFoot (commented):
Nice work!

leftwm-core/src/handlers/focus_handler.rs
```diff
@@ -10,12 +10,18 @@ impl State {
     //NOTE: should only be called externally from this file
```
VuiMuich, line 10: Nitpick, but when you add the space in line 13, please add it also here :grin: 
```diff
- //NOTE: should only be called externally from this file
+ // NOTE: should only be called externally from this file
```

leftwm-core/src/handlers/focus_handler.rs
```diff
@@ -229,6 +236,16 @@ fn focus_tag_work(state: &mut State, tag: TagId) -> Option<()> {
...
+    // Add the currently focused window to the history
+    if let Some((tag, handle)) = state
+        .focus_manager
+        .tag(0)
+        .zip(state.focus_manager.window(&state.windows).map(|w| w.handle))
+    {
+        state.focus_manager.tags_last_window.insert(tag, handle);
+    }
+
     //clean old ones
```
VuiMuich, line 249: also here:
```diff
- //clean old ones
- state.focus_manager.tag_history.truncate(10);
- //dd this focus to the history
+ // clean old ones
+ state.focus_manager.tag_history.truncate(10);
+ // dd this focus to the history

File: leftwm-core/src/handlers/command_handler.rs [hunk 4]
```diff
@@ -598,17 +598,12 @@ fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
 fn focus_workspace_change(state: &mut State, val: i32) -> Option<bool> {
     let current = state.focus_manager.workspace(&state.workspaces)?;
     let workspace = helpers::relative_find(&state.workspaces, |w| w == current, val, true)?.clone();
-    state.focus_workspace(&workspace);
+    let result = state.focus_workspace(&workspace);
```
VuiMuich, line 601: I am wondering, if we really need to have this before the following `if`. Like this it feels a bit weird to define the `result` before checking for diverging behaviour.
But if we need to keep this order, maybe a little commentary or a different name might help in the future.

VuiMuich (commented):
This looks great.
I just have a few minor nitpick improvements while passing by.

leftwm-core/src/handlers/command_handler.rs
kid: Actually, do we even need to call `focus_workspace` if the focus behavior is sloppy? shouldn't moving the cursor trigger the the call

leftwm-core/src/handlers/focus_handler.rs
kid: humm, something's broken if I do that, `FocusWorkspaceNext` works, but `FocusWorkspacePrevious` somehow focus the top window on the current workspace.

leftwm-core/src/handlers/command_handler.rs
VuiMuich: I guess this was once introduced as a bug-fix I guess, but this means we need the documentation, why it's there even more.
@AethanFoot may we have some insight from you, as I guess you might have touched this line with #356

leftwm-core/src/handlers/focus_handler.rs
kid: hum, this isn't actually related to this, it's because the cursor got moved to the center of the screen, which was only showing the margin between my 2 windows, so the focus got borked until I moved the cursor.

leftwm-core/src/handlers/command_handler.rs
VuiMuich: Ah, I think I get it now: this moves the cursor to the center of the focused window in `Sloppy` in order to not trigger a refocus if mouse is moved after the workspace focus changed was called by command and not by mouse movement.

kid: Yes, but I think it should move it to the center of the last focused window instead, otherwise it can be moved to a position that does not have any window, e.g. if you have 2 windows in `EvenHorizontal` or `EvenVertical`, the cursor get positioned in the margin between 2 windows, cf my [comment](https://github.com/leftwm/leftwm/pull/642#discussion_r777214932) on the other thread

AethanFoot: I think that action is there in case there is no window on the workspace. So as the window focusing has moved the focus_workspace can be after, you could also check if there is a window on the workspace to prevent the extra action.

AethanFoot: We should probably add the handle_focus part within the focus_window function which would solve these issues and reduce the code in command_handler.

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -598,17 +598,12 @@ fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
...
     if state.focus_manager.behaviour == FocusBehaviour::Sloppy {
         let act = DisplayAction::MoveMouseOverPoint(workspace.xyhw.center());
         state.actions.push_back(act);
     }
```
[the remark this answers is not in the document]
kid: Didn't have much time lately, and probably won't until end of next week, but I can quickly revert this.

However, I don't think it make sense to merge this without at least moving the cursor to the center of the focused window (with a fallback to the center of the workspace if none is found)

kid:
Ok, finally got this one to work, there was a few edge cases because `tags_last_window` was not always updated.

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -585,28 +585,28 @@ fn focus_window_top(state: &mut State, toggle: bool) -> Option<bool> {
         (Some(next), Some(cur), Some(prev)) if next == cur && toggle => {
             Some(handle_focus(state, prev))
         }
-        (Some(next), Some(cur), _) if next != cur => {
-            state.focus_manager.tags_last_window.insert(tag, cur);
-            Some(handle_focus(state, next))
-        }
+        (Some(next), Some(cur), _) if next != cur => Some(handle_focus(state, next)),
```
mautamu, line 588: Awesome, much simpler

leftwm-core/src/handlers/focus_handler.rs
```diff
@@ -256,6 +294,49 @@ mod tests {
...
+
+        let actual = manager
+            .state
+            .focus_manager
+            .window(&manager.state.windows)
+            .map(|w| w.handle);
+
+        assert_eq!(expected, actual);
+    }
+
```
mautamu, line 339: Thanks for adding a test to this!

mautamu (approved):

VuiMuich:
Looking very good!

---
@kid may I take the opportunity to "assign" you a fix for #416 as this seems to be distantly related. Of course this is totally optional and it's up to your liking if you include it in this present PR or open a new one.

kid:
Sure I can have a look, but I probably won't have time until a couple of weeks though, so best to merge this one first

VuiMuich:
Sure, just thought I might ask you as you now familiar with a corner of the code base, that might be overlapping.
And I'd like to start getting boxes ticked for `3.0.0`, but we are in no hurry at all.

Thanks!

PR #643, opened 2022-01-02
fix: include display id in command pipe name

kid:
This change allow running multiple session of leftwm. Without this, the
second session would destroy the file of the first one, leaving it
unsuable.


mautamu (approved):
Lgtm

Issue #644, opened 2022-01-03
XDG autostart not working properly


nblxxx:
Hi there fellow adventurers!
It seems line 55 should come before 48 here, as the env var is queried inside the call to autostart.
This currently causes autostart to launch everything and to even ignore the "Hidden=true" in desktop entries for some reason.
https://github.com/leftwm/leftwm/blob/acf2407bc48286ff5388a3d301b5adef59150d52/leftwm/src/bin/leftwm.rs#L47-L55

VuiMuich:
Welcome adventurer!
Good catch.
Would you have a chance to checkout #646? I'm currently blocked from exiting my session for at least the next 24h..

nblxxx:
Sure! I just did, checked my desktop entries against running procs and the rules seem to apply properly this time.

PR #646, opened 2022-01-04
fix 644

VuiMuich:
Just a small patch:
- fix order of setting XDG Autostart env, solves #644
- solve a few minor clippy pedantic lints

---
There is currently one `clippy pedantic` lint left, about missing `error doc` of `impl keybind` ind `leftwm/src/config/mod.rs` but I guess this takes a little more time to be done properly.


mautamu:
Blocks #638 until later commits in 638 are reverted.

VuiMuich:
Oops, totally missed that those were already there. I guess those are easier reverted here, then in #638

PR #651, opened 2022-01-09
Fix command pipe stuff

AethanFoot:


leftwm/src/bin/leftwm-command.rs
```diff
@@ -49,12 +49,15 @@ async fn main() -> Result<()> {
...
-        
+
         UnloadTheme
         SoftReload
         ToggleFullScreen
+        ToggleSticky
         SwapScreens
         MoveWindowToLastWorkspace
+        MoveWindowToNextWorkspace
+        MoveWindowToPreviousWorkspace
```
VuiMuich, line 60: Thanks for adding these, I was thinking recently that I somehow miss some commands, but then thought we might just have disuccesd them but not implemented yet...

VuiMuich (approved):

PR #656, opened 2022-01-10
Various fixes

### LINKED ISSUE #637: Can't bring a window with wmctrl.
**Describe the bug**
The command `wmctrl -i -R <WIN>` does nothing

**To Reproduce**
Get any window id with the command `wmctrl -l`
Type the command `wmctrl -i -R <WIN> `.
The command `wmctrl -i -a <WIN>` is working as intended to  switch to the desktop containing the window `<WIN>`, raise the window, and give it focus,  so I know the problem is not the window id.

**Expected behavior**
As stated in the wmctrl manpage:
Move the window `<WIN>` to the current desktop, raise the window, and give it focus. 
Which should work in any  EWMH compliant window manager

**Environment:**
 - OS/Distro: Arcolinux
 - Output of `leftwm-check`:
```
:: LeftWM version: 0.3.0
:: LeftWM git hash: 55bae63-modified
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK
```

### LINKED ISSUE #650: Dialog boxes follows Tag Swap
**Describe the bug**
Every dialog box follows every Tag instead of staying in the Tag were the application is.
An example is with editing metadata in Calibre (added xprop at the end).

**To Reproduce**
1. Open Calibre
2. Click on 'Edit metadata' to open the Edit dialog
3. Swap Tag
4. The dialog follows

**Expected behavior**
The dialog should stay in the Tag were Calibre is open.

**Environment:**
 - OS/Distro: Arch Linux - leftwm 1:0.2.11-1
 - Output of `leftwm-check`
```
:: LeftWM version: 0.2.11
:: LeftWM git hash: 399345b
    -> Configuration loaded OK 
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK 
:: Checking theme . . .
    -> Theme OK
``` 

```
_NET_WM_DESKTOP(CARDINAL) = 1
WM_STATE(WM_STATE):
		window state: Normal
		icon window: <field not available>
_NET_WM_USER_TIME(CARDINAL) = 14822979
_NET_WM_STATE(ATOM) = _NET_WM_STATE_MODAL
WM_TRANSIENT_FOR(WINDOW): window id # 0x4400017
_NET_WM_ICON_NAME(UTF8_STRING) = 
_NET_WM_ICON(CARDINAL) = 	Icon (128 x 128):
	(not shown)


XdndAware(ATOM) = BITMAP
WM_NAME(STRING) = "Edit metadata - xxxxxxx -  [1 of xxx]"
_NET_WM_NAME(UTF8_STRING) = "Edit metadata - xxxxxxx -  [1 of xxx]"
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x26, 0x1e, 0x0, 0x0
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG, _NET_WM_WINDOW_TYPE_NORMAL
_XEMBED_INFO(_XEMBED_INFO) = 0x0, 0x1
WM_CLIENT_LEADER(WINDOW): window id # 0x440000e
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: True
		window id # of group leader: 0x440000e
WM_CLIENT_MACHINE(STRING) = "tardis"
_NET_WM_PID(CARDINAL) = 165432
_NET_WM_SYNC_REQUEST_COUNTER(CARDINAL) = 71303478
WM_CLASS(STRING) = "calibre-gui", "calibre"
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, WM_TAKE_FOCUS, _NET_WM_PING, _NET_WM_SYNC_REQUEST, _NET_WM_CONTEXT_HELP
WM_NORMAL_HINTS(WM_SIZE_HINTS):
		user specified location: 33, 90
		user specified size: 1852 by 928
		program specified minimum size: 376 by 154
		window gravity: Static
```

AethanFoot:
Fixes #650, fixes #637, restore window tag on reload, and clippy bits.
Thanks!

### DIFF
--- a/leftwm-core/src/command.rs
+++ b/leftwm-core/src/command.rs
@@ -1,4 +1,7 @@
-use crate::{layouts::Layout, models::TagId};
+use crate::{
+    layouts::Layout,
+    models::{TagId, WindowHandle},
+};
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
@@ -11,7 +14,10 @@ pub enum Command {
     ToggleScratchPad(String),
     ToggleFullScreen,
     ToggleSticky,
-    GoToTag { tag: TagId, swap: bool },
+    GoToTag {
+        tag: TagId,
+        swap: bool,
+    },
     FloatingToTile,
     TileToFloating,
     ToggleFloating,
@@ -25,7 +31,10 @@ pub enum Command {
     FocusWindowTop(bool),
     FocusWorkspaceNext,
     FocusWorkspacePrevious,
-    SendWindowToTag(TagId),
+    SendWindowToTag {
+        window: Option<WindowHandle>,
+        tag: TagId,
+    },
     MoveWindowToLastWorkspace,
     MoveWindowToNextWorkspace,
     MoveWindowToPreviousWorkspace,
--- a/leftwm-core/src/state.rs
+++ b/leftwm-core/src/state.rs
@@ -183,6 +183,7 @@ impl State {
                 ordered.push(new_window.clone());
                 self.windows.remove(index);
 
+                // Make the x server aware of any tag changes for the window.
                 let act = DisplayAction::SetWindowTags(new_window.handle, new_window.tags.clone());
                 self.actions.push_back(act);
             }
--- a/leftwm-core/src/state.rs
+++ b/leftwm-core/src/state.rs
@@ -182,6 +182,9 @@ impl State {
                 new_window.set_states(old_window.states());
                 ordered.push(new_window.clone());
                 self.windows.remove(index);
+
+                let act = DisplayAction::SetWindowTags(new_window.handle, new_window.tags.clone());
+                self.actions.push_back(act);
             }
         });
         if had_strut {
--- a/leftwm-core/src/utils/command_pipe.rs
+++ b/leftwm-core/src/utils/command_pipe.rs
@@ -58,7 +58,7 @@ impl CommandPipe {
     pub fn pipe_name() -> PathBuf {
         let display = env::var("DISPLAY")
             .ok()
-            .and_then(|d| d.rsplit_once(":").map(|(_, r)| r.to_owned()))
+            .and_then(|d| d.rsplit_once(':').map(|(_, r)| r.to_owned()))
             .unwrap_or_else(|| "0".to_string());
 
         PathBuf::from(format!("command-{}.pipe", display))
@@ -134,7 +134,10 @@ fn build_send_window_to_tag(raw: &str) -> Result<Command, Box<dyn std::error::Er
     let headless = without_head(raw, "SendWindowToTag ");
     let parts: Vec<&str> = headless.split(' ').collect();
     let tag_id: TagId = parts.get(0).ok_or("missing argument tag_id")?.parse()?;
-    Ok(Command::SendWindowToTag(tag_id))
+    Ok(Command::SendWindowToTag {
+        window: None,
+        tag: tag_id,
+    })
 }
 
 fn build_send_workspace_to_tag(raw: &str) -> Result<Command, Box<dyn std::error::Error>> {
--- a/leftwm-core/src/utils/state_socket.rs
+++ b/leftwm-core/src/utils/state_socket.rs
@@ -23,7 +23,7 @@ pub struct StateSocket {
 impl Drop for StateSocket {
     fn drop(&mut self) {
         assert!(
-            std::thread::panicking() || !self.listener.is_some(),
+            std::thread::panicking() || self.listener.is_none(),
             "StateSocket has to be shutdown explicitly before drop"
         );
     }
--- a/leftwm-core/src/display_event.rs
+++ b/leftwm-core/src/display_event.rs
@@ -12,7 +12,7 @@ pub enum DisplayEvent {
     WindowCreate(Window, i32, i32),
     WindowChange(WindowChange),
     WindowDestroy(WindowHandle),
-    MouseEnteredWindow(WindowHandle),
+    WindowTakeFocus(WindowHandle),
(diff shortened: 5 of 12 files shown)


leftwm-core/src/state.rs
```diff
@@ -182,6 +182,9 @@ impl State {
                 new_window.set_states(old_window.states());
                 ordered.push(new_window.clone());
                 self.windows.remove(index);
+
+                let act = DisplayAction::SetWindowTags(new_window.handle, new_window.tags.clone());
+                self.actions.push_back(act);
```
VuiMuich, line 187: I think here would be some commentary helpful, why we want to `push_back` here exactly.

leftwm-core/src/utils/command_pipe.rs
```diff
@@ -134,7 +134,10 @@ fn build_send_window_to_tag(raw: &str) -> Result<Command, Box<dyn std::error::Er
     let headless = without_head(raw, "SendWindowToTag ");
     let parts: Vec<&str> = headless.split(' ').collect();
     let tag_id: TagId = parts.get(0).ok_or("missing argument tag_id")?.parse()?;
-    Ok(Command::SendWindowToTag(tag_id))
+    Ok(Command::SendWindowToTag {
+        window: None,
+        tag: tag_id,
```
VuiMuich, line 139: I guess this should get some error-handling similar to `MoveToTag`, but for both `window` and `tag` as well, right?

leftwm-core/src/utils/state_socket.rs
```diff
@@ -23,7 +23,7 @@ pub struct StateSocket {
 impl Drop for StateSocket {
     fn drop(&mut self) {
         assert!(
-            std::thread::panicking() || !self.listener.is_some(),
+            std::thread::panicking() || self.listener.is_none(),
```
VuiMuich, line 26: Just out of curiosity, did clippy complain about this, or did you just notice this by your self?

leftwm-core/src/utils/command_pipe.rs
```diff
@@ -58,7 +58,7 @@ impl CommandPipe {
     pub fn pipe_name() -> PathBuf {
         let display = env::var("DISPLAY")
             .ok()
-            .and_then(|d| d.rsplit_once(":").map(|(_, r)| r.to_owned()))
+            .and_then(|d| d.rsplit_once(':').map(|(_, r)| r.to_owned()))
```
VuiMuich, line 61: Is it of any meaning to use single quotes here? Just asking for educational purpose..

VuiMuich (commented):
Nothing major here, just a few personal curiosities and some minore improvements on the doc side

BTW: at some point I guess it would be nice, if `leftwm command` would become a bit more chatty with errors, as it feels odd, that it behaves as if everything went well, even if the command internally failed. Of course it gets potentially logged if logging is on, but a STDOUT error and some Exit-code might be helpful eventually.

VuiMuich:
Regarding the fix for #637: I was trying if we could solve the request on the discord from earlier today and tried to pull up a window from the `Deck` in `MainANdDeck` or `Monocle`, but that causes weird issues with focus. It's kind of hard to explain, but I guess the focus moves to the hidden window but doesn't unhide the window. Therefore the focus can't move away from the window, the cycling of the deck works somewhat correctly though.
The actual issue of #637 looks fixed to me.

leftwm-core/src/command.rs
```diff
@@ -11,7 +14,10 @@ pub enum Command {
     ToggleScratchPad(String),
     ToggleFullScreen,
     ToggleSticky,
-    GoToTag { tag: TagId, swap: bool },
+    GoToTag {
+        tag: TagId,
+        swap: bool,
```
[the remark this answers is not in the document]
AethanFoot: An Option<bool> would be redundant as both are "binary switches" (couldn't think of another way to say), e.g. saying None add nothing as that would be false.

leftwm-core/src/command.rs
```diff
@@ -25,7 +31,10 @@ pub enum Command {
     FocusWindowTop(bool),
     FocusWorkspaceNext,
     FocusWorkspacePrevious,
-    SendWindowToTag(TagId),
+    SendWindowToTag {
+        window: Option<WindowHandle>,
+        tag: TagId,
+    },
```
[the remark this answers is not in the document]
AethanFoot: This isn't exposed to the user yet. When we do it might be worth it having the tag be an option as well, e.g. if no tag is entered we bring the window to the focused tag.

leftwm-core/src/utils/command_pipe.rs
AethanFoot: Was actually a clippy, but the reason is single quotes represents a char whereas double represents a string. Idk hope the backend works exactly but I would presume in the compiler the string would be optimised into a char, as type char has a smaller sized memory slot.

leftwm-core/src/utils/state_socket.rs
AethanFoot: Ye a new clippy

VuiMuich: Ah, being new makes sense, as it didn't complain about this, when I pushed #646

AethanFoot:
That may require a bit of restructuring which would be beneficial. We would need to move more focus logic into the handler so that focus_window handles the logic in command_handler's focus changes, so these are run whenever focus_window is called. Basically focus_window calls that logic instead of the logic calling focus_window, e.g. then when focus_window is called it will move the window up in the stack when in monocle. Hope that makes sense.

PR #663, opened 2022-01-13
Fix telegram and unmap windows to hide them

### LINKED ISSUE #591: [bug] hidden window's off-screen 'parking position' souldn't be affected by `MarginMultiplier`
**Describe the bug**
when changing the margin multiplier to some big value, hidden windows protrude from the edge as it appears their 'parking position' is dependant on size of the visible windows.

**To Reproduce**
Steps to reproduce the behaviour (margins > 0 need to be configured in theme):
1. Open any two windows on layout `Monocle` (layout not specific, but this shows it the easiets)
2. issue `leftwm command "SetMarginMultiplier 20"`
3. see the hidden window protruding from the edge of the screen

**Expected behaviour**
The window parking position should be determined on the unaffected workspace and window size. so it is definitely off screen.

**Screenshots**
Margin Multiplier `1`2021-11-30-210139_2560x1440_scrot
Margin Multiplier `13.5`
2021-11-30-210154_2560x1440_scrot

**Environment:**
should be irrelevant

**Additional context**
Sometimes it happens after `SoftReload`, that hidden windows protrude from the edge as well. I think this happens when the windows previously where on a different tag but are moved by 'restore logic' to the first tag.

### LINKED ISSUE #653: Telegram-desktop freezes when viewing an image on Arch Linux or ArcoLinux
**Describe the bug**
Telegram-desktop freezes when viewing an image on Arch Linux or ArcoLinux

**To Reproduce**
0. use telegram-desktop - telegram-desktop 3.4.3-2
1. Go to a channel and view an image - exit - try to see the next image then freeze occurs

We use our own pkgbuild - based on your code of r567.e93422f-1
https://github.com/leftwm/leftwm/commit/e93422f033475b1d2595a17f188baba78394dd9c

**Expected behavior**
Image closes and telegram does not freeze 

**Video**
https://youtu.be/qW2zeng2lmk

**Environment:**
 - OS/Distro: ArcoLinux and Arch Linux
 - Output of `leftwm-check` - all ok - all green

I have been observing this for several weeks now.
My workaround was to kill telegram and restart it.

Thinking it will be an update of Telegram gone wrong.

But Telegram works fine on XFCE4 - installed on the same machine leftwm is. Making the conclusion that it must be something related to leftwm.

I hope you can figure this one out. Users have been reporting this as well. Since it is no longer "just me", I post this issue.

AethanFoot:
Fixes #653, fixes #591 (as the window no longer are hidden of screen). Also standard picom fading works now (idk about the animations on jonaburgs fork though).
Thanks!


VuiMuich:
Looks like slide fading in jonaburg fork is broken by this.
How about to feature-gate (with default `on`) this for now.
Then later we can provide a sophisticated solution for parking positions (wanted to eventually add top and bottom ones anyways).

VuiMuich:
Oooh, this one does bad stuff, when you revert away from this PR. I wanted to check if the jonaburg sliding still works when I go back to `main`, but on the `SoftReload` is 'lost' all my windows, except the ones on the current tag. The processes are still all there, but they are so much unmanaged that even xprop can't see them anymore.\nWill come back with more details in a minute.

AethanFoot:
Ye working on fixing it for itself. It will cause issues when switching back and forth as the windows will be unmapped.

VuiMuich:
Do you think there is a chance to 'refind' the windows? Or is there no way around leaving the session?

AethanFoot:
Ok windows should restore properly now, however it will still "break" windows when reverting back.

VuiMuich:
@AethanFoot You are such a wizzard!! :rocket: :heart:

AethanFoot:
@VuiMuich Do animations have some effect now?

VuiMuich:
:thinking: sometimes they do, but most of the time they don't. Have no pattern yet, when they exactly trigger the animation.

AethanFoot:
Hmmm, the animations doesn't seem to want to work with this system and I don't think this is big enough to be a feature flag. This would be a good point to start using patches, where this patch would go back to similar to how it was but more extensive like how you said. The main reasons for preferring this over the old system is 1: that this works with fading in the main picom (compared to an ageing fork, which I found a bit buggy), and 2: I feel this is a more proper way of handling hiding a window and opens better ease if we add window swallowing.

Other than that I would say this is ready, Many thanks!

mautamu:
I think a libleftwm-theme-common could be pretty important for a number of reasons. I believe the issue had been brought up before as a no-go because we didn't want to make something specifically for leftwm-theme, but if leftwm itself is using it (or git for that matter) and it's modifiable outside of just leftwm-theme I think it's doable.

VuiMuich:
Agree, that having the jonaburg compatibility in a patch would be the best way to go forward.

Should we still keep patches (not that there already would exist any 😂) in `leftwm/leftwm-contrib/patches` repo and pull them from there?

AethanFoot:
Ye that would make sense. We could do a known.toml again so users can add there own repos.

PR #667, opened 2022-01-17
Fix #664

0323pin:


VuiMuich:
May I ask you a little favor? Please add `####` Title marks for each package manager topic so especially the two possibilities in NetBSD are less confusing.

0323pin:
@VuiMuich ok?

VuiMuich:
> @VuiMuich ok?

Thanks!

PR #669, opened 2022-01-18
Fix telegram and unmap windows to hide them (#663)

lex148:
* Fix telegram and unmap windows to hide them

* Restore windows properly

* clippy

* comments and maybe animation?

* Remove animation stuff, is buggy and incostistent

* Fix task bar icons flashing at times

* Set visibility after so we focus correct position

### DIFF
--- a/leftwm-core/src/display_servers/mod.rs
+++ b/leftwm-core/src/display_servers/mod.rs
@@ -2,7 +2,6 @@ use crate::config::Config;
 use crate::display_action::DisplayAction;
 use crate::models::Window;
 use crate::models::Workspace;
-use crate::state::State;
 use crate::DisplayEvent;
 #[cfg(test)]
 mod mock_display_server;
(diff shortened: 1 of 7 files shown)


lex148:
@AethanFoot I reverted this merge this morning. It was causing serious issues when I was changing between tags. Specifically the tag I was asking it to focus was not going to the workspace that was active. 

I think this is great work and needs to get into main as soon as it can, but was worried about causing regression issues for people running on main.

AethanFoot:
Hmmm, that's strange. Presuming you are using sloppy focus, I think we are missing the enter notifies for some reason will do some testing.

AethanFoot:
Ok, this is from my minimising of root window masks. Will mark the change needed

lex148:
> 

Thanks your awesome AethanFoot

AethanFoot:
@lex148 Have you been able to test this, from what I can tell it is working properly. Just want to check I fixed the right bug :laughing:.

lex148:
It sure did. thanks again. I'll Hit the merge button

PR #671, opened 2022-01-19
Make workspace layouts an option

VuiMuich:
This makes the `layouts` field in workspace config optional.
This was mainly a little papercut, when workspaces are setup manually, but share the same set of layouts.

### DIFF
--- a/leftwm-core/src/models/layout_manager.rs
+++ b/leftwm-core/src/models/layout_manager.rs
@@ -28,7 +28,6 @@ impl LayoutManager {
             .workspaces()
             .unwrap_or_default()
             .iter()
-            // .filter(|ws| ws.id.is_some() && ws.layouts.is_some())
             .map(|ws| {
                 (
                     ws.id.unwrap_or_default(),
--- a/leftwm-core/src/models/layout_manager.rs
+++ b/leftwm-core/src/models/layout_manager.rs
@@ -124,11 +124,11 @@ mod tests {
             workspaces: Some(vec![
                 crate::config::Workspace {
                     id: Some(0),
-                    layouts: vec![
+                    layouts: Some(vec![
                         Layout::CenterMain,
                         Layout::CenterMainBalanced,
                         Layout::MainAndDeck,
-                    ],
+                    ]),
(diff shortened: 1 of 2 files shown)


VuiMuich:
Should we add a test for this case?

leftwm-core/src/models/layout_manager.rs
```diff
@@ -28,7 +28,7 @@ impl LayoutManager {
             .workspaces()
             .unwrap_or_default()
             .iter()
-            .filter_map(|ws| ws.id.map(|id| (id, ws.layouts.clone())))
+            .filter_map(|ws| ws.id.map(|id| (id, ws.layouts.clone().unwrap_or_default())))
             .collect();
```
AethanFoot, line 32: I'm not sure if this would work better as a separate filter and map, as we don't really need to store a value if it is None. E.g.: 
```rust
.filter(|ws| ws.id.is_some() && ws.layouts.is_some())
.map(|ws| (ws.id.unwrap_or_default(), ws.layouts.clone().unwrap_or_default()));
```
or this might work:
```rust
.filter_map(|ws| (ws.id.unwrap_or_default(), ws.layouts.clone().unwrap_or_default()));
```

AethanFoot (approved):
Looks good.

leftwm-core/src/models/layout_manager.rs
VuiMuich: Will look into this later.

mautamu (approved):
Nothing to add to @AethanFoot 's comment. LGTM

leftwm-core/src/models/layout_manager.rs
VuiMuich: do we even need the filter then?
When trying your second suggestion I got [this clippy](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_filter_map), so I tried:
```rust
            .map(|ws| { ( ws.id.unwrap_or_default(), ws.layouts.clone().unwrap_or_default(), ) })
```
which compiles and passes the tests. Will check in a life system in a minute..

PR #673, opened 2022-01-21
[bugfix] `leftwm-check` didn't reset color

VuiMuich:
This fixes a report from Discord, by user Jalish#4407, that the line `Theme OK` did not reset the terminal color. So when no prompt styling is present any further characters in the terminal are printed in green.

While allready at this I fixed this for all `println` in `leftwm-check.rs`.


arcolinuxz:
thanks

PR #678, opened 2022-01-26
Fussy clippy

AethanFoot:
Just the fussier clippy lints (there are still the `future cannot be sent between threads safely` lints which I don't know how to deal with clearly).

Thanks!


VuiMuich:
lgtm :+1:

PR #682, opened 2022-01-28
Integrate lefthk

AethanFoot:
Pushing this as a discussion point. Until we rewrite the config code, do we want to make this a non-braking change? To do this we would need to parse the BaseCommands into external commands that lefthk can execute. This would block any lefthk commands being able to be used until the rewrite (Reload, Kill,  Chord, and ExitChord). Meaning that the user won't see any change from this other than being able to build without the hotkey daemon. 

This will be a large pr so any feedback and opinions will be much appreciated.

Thanks!


0323pin:
@AethanFoot Would this break current configurations and custom keybindings or, will `lefthk` read those? Are you guys planning on it to become a bit like `bspwm` and one could use `lefthk` with other wms or, will be integrated into `leftwm`?
I'm a bit curious of the way forward since, I'm running `leftwm` built from git-HEAD on NetBSD and, I'm considering it for my Void Linux laptop where I'd be using the release version. Will it be possible to have similar configurations on both machines?

As you can see, I'm rather curious about what this means in practice. Would you mind a slightly deeper walk through?
Thanks

AethanFoot:
So if this is going to be done before the config language change, it will likely not be a breaking change as we would be able to parse the current setup into something lefthk can run (e.g. convert it to leftwm-command ...). With lefthk I took the same approach we want for leftwm, that being the core functionality being a library which others can use to build their own version. So the integration with leftwm will use the core library of lefthk using its own implementation of the config and starting it, etc. This then would allow us to easily place the hotkey daemon shipped with leftwm behind a build flag allowing users to remove the hotkey functionality and use their own hotkey daemon, whether that be sxhkd or lefthk standalone (as the standalone version of lefthk will likely use and build upon more features than the integrated version).

There will be (and currently is, but is in very earlier but usable stages) a standalone implementation of lefthk like sxhkd, which as mentioned will likely implement more functionality of the core library and add more to it.

0323pin:
Thanks for the explanation.

VuiMuich:
> This would block any lefthk commands being able to be used until the rewrite (Reload, Kill, Chord, and ExitChord). Meaning that the user won't see any change from this other than being able to build without the hotkey daemon.

Would `Chords` really break compatibility? Not if we don't insist on a default keybind for it, right? Blocking `Kill` and `Reload` absolutely makes sense.

AethanFoot:
We could add chords with toml, it would just take a bit more parsing trickery.

VuiMuich:
> We could add chords with toml, it would just take a bit more parsing trickery.

Ahh riiight, TOML :roll_eyes: 
scratch that idea :joy:

VuiMuich:
Never really have checked what happens if you completely remove some of the default keybinds, will this break parsing?

If so we should add `use_external_hotkey_daemon: bool` as a config option, and ignore parsing errors for keybinds. Probably should throw a Warnung with `leftwm-check` like:
> WARNING: You are ignoring checks on keybinds as you set for using an external hot key daemon.

AethanFoot:
Ok rebasing didn't give the effect I was expecting.

AethanFoot:
This seems to be mostly working. However I am uncertain how we should handle killing lefthk and if lefhk crashes. Also when killed lefthk seems to be leaving ghost processes, I am unsure what is causing this.

AethanFoot:
Also to test this you need to completely kill leftwm as there are changes in the binary.

VuiMuich (commented):
Tiny review round 1:
Just a reminder, that we need to update the link to `keysym.rs` in [the wiki](https://github.com/leftwm/leftwm/wiki/Config#keybind).

leftwm/src/config/mod.rs
```diff
@@ -91,14 +91,14 @@ pub struct Config {
     pub layouts: Vec<Layout>,
     pub layout_mode: LayoutMode,
     pub scratchpad: Option<Vec<ScratchPad>>,
+    pub window_rules: Option<Vec<WindowHook>>,
     //of you are on tag "1" and you goto tag "1" this takes you to the previous tag
```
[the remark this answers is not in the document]
VuiMuich: :see_no_evil: meant the comment line of course...

leftwm/src/bin/leftwm.rs
```diff
@@ -178,3 +189,15 @@ fn handle_help_or_version_flags(args: &[String], subcommands: &BTreeMap<&str, &s
...
+    let path = BaseDirectories::with_prefix("lefthk").expect("couldn't find base directory");
+    let pipe_file = path
+        .place_runtime_file("commands.pipe")
+        .expect("couldn't find commands.pipe");
+    let mut pipe = fs::OpenOptions::new()
+        .write(true)
+        .open(&pipe_file)
+        .expect("couldn't open file");
+    writeln!(pipe, "{}", command).expect("couldn't write to file");
+}
```
VuiMuich, line 203: Is this causing the reload issue? Why have you put this here and not in the `leftwm-worker`?

AethanFoot: As this is used to send kill to lefthk when leftwm-worker reloads.

VuiMuich: Ohh, `lefthk` not `leftwm`, should read more carefully...

But anyways, shouldn't follow this the same scheme as `leftwm` with `command-<$DISPLAY>.pipe`?

AethanFoot: Yep will update lefthk, may need to add a lefthk-command to handle it properly.

AethanFoot: This may need extra care as multiple lefthks will conflict slightly when using external commands.

VuiMuich: Oh, so the extra running `lefthk` instance might be the issue for blocking reloads as well right? I auto start this with xdg_autostart.

VuiMuich: I killed the standalone `lefthk` process and now reloading works again.

VuiMuich:
I might have been late to review #704 (all-family weekend 🤷🏼‍♂️), but I have been running on this PR all the last week and haven't found any hiccups or even major problems. (Despite the one thing, that stand-alone `lefthk` and this PR don't mix well for `SotReload`ing `leftwm` due to using the same `command.pipe` file.
🚀

leftwm/src/bin/leftwm.rs
```diff
@@ -55,12 +58,17 @@ fn main() {
         env::set_var("_JAVA_AWT_WM_NONREPARENTING", "1");
 
         let worker_path = current_exe.with_file_name("leftwm-worker");
+        let lefthk_path = current_exe.with_file_name("lefthk-worker");
```
VuiMuich, line 61: I am thinking if this should possibly be ̀ hotkeydaemon_path` or something. Opinions?

leftwm/src/bin/leftwm.rs
```diff
@@ -75,6 +83,9 @@ fn main() {
                 }
                 // Either worker or autostart program exited.
             }
+            // Kill off lefthk when reloading.
+            send_lefthk_command("Kill");
```
VuiMuich, line 87: Same as line 61, maybe `send_hkd_command`

leftwm-core/src/utils/command_pipe.rs
```diff
@@ -90,35 +90,38 @@ async fn read_from_pipe(pipe_file: &Path, tx: &mpsc::UnboundedSender<Command>) -
 fn parse_command(s: &str) -> Result<Command, Box<dyn std::error::Error>> {
     let head = *s.split(' ').collect::<Vec<&str>>().get(0).unwrap_or(&"");
     match head {
+        "CloseWindow" => Ok(Command::CloseWindow),
```
VuiMuich, line 93: I like the reordering you did!
What do you think adout separating the categories with comment headlines?
Suggestions: ̀ General`, `Focus`, `Move`, `Layout`, `Workspace/Tag`
Sort stictly alphabetical inside category.

We could keep this for a separate PR though...

VuiMuich (commented):
There is a bunch of places, where I stumbled from the simlarity of ̀ leftwm` and `lefthk` and had to read twico to catch the differnce. This might be just my tired eyes and brains, but maybe we could also avoid some confusion for newcomers if we generalise it as much as possible to ̀ hotkeydaemon` or something along that.

Did I get this right that `modmask_lookud.rs` is the only keyboard related tie into ̀ xlib`? Could we feasibly use something display server independant here, already or should we keep this for later, when making the core display server agnostic?

All in all another amazing 🪄 **Wizzard Aethan** PR 😀

leftwm/src/bin/leftwm.rs
VuiMuich: Another thought in this direction, maybe this is stupid though: rename `lefthk-worker` to `lefthk-daemon` alltogether?

VuiMuich:
Nice, thanks!

The dummy `logging` feature-gate is a good idea, but don't put to much effort in it as #508 will refactor the complete logging anyways.

leftwm/src/bin/leftwm-check.rs
```diff
@@ -50,6 +50,7 @@ async fn main() -> Result<()> {
             }
             config.check_mousekey(verbose);
             config.check_workspace_ids(verbose);
+            #[cfg(feature = "lefthk")]
```
mautamu, line 53: We might want to go ahead and add the notice discussed earlier in the thread here?
```rust
#[cfg(not(feature="lefthk"))]
warn!("Ignoring checks on keybinds as you compiled for an external hot key daemon."); // or should this be a warning? idk
```

mautamu (commented):
What else do we want to accomplish with the feature gates? What's here now looks good to me

To quote Vui:
All in all though, another amazing :magic_wand: Wizard Aethan PR :grinning:

AethanFoot:
Okay, lefhk-core should be ready to be used at a reasonable level.

0323pin:
@AethanFoot Gave it a try yesterday out of curiousity.
It builds fine but panics while starting X.

It looks like it's unable to find `xkb`. Would you happen to know if it needes to be linked against `libxkbcommon` at compile time?

I might kick another build doing that if I have the time tomorrow.

AethanFoot:
Hmm, it shouldn't require anything different as the wrapper uses basically the same code that leftwm currently has. :thinking:

0323pin:
@AethanFoot Here's what I get,

`thread 'main' panicked at 'failed to start lefthk: OS { code: 2, kind: NotFound, message: "No such file or directory" }', leftwm/src/bin/leftwm.rs:75:18`

I'm now building with `libxkbcommon` linking.

AethanFoot:
Are you able to build lefthk standalone?

0323pin:
@AethanFoot Yes, I can build it standalone

```
   Compiling lefthk-core v0.1.3 (/usr/pkgsrc/wip/lefthk/work/lefthk-74033a46314c308fd0455cc959e1809752dbe162/lefthk-core)
   Compiling phf_macros v0.8.0
   Compiling phf v0.8.0
   Compiling kdl v3.0.0
   Compiling serde_kdl v0.1.0
   Compiling lefthk v0.1.3 (/usr/pkgsrc/wip/lefthk/work/lefthk-74033a46314c308fd0455cc959e1809752dbe162/lefthk)
    Finished release [optimized] target(s) in 16m 02s
```

0323pin:
@AethanFoot I'm wondering if this is the reason why it's failing?

> The configuration file should be created in ~/.config/lefthk/ and called config.kdl. If the configuration file is not created the program will exit.

AethanFoot:
That is with the implementation of the core library with the stand alone. It is setup like leftwm, where the core library doesnt care about the config side at all. Leftwm has its own implementation of lefthk-core, so this doesnt make a difference. I am wondering whether there is an issue with lefthk-core and leftwm-core using different version of x11-dl are causing some weird issue where the build tools may not be 100% on some platforms (wonder if this is an issue on void as well). If it is not that then I am a bit stumped.

AethanFoot:
Also to test lefthk, make a basic config then run the lefthk file placed in `target/release`.

0323pin:
> to test lefthk, make a basic config then run the lefthk file placed in target/release

Yeah, I did that already but, I need to re-build the binary linking to `libX11` as it can't find it otherwise.

> the core library doesnt care about the config side at all

That was also what I thought, thanks for confirming.

> I am wondering whether there is an issue with lefthk-core and leftwm-core using different version of x11-dl are causing some weird issue where the build tools may not be 100% on some platforms

This could be a possibility, although they are different binaries and should be fine having different crate depemdencies. I'll have a look at both versions of `x11-dl` but, I'm AFK today.

> wonder if this is an issue on void as well

On Void, I'm using `leftwm`-0.2.11 since a week. Moved from `frankenwm` last Sunday. But, I try not to mess with that machine, it's a back-up in case something goes wrong.

0323pin:
@AethanFoot Found some time to re-build it after all.
Now, I'm really confused. Not only, stand alone `lefthk` builds without issues but, when liking against `libX11` at compile time it works as expected.

Also, the version of `x11-dl` is the same everywhere. Looking at Cargo-lock, I see

```
leftwm-0.3.0 without lefthk

[[package]]
name = "x11-dl"
version = "2.19.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "ea26926b4ce81a6f5d9d0f3a0bc401e5a37c6ae14a1bfaa8ff6099ca80038c59"
dependencies = [
 "lazy_static",
 "libc",
 "pkg-config",
]



leftwm-0.3.0 with lefthk (your branch)

[[package]]
name = "x11-dl"
version = "2.19.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "ea26926b4ce81a6f5d9d0f3a0bc401e5a37c6ae14a1bfaa8ff6099ca80038c59"
dependencies = [
 "lazy_static",
 "libc",
 "pkg-config",
]



lefthk (stand alone)

[[package]]
name = "x11-dl"
version = "2.19.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "ea26926b4ce81a6f5d9d0f3a0bc401e5a37c6ae14a1bfaa8ff6099ca80038c59"
dependencies = [
 "lazy_static",
 "libc",
 "pkg-config",
]
```

I'm lost :(

AethanFoot:
I just thought, is your install line updated to install lefthk-worker (I probably need to update the Makefile)

0323pin:
> is your install line updated to install lefthk-worker

Oh man, this is embarrassing! I didn't even think about the obvious.

All good,

```
2022-03-06 15:06 > leftwm -V
LeftWM 0.3.0, Git-Hash: 2d1b741303
```

AethanFoot:
Ye nw, didnt think of it either.

0323pin:
@AethanFoot Day 4 using this branch and I haven't found any issues with it 👍 

Do you plan on merging this soon or to re-base the branch? `leftwm:main` i now 5 commits ahead, thought three of those are mostly cosmetic. But, the other two would be nice to have included.

VuiMuich:
IMO @AethanFoot you could go ahead and resolve and merge in one go.

VuiMuich:
This would be merge ready now the only thing is, that switching to a build based on this commit requires a restart of `leftwm-bin` in order to properly launch the `leftlk-worker` how would we like to communicate this?

leftwm/src/config/keybind.rs
```diff
@@ -36,6 +36,8 @@ impl Keybind {
             }
             BaseCommand::FocusWindowTop if value_is_some => {
                 bool::from_str(&self.value).context("invalid boolean value for FocusWindowTop")?;
+            BaseCommand::FocusNextTag => leftwm_core::Command::FocusNextTag,
+            BaseCommand::FocusPreviousTag => leftwm_core::Command::FocusPreviousTag,
```
VuiMuich: @AethanFoot I dropped these two lines with `4e9cfd3` as I had no idea, where to move them to.
So far everything is working fine here on my machine.

AethanFoot: Ye that is right as the match statement is just used to check whether a command which needs a value has a valid one.

AethanFoot:
In terms of passing that a restart is needed, it may be difficult without notifications. It may be best for this message to be added on the package maintainer side (and the make file) when they install it.

VuiMuich:
Maybe we could add a tiny `build.rs` to print a [warning](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargo-warning) at the end of compiling.

0323pin:
@VuiMuich and @AethanFoot Do you guys mean in the event an user is running `leftwm`, performs a system update and gets the WM updated? If so, it's sure on our side. I don't know what others do when upgrading their systems but, I always run upgrades as root on tty with no X running to avoid surprises. So, in my case, I wouldn't even notice this issue.

VuiMuich:
> @VuiMuich and @AethanFoot Do you guys mean in the event an user is running `leftwm`, performs a system update and gets the WM updated? If so, it's sure on our side. I don't know what others do when upgrading their systems but, I always run upgrades as root on tty with no X running to avoid surprises. So, in my case, I wouldn't even notice this issue.

Yea, especially thinking about the `leftwm-git` AUR package. Arch-Users have a tendency to run system updates in rapid fire, you know?
And the state after a reload is a bit awkward, as the WM remains working fine, just all keybinds seem gone. I guess manually launching `lefthk-worker` might be a temporary workaround, but you would need to be aware of that of course.
But maybe going ahead and keep spamming "have you tried logging out and back in again" support messages on discord is actually a valid option 🤣

Speaking of updates: I for myself rebuild `leftwm` often several times a day, to checkout various branches, so that's how I experienced this issue.

VuiMuich:
🤔 suddenly I have again issues with reloading, while on this branch: I have to restart `leftwm-worker` manually after each reload.

0323pin:
@VuiMuich I'm seeing the same but not on every reload. I had to kill X and re-start it while playing around with `template.liquid`

AethanFoot:
Is it only on this branch, or is it on main. I don't see anything in the merge that would cause issues. Will do some testing now.

0323pin:
@AethanFoot I've just re-built from [this commit](https://github.com/AethanFoot/leftwm/commit/ca50136d5103c07c5b5cfd0f1d2087ff0c61d15d) and the issue with reloading is there also. It doesn't happen every time, though.

VuiMuich:
> Is it only on this branch, or is it on main. I don't see anything in the merge that would cause issues. Will do some testing now.

Previously I experienced this only when switching from this branch to another or vice versa, but now it also happened when I did some tests to add a `build.rs` (struggling to get it triggered so far, btw).

0323pin:
> When first time building with `lefthk` you need to completely restart `leftwm` in order to start the hotkey daemon proprerly. A `SoftReload` or `HardReload` will leave you with a session non responsive to keybinds but otherwise running well.

@VuiMuich The thing is that I'm getting into this situation when doing a 'SoftReload' after changing settings in for _e.g._ `template.liquid`. This doesn't happen every time but I can trigger it rather often.

Say I edit `template.liquid` and have a typo or, simply put in something that's not supposed to work and do a 'SoftReload', then I'm also getting into _a session non responsive to keybinds but otherwise running well._

I've been running on this branch for more then a week now.

VuiMuich:
Hmm, today I had none of these issues and was reloading a lot, when testing my local merge of this PR and #734.
I would absolutely love to see this merged, but if this makes reloads a game of chance we probably should give this another round of trial.

VuiMuich:
@0323pin could you try, if having symlinking `lefthk-worker`just like the other `leftwm*` biraries helps?
Also, where is `which lefthk-worker` pointing to? For me it is `/sbin/lefthk-worker` :thinking:

0323pin:
@VuiMuich To answer this you need to understand **how** NetBSD separates things and how I have installed the package. I'll try to be as brief as possible.

Unlike Linux, NetBSD is a full OS and not a _kernel + tools_. The tools are developed together with the kernel and form what is usually refer to as **base system**.

The base system contains the kernel, core-utils, xorg, some fonts,..., you get the picture. These are installed in `/bin`, `/sbin`, `/usr/bin`, `/usr/sbin` and `/usr/X11R7`. These directories are reserved to _base system_ binaries.

Third party software, like your WM or DE, your web-browser, your file-manager,... go into `/usr/pkg/bin` (not into `/usr/local` like on FreeBSD or OpenBSD, `/usr/local` on NetBSD is for **the user**, see it like `/opt` on Linux).

Now, I've made the `leftwm` package, so naturally the binaries are installed into `/usr/pkg/bin`. I keep this location also when building them from git.

With this in mind, what would you like me to test? I'm running `leftwm`-0.2.11 on Void musl also and should see if this is something that was already there. I find it hard to believe that it's related to `lefthk`to be honest.
Also, I can `SoftReload` without issues sometimes, the issue isn't triggered on every reload.

VuiMuich:
> > For me it is `/sbin/lefthk-worker` thinking
> 
> Wait, wait. I'm sorry, I know this is completely off-topic, but _how_ did the binary end up in `/sbin`? `/sbin` is usually reserved for binaries that are required to boot the system. Again, I'm sorry. I follow along with LeftWM's development (even if I'm not that active), but sometimes there are little things like this that intrigue me.

Ah, my bad I messed up my symlink: I didn't put it in `/usr/local/bin` but in `/usr/bin` which `sbin` is symlinked to. Now I have a clean linking:
```
>>> ll /usr/local/bin/left*
lrwxrwxrwx 54 root root 17 Mar 11:13  /usr/local/bin/lefthk-worker -> /home/vuimuich/Git/leftwm/target/release/lefthk-worker
lrwxrwxrwx 47 root root 10 Nov  2021  /usr/local/bin/leftwm -> /home/vuimuich/Git/leftwm/target/release/leftwm
lrwxrwxrwx 53 root root 10 Nov  2021  /usr/local/bin/leftwm-check -> /home/vuimuich/Git/leftwm/target/release/leftwm-check
lrwxrwxrwx 55 root root 10 Nov  2021  /usr/local/bin/leftwm-command -> /home/vuimuich/Git/leftwm/target/release/leftwm-command
lrwxrwxrwx 53 root root 10 Nov  2021  /usr/local/bin/leftwm-state -> /home/vuimuich/Git/leftwm/target/release/leftwm-state
lrwxrwxrwx 54 root root 10 Nov  2021  /usr/local/bin/leftwm-worker -> /home/vuimuich/Git/leftwm/target/release/leftwm-worker
```

@0323pin thanks for this brief introduction to NetBSD. So your `lefthk-worker` is in `/usr/pkg/bin` as any other `leftwm` related binaries, right?

I am pretty sure this is something in `0.3.0` for sure. If it is not coming from this lefthk branch, it might be related to the `command-<n>.pipe` change if you can reproduce on `main` and after a 'clean restart' of leftwm or even X11.

0323pin:
> So your lefthk-worker is in /usr/pkg/bin as any other leftwm related binaries, right?

Correct.

>  if you can reproduce on main and after a 'clean restart' of leftwm or even X11.

This I can do. I'll kick a build from the latest commit in main and will report back.

0323pin:
@VuiMuich 😞 I can't seem to repruduce it in the main branch. I'll try a few more times, though. @AethanFoot does a change in the configuration trigger a restart of `lefthk-worker`?

VuiMuich:
> @VuiMuich 😞 I can't seem to repruduce it in the main branch. I'll try a few more times, though. 

Well in terms of triaging this is some progress as well 😀 

> @AethanFoot does a change in the configuration trigger a restart of `lefthk-worker`?

as far as I could tell it from the code in `leftwm.rs` it might be that the restart might occur before the kill was finished.

AethanFoot:
Ye the issue seems to be that lefthk is missing the kill command meaning that we hang trying to reap lefthk.

VuiMuich:
> Ye the issue seems to be that lefthk is missing the kill command meaning that we hang trying to reap lefthk.

So lefthk misses to catch the kill command, or is leftwm missing to send?

Regarding other open PRs, especially #737 and #738 I guess we proceed to merge and merge main once again here, right?

0323pin:
I still can't reproduce the _keybindings not responding after reload_ on the main branch.

OT: @AethanFoot Anything against if I create a stand-alone package for `lefthk`, so users could use this in place of, say `sxhkd`? Or, would you prefer I wait?

AethanFoot:
Ok this should be working now, couldn't be bothered to work out why the command wasn't coming through, so just forced it to die.

VuiMuich:
Would we want to leave a _TODO_ comment to implement a clean kill later?

Most certainly have time today evening to check this out and possibly do a bunch of merges of open PRs.
So far I don't think we have critical overlaps with other PRs that might need further testing after rebase, right? The closest would be #734, but that's already running well locally for me.

AethanFoot:
Yeah they all look fine, would just need to add the new commands to the keybind parser

VuiMuich:
@0323pin if there is no hickups or regression that I introduced while resolving merge conflicts this will go live asap.

0323pin:
@VuiMuich Do you mean you're merging this into main soon or, do you want me to build it from this branch and test?

VuiMuich:
> @VuiMuich Do you mean you're merging this into main soon or, do you want me to build it from this branch and test?

Most probable gonna merge this evening. If you have time to test that would be much appreciated, yes 🙏

0323pin:
Sure, no problem. I'll start a build now and report back later in the afternoon.

0323pin:
@VuiMuich and @AethanFoot Sorry, but the issue is still there 😢 

Actually, it's even worst as killing X doesn't do it anymore, it requires a reboot.
You can find my config here, https://codeberg.org/pin/leftwm-netbsd/src/branch/main/home/.config/leftwm if you'd like to try to reproduce it.

The only thing I did was adding/removing a dash from `%{F#232323}%{B#FADB9B}  {{ "now" | date: "v%W  %Y-%m-%d" }}  %{B-}%{F-}` in `template.liquid` and do a `SoftReload`, _i.e._ change `{{ "now" | date: "v%W  %Y-%m-%d" }}` to `{{ "now" | date: "v%W - %Y-%m-%d" }}` and reload, keybindings stopped working.

Don't know if this tell you something but, I could read it in the tty upon restarting my laptop,
`thread main panicked at Null pointer in display, leftwm-core/src/display_servers/xlib_display_server/xwrap/mod.rs:99:9`

It would also be nice to move the message `When first time building with `lefthk` you need to completely restart `leftwm` in order to start the hotkey daemon proprerly. A `SoftReload` or `HardReload` will leave you with a session non responsive to keybinds but otherwise running well.` to the end of compilation, as now it displays in the middle of the build and people may miss it.

Sorry if this weren't the news you were expecting. Happy to test any changes.

AethanFoot:
Hmm, that is strange. The reloading (with or without the local time changes) seems to be working on my end.

I noticed before that the build warning was appearing randomly, the only issue is I don't think you can chose where it appears.

0323pin:
Yeah but, I've just rebuild it from the `main` branch and this doesn't happen there 😮

VuiMuich:
Yeah, I tried different things to force it to the end, but so far nothing worked out.
I even asked on the rust discord, but got no answer 🥲

Also tried to trigger a `notify-send`, without success. But this was probably just missing some little detail there so if you have any idea how that should work it would be great to have this in the build script.

0323pin:
@VuiMuich Not at compile time, I don't. But, at least at our end, I can easily add an install message that displays on screen when installing the package. The problem with this approach is that people tend to miss those but, it's better than nothing 😄 

I'm more worried about the issue with the keybindings. Can you reproduce it?
I've only added/removed a `-` in `templete.liquid` and did a `SoftReload`

VuiMuich:
> @VuiMuich Not at compile time, I don't. But, at least at our end, I can easily add an install message that displays on screen when installing the package. The problem with this approach is that people tend to miss those but, it's better than nothing 😄

Yeah, same for the AUR packages. At least you can say 'it said so during the install/update' why didn't you read carefully 😝 

> I'm more worried about the issue with the keybindings. Can you reproduce it? I've only added/removed a `-` in `templete.liquid` and did a `SoftReload`

I'll have to check when I am home from work. Maybe I didn't notice this issue if it is actually ̀ liquid` related, as i parse the `json` of `leftwm-state` directly in eww and skip the template...

VuiMuich:
Just to verify: when you reload without changes to `template.liquid` it works clean?

I ran `leftwm-state -w0 -t Git/leftwm/themes/basic_polybar/template.liquid` in alacritty to test this but the process dies immediately (I guess in the woke of `leftwm-worker` being killed) during the `SoftReload`.
Do you get the expected output, when you run `leftwm-state` in terminal with your template? I have a suspicion a failing template might hinder `lefthk-worker` to be reaped.

0323pin:
Sorry, had to do other things and was away.

> Upate: I installed lemonbar and switched to the basic_lemonbar theme and immediately had the issue you described, that the old lefthk-worker did not die.

Am I glad you could reproduce it?
Yes and no 😞

This doesn't look good _in the wild_ but, I can live with it 😀

AethanFoot:
One I have noticed with the lemonbar config is it still has localtime which was removed.

AethanFoot:
I changed to the basic polybar theme, and messed around with the date line and found no correlation to issues happening. The thing I found is every now and then the keybinds don't seem to start up properly. When they stopped working it wasn't the same as before as leftwm reloaded completely (which meant lefthk was killed), it seemed to just not grab the bindings.

VuiMuich:
I am biting myself that I missed to pay attention to the PID before I manually killed `lefthk-worker` but I have a strong suspicion that the original process somehow survives and therefore causes some confusion among the command-pipe files.
I'll try to confirm this later.

0323pin:
@AethanFoot 
> One I have noticed with the lemonbar config is it still has localtime which was removed.

Which, I'm not using, here's my `template.liquid`,

```
{% assign mine_open = '%{F#232323}%{B#A6BC99}' %}
{% assign mine_close = '%{B-}%{F-}' %}
{% assign visible_open = '%{F#232323}%{B#C0C5CE}' %}
{% assign visible_close = '%{B-}%{F-}' %}
{% assign busy_open = '%{F#232323}%{B#FADB9B}' %}
{% assign busy_close = '%{B-}%{F-}' %}

{% for tag in workspace.tags %}
{% if tag.mine %}
{{mine_open}}  {{ tag.name }}  {{mine_close}}
{% elsif tag.visible  %}
{{visible_open}}  {{ tag.name }}  {{visible_close}}
{% elsif tag.busy  %}
{{busy_open}}  {{ tag.name }}  {{busy_close}}
{% else %}
  {{ tag.name }}   
{% endif %}
{% endfor %}

%{c}
{{ window_title | truncate: 20 }}

%{r}
%{F#232323}%{B#A6BC99}  {{ workspace.layout }}  %{B-}%{F-}
%{F#232323}%{B#FADB9B}  {{ "now" | date: "v%W  %Y-%m-%d" }}  %{B-}%{F-}
```

@VuiMuich 
> So I guess my theory is proven wrong by this

Did you get any further?
I guess you guys could just merge this as it seems is only affecting `lemonbar`, I'll try to remember to reboot instead of reload when changing something.

The thing is that `polybar` has so many things that would need to be switched-off to make a package, that at this point is no longer worthy. As for `eww`, it doesn't build on NetBSD do some capricious requirement on `sysinfo`, I haven't looked if that can be switched-off at compile time. So, `lemonbar` it is.

AethanFoot:
I am hoping to be able to get this started up again soon, as I only have my dissertation and a couple exams left. Now I have a fresh pairs of eyes hopefully I can get this rock solid (with the main issue being lefthk just doesn't start every time).

Many Thanks!

0323pin:
@AethanFoot keep us informed 😄 
Thanks

AethanFoot:
Okay this may be fixed and sorted now (hopefully didn't bork on the merge). Sorry for the delay, lifes been a tad s***.
Many Thanks!

0323pin:
Awesome 😀
I'll build it from your branch @AethanFoot as soon as I find the time, most probably not before Monday.

0323pin:
Unfortunately, I can not build this without a lot of work around/patching. I need to link at compile time against `libX11` for the WM to work, which means I need to build it inside pkgsrc infrastructure and not simply using `cargo`.

The problem is that within pkgsrc **online** builds aren't that easy, it's designed to avoid fetching distfiles from the internet at compile time. So, every crate should be a release to be able to fetch it before starting the build.

This, `lefthk-core = { git = "https://github.com/leftwm/lefthk", branch = "simplify_and_get_ready", optional = true }` in `Cargo.toml` causes a lot of issues. The build stops and complains the needed distfile can't be fetched. As said, I can patch this to fetch the specific commit. But, I guess I'll just wait until this is fixed at your end.

AethanFoot:
Source is now from crates.io

0323pin:
@AethanFoot Thank you! I'm running on this branch now and intend to stay on it until it's merged.
Still, I have the same issue with `SoftReload`, although now it's more random. Sometimes I can `SoftReload` and everything works and, sometimes the keybindings are dead and I need to kill X and restart it.

I'm happy with my theme, though, so I have no need to reload it. I've been running on this theme for 10 months.

AethanFoot:
Hmm, yeah. Just about managed to get it to fail after like 10+ tries, thats annoying but better.

0323pin:
@AethanFoot It's much more stable than before. Do let me know if/when to re-build, I'm happy to help testing.

AethanFoot:
I cant seem to get lefthk-core 0.1.7 to crash.

VuiMuich:
Haven't seen any issues the last 4 days either, but tbh I haven't had a terrible amount of reloads either.

---

On a side note (and this might actually be worth an issue in the `lefthk` repo): I think there should be the possiblilty to create a keybind like (using `ron` :smile:):
```rust
keybind: [
   (command: Execute,
    value: "rofi -show drun",
    modifier: ["modkey"],
    key: [])
]
```

0323pin:
@AethanFoot Neither do I 🎆 

Time to get rid of this warning 😄 

```
warning: When first time building with `lefthk` you need to completely restart `leftwm` in order to start the hotkey daemon proprerly. A `SoftReload` or `HardReload` will leave you with a session non responsive to keybinds but otherwise running well.
```

One thing, though. Somehow our script didn't pick `lefthk-core` as a dependency, I've added it manually before fetching the disfiles and everything's working.

Anyway, the annoying bug seems to be gone and `SoftReload` is working again. Congratulations!!

AethanFoot:
Yeah, think crates.io is having a update error as the nix package seems to be having a similar issue.

0323pin:
I see it now. Weird, I've added the dependency manually and it did fetch it.
I'll keep riding on this branch but, I don't expect to find any issues as I did try tenths of reloads to see if I could trigger the issue.
As far as I'm concerned, it's ready for prime time.

Which brings me back to a question I had before. Are you ok if I package and merge `lefthk` as a standalone application that maybe used with other WM's that do not provide their own keybinding daemon?

VuiMuich:
> Time to get rid of this warning smile
> 
> ```
> warning: When first time building with `lefthk` you need to completely restart `leftwm` in order to start the hotkey daemon proprerly. A `SoftReload` or `HardReload` will leave you with a session non responsive to keybinds but otherwise running well.
> ```
I don't think we can loose this yet, as we still need to kill the `leftwm` bin so `leftlk-worker can be launched.
The only way to avoid this, that  comes to mind is if `leftwm-worker` would be responsible for starting the hotkey daemon.

0323pin:
@VuiMuich You're right, of course.

0323pin:
One week running on this branch, no issues ❤️ 

Whenever you are ready to merge, `lefthk`is ready 😄

AethanFoot:
Awesome, will make it look neater and stuff then should be ready. 
Many Thanks!

0323pin:
@AethanFoot Any chance to merge the last 8 commits from Main into your branch? I've been running on this branch for 22 days and I haven't had any issues but, it would be nice to get the latest improvements.

AethanFoot:
Okay, for some reason nix (and I think the same for you) doesn't pick up lefthk-core when a general version is given, e.g. 0.1. However now nix is detecting the new version, but atleast it is detecting it. (Also the last version should a 0.2 but I will update that later and maybe it will update).

0323pin:
Thanks, that was quick!
I'll kick a build later today. Yeah, I had issues picking `lefthk-core` from crates.io last time but, adding it manually fixed it. I'll see if it happens again. If you bump it, just let me know.

leftwm/src/bin/leftwm-check.rs
VuiMuich: This should be more in line with the rest of the current `leftwm-check` code
```suggestion
            #[cfg(not(feature = "lefthk"))]
            println!("\x1b[1;93mWARN: Ignoring checks on keybinds as you compiled for an external hot key daemon.\x1b[0m");
           #[cfg(feature = "lefthk")]
```

leftwm/src/bin/leftwm-worker.rs
```diff
@@ -4,8 +4,7 @@ use slog::{o, Drain};
 use std::panic;
 
 fn main() {
-    //let _log_guard = setup_logfile();
-    #[cfg(feature = "logging")]
+    // let _log_guard = setup_logfile();
```
VuiMuich, line 7:
```suggestion
    #[cfg(feature = "log-to-file")]
    let _log_guard = setub_logfile();
```
and of course add this feature to `leftwm/Cargo.toml`

VuiMuich (approved):
As the changes from a few weeks ago were already ok, I don't see any trouble with this new commits.
The two little suggestions above are no real blockers for me.

hertg (commented):
I wasn't involved with that PR, so I won't start reviewing it now, since I think it's pretty much ready based on the already posted reviews.

image
I do like that a lot

AethanFoot:
Is it worth making a discussion of things we can modularise, things like you did with layouts. A error handling module would be useful I feel, also we want to make the display servers modules as well.

hertg:
I think it could make sense to encapsulate and thoroughly test some logic. But we probably shouldn't overdo it, otherwise we'll need to create releases of 5+ crates every time we want to change something :stuck_out_tongue_closed_eyes: 
We probably should discuss specifics on Discord or in the Discussions tab :)

0323pin:
@AethanFoot All good :)
Running on the latest commit from one hour ago. It fetched `lefthk-core` correctly, no manual intervention this time.
Thank you!

0323pin:
Actually, something is wrong. I can not cycle through open windows with the keybindings 😢 
It works if all the windows are terminals but, not if one of them is a GUI app, with the exception of firefox.
The bar says I'm cycling through the open widows but, it doesn't focus the new window. Moreover, if I close the window it doesn't go away until I click another one. I can also focus the window by clicking on it but, not with the keybindings.

Just in case,
```
# leftwm config.toml

modkey           = "Mod4"
mousekey         = "Mod1"
layout_mode      = "Tag"
layouts          = [
    "MainAndVertStack",
    "CenterMainBalanced",
    "EvenVertical",
]

focus_behaviour  = "ClickTo"
focus_new_window = true
tags             = ["1", "2", "3", "4", "5", "6"]

[[keybind]]
command          = "Execute"
value            = "alacritty"
modifier         = ["modkey"]
key              = "Return"

[[keybind]]
command          = "Execute"
value            = "firefox"
modifier         = ["modkey"]
key              = "b"

[[keybind]]
command          = "Execute"
value            = "featherpad"
modifier         = ["modkey"]
key              = "e"

[[keybind]]
command          = "Execute"
value            = "alacritty -e spotify_player"
modifier         = ["modkey"]
key              = "s"

[[keybind]]
command          = "Execute"
value            = "alacritty -e tuifeed"
modifier         = ["modkey"]
key              = "a"

[[keybind]]
command          = "Execute"
value            = "xpdf -rv"
modifier         = ["modkey"]
key              = "p"

[[keybind]]
command          = "Execute"
value            = "oculante"
modifier         = ["modkey"]
key              = "o"

[[keybind]]
command          = "Execute"
value            = "mixerctl -w outputs.master-=12"
modifier         = []
key              = "XF86XK_AudioLowerVolume"

[[keybind]]
command          = "Execute"
value            = "mixerctl -w outputs.master+=12"
modifier         = []
key              = "XF86XK_AudioRaiseVolume"

[[keybind]]
command          = "Execute"
value            = "scrot -d2"
modifier         = ["modkey", "Shift"]
key              = "s"

[[keybind]]
command          = "Execute"
value            = "pkill leftwm"
modifier         = ["modkey", "Shift"]
key              = "q"

[[keybind]]
command          = "Execute"
value            = "please reboot"
modifier         = ["modkey"]
key              = "Delete"

[[keybind]]
command          = "Execute"
value            = "please shutdown -p now"
modifier         = ["modkey", "Shift"]
key              = "Delete"

[[keybind]]
command          = "TileToFloating"
modifier         = ["modkey"]
key              = "f"

[[keybind]]
command          = "FloatingToTile"
modifier         = ["modkey"]
key              = "t"

[[keybind]]
command          = "CloseWindow"
modifier         = ["modkey", "Shift"]
key              = "c"

[[keybind]]
command          = "SoftReload"
modifier         = ["modkey", "Shift"]
key              = "r"

[[keybind]]
command          = "MoveWindowUp"
modifier         = ["modkey", "Shift"]
key              = "k"

[[keybind]]
command          = "MoveWindowDown"
modifier         = ["modkey", "Shift"]
key              = "j"

[[keybind]]
command          = "FocusWindowUp"
modifier         = ["modkey"]
key              = "k"

[[keybind]]
command          = "FocusWindowDown"
modifier         = ["modkey"]
key              = "j"

[[keybind]]
command          = "NextLayout"
modifier         = ["modkey", "Shift"]
key              = "z"

[[keybind]]
command          = "PreviousLayout"
modifier         = ["modkey", "Shift"]
key              = "x"

[[keybind]]
command          = "IncreaseMainWidth"
value            = "2"
modifier         = ["modkey"]
key              = "l"

[[keybind]]
command          = "DecreaseMainWidth"
value            = "2"
modifier         = ["modkey"]
key              = "h"

[[keybind]]
command          = "GotoTag"
value            = "1"
modifier         = ["modkey"]
key              = "1"

[[keybind]]
command          = "GotoTag"
value            = "2"
modifier         = ["modkey"]
key              = "2"

[[keybind]]
command          = "GotoTag"
value            = "3"
modifier         = ["modkey"]
key              = "3"

[[keybind]]
command          = "GotoTag"
value            = "4"
modifier         = ["modkey"]
key              = "4"

[[keybind]]
command          = "GotoTag"
value            = "5"
modifier         = ["modkey"]
key              = "5"

[[keybind]]
command          = "GotoTag"
value            = "6"
modifier         = ["modkey"]
key              = "6"

[[keybind]]
command          = "MoveToTag"
value            = "1"
modifier         = ["modkey", "Shift"]
key              = "1"

[[keybind]]
command          = "MoveToTag"
value            = "2"
modifier         = ["modkey", "Shift"]
key              = "2"

[[keybind]]
command          = "MoveToTag"
value            = "3"
modifier         = ["modkey", "Shift"]
key              = "3"

[[keybind]]
command          = "MoveToTag"
value            = "4"
modifier         = ["modkey", "Shift"]
key              = "4"

[[keybind]]
command          = "MoveToTag"
value            = "5"
modifier         = ["modkey", "Shift"]
key              = "5"

[[keybind]]
command          = "MoveToTag"
value            = "6"
modifier         = ["modkey", "Shift"]
key              = "6"
```

VuiMuich:
Ooooh, good catch. I can reproduce. This affects focus and moving windows.
Also managed to get the tiled windows to an unmapped state so the were unfocussable with `Driven` focus mode and scratchpad window rendering behind them (we already had this issue some time ago, maybe it got 'refactored' in recently?).

0323pin:
@VuiMuich I can try tomorrow, already very late here and I need to get up in 5 hours.
It was not present in 451f6861abd3d3b00de079037fd95540c763a9fa

AethanFoot:
This seems to only happen in ClickTo and Driven, and appears to be a lack of update rather than it not happening. Also picom doesn't start when reloading but I suspect that is because we moved the up scripts to before the first loop (maybe we are too fast for it)

VuiMuich:
:thinking: I thought I put `EventResponse::Continue` in places, where the original code befor the evevnloob refactor continued out of the loop - but seems like the jumping through function calls messes with the order of things happening I guess.

0323pin:
I'll try to kick another build when I'm home late this evening. Can't promise I'll manage, though. Else, early on Thursday morning, sorry, on mobile at the moment.

VuiMuich:
On my system all running fine now.

0323pin:
@AethanFoot, @VuiMuich 😪 All good, everything is working as it should but, ... you knew that already 😉 
`picom` is also working after a soft-reload.

0323pin:
@AethanFoot ❤️ 🥳

AethanFoot:
@0323pin on another note, does zbus still fail to build on netbsd?

0323pin:
@AethanFoot haven't tried in a while. Easiest if I had something to test build. Would you like me to try building it stand-alone?

AethanFoot:
@0323pin Yeah that would be helpful.

0323pin:
@AethanFoot I'll give it a go after building the new git-head from `leftwm` and `starship` latest release.
Back soon!

0323pin:
@AethanFoot Ready to test a build of whatever you have in mind 😄 

```
   Compiling zbus_macros v3.1.0 (/usr/pkgsrc/wip/zbus/work/zbus-zbus-3.1.0/zbus_macros)
   Compiling zvariant v3.6.0 (/usr/pkgsrc/wip/zbus/work/zbus-zbus-3.1.0/zvariant)
   Compiling zbus_names v2.2.0 (/usr/pkgsrc/wip/zbus/work/zbus-zbus-3.1.0/zbus_names)
   Compiling zbus v3.1.0 (/usr/pkgsrc/wip/zbus/work/zbus-zbus-3.1.0/zbus)
   Compiling zbus_polkit v2.0.2 (/usr/pkgsrc/wip/zbus/work/zbus-zbus-3.1.0/zbus_polkit)
   Compiling zbus_xmlgen v2.0.1 (/usr/pkgsrc/wip/zbus/work/zbus-zbus-3.1.0/zbus_xmlgen)
    Finished release [optimized] target(s) in 4m 30s
```

0323pin:
Oh, and I had to patch `zvariant` 

```
--- zvariant/Cargo.toml.orig	2022-09-02 17:10:31.000000000 +0000
+++ zvariant/Cargo.toml
@@ -35,7 +35,7 @@ time = { version = "0.3.9", features = [
 doc-comment = "0.3.3"
 serde_json = "1.0"
 serde_repr = "0.1.5"
-glib = { git = "https://github.com/gtk-rs/glib", rev = "c9ee583cea0" }
+glib = "0.10.0"
 rand = "0.8.3"
 criterion = "0.3"
```

As already mentioned, we don't pull distfiles at compile time, our builds are all done in `--offline` mode.

AethanFoot:
I presume it is good now then, I remember before it wasn't compiling (which u mentioned on future plans issue). Just wanted to check as I was gonna dive into us using dbus for things.

0323pin:
> I presume it is good now then

Yeah! Go ahead and dump something to build 🤣 

> Is that zvariant an issue?

It depends. If there's some functionality that is required, yes. They should release a new version, so we can fetch distfiles before compiling. If, it's just for development, no. I don't mind keeping a patch.

0323pin:
Oh, btw.
If you want, we don't need to keep this discussion here. Just drop me a line directly pin@NetBSD.org

0323pin:
@AethanFoot Hmm ... the latest changes should warrant a new release. It's miles ahead of 0.3.0.

We have a commit freeze coming-up soon. Is there a plan for when 0.4.0 should hit the wild? From our side, within the next few days or, on the last week of this month.

0323pin:
@AethanFoot there might still be issues with `zbus`, from a fellow dev https://gitlab.gnome.org/World/pika-backup/-/issues/269

I know it's not the same. In any case what's wrong with `dbus-rs` and why is `zbus` a prefered choice?

hertg:
> I know it's not the same. In any case what's wrong with `dbus-rs` and why is `zbus` a prefered choice?

To my knowledge `zbus` is completely written in rust whereas dbus-rs is "just" rust bindings to a C library. It also seems to me that going through the C library is less performant, but I can't link any benchmarks to back that up right now. AFAIK dbus-rs is probably more "stable" but `zbus` is more performant + rust native and it seems to still be well maintained. But maybe @AethanFoot has more info or can correct me if I'm wrong.

0323pin:
@hertg thanks for the feedback. Hopefully, we won't run into trouble with `zbus`.

AethanFoot:
Yeah, again I don't mind either or. To add to @hertg points, is that they have some nice abstractions that make the interfaces easy to setup and interact with through code. These abstractions obviously make it a bit easier on me, but have a greater impact for others working on the code, as the easier it is to understand the more can work on it. Their book shows nicely how to do it from a super low level point (not quite as low as the c-bindings), and how the abstractions are built up, https://dbus.pages.freedesktop.org/zbus/introduction.html . I am thinking of testing this with lefthk first which has lower demands than leftwm.

0323pin:
Thanks! Do let me know when there's something to test. Hope not to hit the same issue where the application is unable to find `dbus`, although it is running, just because it's looking in the wrong place.

0323pin:
@AethanFoot Go ahead and implement `zbus`. I think I found the reason for my fellow dev issue above after he sent me some debug data.

In case you are interested, here are my thoughts on the debug data. Please do say if I misunderstood it.

```
Let's see if I can make some sense of my thoughts.
Initially, your output looks just fine an it connects to dbus.

signal time=1663583063.914001 sender=org.freedesktop.DBus -> destination=:1.4 serial=2 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameAcquired
   string ":1.4"
method call time=1663583063.914245 sender=:1.4 -> destination=org.freedesktop.DBus serial=2 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=RequestName
   string "org.gnome.World.PikaBackup"
   uint32 4
signal time=1663583063.914257 sender=org.freedesktop.DBus -> destination=(null destination) serial=6 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameOwnerChanged
   string "org.gnome.World.PikaBackup"
   string ""
   string ":1.4"

But then, it looks like it's expecting a second connection.
After all, that's what dbus is supposed to do, intraprocess communication.

method call time=1663583064.131676 sender=:1.4 -> destination=org.freedesktop.DBus serial=3 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=AddMatch
   string "type='signal',sender='org.freedesktop.DBus',interface='org.freedesktop.DBus',member='NameOwnerChanged',path='/org/freedesktop/DBus',arg0='org.gnome.SessionManager'"
method return time=1663583064.131706 sender=org.freedesktop.DBus -> destination=:1.4 serial=5 reply_serial=3
method call time=1663583064.131787 sender=:1.4 -> destination=org.freedesktop.DBus serial=4 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=GetNameOwner
   string "org.gnome.SessionManager"
   
Here it's lookig for another process to connect from dbus.
In this case is looking for Gnome session.

When it doesn't find it, it goes on and removes the match attempt.

error time=1663583064.131795 sender=org.freedesktop.DBus -> destination=:1.4 error_name=org.freedesktop.DBus.Error.NameHasNoOwner reply_serial=4
   string "Could not get owner of name 'org.gnome.SessionManager': no such name"
method call time=1663583064.132059 sender=:1.4 -> destination=org.freedesktop.DBus serial=5 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=RemoveMatch
   string "type='signal',sender='org.freedesktop.DBus',interface='org.freedesktop.DBus',member='NameOwnerChanged',path='/org/freedesktop/DBus',arg0='org.gnome.SessionManager'"

And goes looking for another option, Xfce session.

method return time=1663583064.132068 sender=org.freedesktop.DBus -> destination=:1.4 serial=7 reply_serial=5
method call time=1663583064.132165 sender=:1.4 -> destination=org.freedesktop.DBus serial=6 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=AddMatch
   string "type='signal',sender='org.freedesktop.DBus',interface='org.freedesktop.DBus',member='NameOwnerChanged',path='/org/freedesktop/DBus',arg0='org.xfce.SessionManager'"
method return time=1663583064.132175 sender=org.freedesktop.DBus -> destination=:1.4 serial=8 reply_serial=6
method call time=1663583064.132290 sender=:1.4 -> destination=org.freedesktop.DBus serial=7 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=GetNameOwner
   string "org.xfce.SessionManager"
   
Which, it also doesn't find
   
error time=1663583064.132300 sender=org.freedesktop.DBus -> destination=:1.4 error_name=org.freedesktop.DBus.Error.NameHasNoOwner reply_serial=7
   string "Could not get owner of name 'org.xfce.SessionManager': no such name"
method call time=1663583064.132527 sender=:1.4 -> destination=org.freedesktop.DBus serial=8 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=RemoveMatch
   string "type='signal',sender='org.freedesktop.DBus',interface='org.freedesktop.DBus',member='NameOwnerChanged',path='/org/freedesktop/DBus',arg0='org.xfce.SessionManager'"

My guess is that after this it goes on and tries to connect to any Desktop,
there are repeated attempt to connect to 'org.freedesktop.portal.Desktop'
Until, it finally gives-up and drops the initial connection to pika-backup.


signal time=1663583064.243915 sender=org.freedesktop.DBus -> destination=(null destination) serial=8 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameOwnerChanged
   string "org.gnome.World.PikaBackup"
   string ":1.4"
   string ""
signal time=1663583064.243973 sender=org.freedesktop.DBus -> destination=:1.4 serial=9 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameLost
   string ":1.4"

In other words, I think zbus is working and the initial connection is established
between pika-backup and dbus.

But, this is only half of what it's needed, it requires a second connection.
When this can not be established, dbus drops the initial one and you get the error
that you have reported on your issue.
Although, I do agree that the error message could be better explained.

If we look at the error message with the above in mind it actually makes sense.

'Failed to create ZBus session connection'

It actually didn't fail to create a connection to zbus, it failed to connect to a session from it.

I start to question if this software is even supposed to run on it's own.
Apparently, it needs a full desktop session running.
```

Issue #668, opened 2022-01-18
Dialogue windows getting cut-off


0323pin:
**Describe the bug**
Dialogue windows get cut-off when tiling client size is smaller than the dialogue window.
![2022-01-18-091558_1366x768_scrot](https://user-images.githubusercontent.com/90570748/149902498-caae19a3-6c14-4b04-a128-70307c04f11b.png)

![2022-01-18-091620_1366x768_scrot](https://user-images.githubusercontent.com/90570748/149902534-7458462f-0906-4813-a883-a3137ca14ee3.png)

![2022-01-18-091708_1366x768_scrot](https://user-images.githubusercontent.com/90570748/149902571-37fc51a5-3fc3-4b92-a95c-18b85fa85e2d.png)

**To Reproduce**
Any action that triggers a dialogue window on a small tile.

**Expected behavior**
Dialogue window is placed in such way that it doesn't get cut or, it's resized to fit available screen state.

**Environment:**
 - OS/Distro: NetBSD-current
 - Output of `leftwm-check`
```
2022-01-18 09:18 > leftwm check
:: LeftWM version: 0.3.0
:: LeftWM git hash: 5e1ccc94d1
:: Loading configuration . . .
    -> Configuration loaded OK 
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK (has XDG_RUNTIME_DIR) 
:: Checking theme . . .
    -> Theme OK 

```

**Additional context**
Not a major issue and probably something you are already aware of.

AethanFoot:
I think the best approach here would be to center the dialog on the workspace when it breaches the edge of its workspace.

0323pin:
@AethanFoot I agree with that. As said, this is a minor improvement, nothing critical. Usually, I have only two clients _per_ tag or, rarely three but, thought I'd just bring it up for discussion.
Thanks for considering it.

cesarFuhr:
@AethanFoot can I fix this?
I checked the code, made some simple changes and tested locally... seems to have solved the problem.

AethanFoot:
@cesarFuhr of course. Open up a pr!

PR #683, opened 2022-01-29
fix: dialogue windows getting cut-off (#668)

cesarFuhr:
This PR fixes the dialog windows getting cut off error by centering them in the workspace if their lower right corner is outside the workspace.

Since I'm starting with Rust, please let me know if there are more "idiomatic" ways of implementing it.
I'm eager to learn, thanks!

image

### DIFF
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -368,8 +368,7 @@ fn set_relative_floating(window: &mut Window, ws: &Workspace, outer: Xyhw) {
         || ws.center_halfed(),
         |mut requested| {
             requested.center_relative(outer, window.border);
-            let (corner_x, corner_y) = requested.lower_right_corner();
-            if ws.contains_point(corner_x, corner_y) {
+            if ws.xyhw.contains_xyhw(&requested) {
                 requested
             } else {
                 ws.center_halfed()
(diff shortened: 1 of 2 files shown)


0323pin:
@cesarFuhr Thanks!

AethanFoot:
Looks good. I would also check the top left corner (would just be the x and y), in case the window overlaps on the top or left sides. This then would nicely fit in as a method for xyhw that checks if another xyhw is within it. So for example in window_handler you would call a method on the workspaces xyhw that checks if it contains the windows xyhw (window.calculated_xyhw), then the method can check if both corners of the window are within the workspace.

VuiMuich:
I am notoriously bad at writing tests myself, but I guess this would be a good thing to test for, right?

May I ask why you put the `#[must_use]`?

AethanFoot:
The #[must_use] is probably a clippy. It says that the return value must be used by the caller, e.g. it is why when a return value isn't used you do `let _ = ...`.

VuiMuich:
I deleted it and not even the pedantic checks triggered :man_shrugging:

AethanFoot:
Hmm strange, it is probably just a copy-pasta. Not sure why clippy decides it is not needed for this but is for center 🤔 .

cesarFuhr:
It is a copy-paste, I am not really familiar with the annotations. Since the method is quite similar to center I thought is was needed.

But I like the idea of writing a `contains` function to check both corners.

VuiMuich:
(Nice, always wanted to try this `Allow edits and access to secrets by maintainers` feature :angel:)

PR #688, opened 2022-01-29
Add `Modifier::is_empty`, use it in `check_mousekey`

vrmiguel:
Simplifies the check in `check_mousekey` considerably


VuiMuich:
Nice catch.
:+1:

PR #689, opened 2022-01-29
Slight refactoring of `default_terminal`

vrmiguel:
Removes the `Vec` allocation and uses the `Iterator`-based `find` which the compiler _should_ optimize better

### DIFF
--- a/leftwm/src/config/mod.rs
+++ b/leftwm/src/config/mod.rs
@@ -189,8 +189,8 @@ pub fn is_program_in_path(program: &str) -> bool {
 /// Returns a terminal to set for the default mod+shift+enter keybind.
 fn default_terminal<'s>() -> &'s str {
     // order from least common to most common.
-    // the thinking is if a machine has a uncommon terminal install it is intentional
-    let terms = vec![
+    // the thinking is if a machine has an uncommon terminal installed, it is intentional
+    let terms = &[
         "alacritty",
         "termite",
         "kitty",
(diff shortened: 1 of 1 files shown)


leftwm/src/config/mod.rs
```diff
@@ -209,12 +209,12 @@ fn default_terminal<'s>() -> &'s str {
...
-            return t;
-        }
-    }
-    "termite" //no terminal found in path, default to a good one
+
+    // If no terminal found in path, default to a good one
+    terms
+        .iter()
+        .find(|terminal| is_program_in_path(terminal))
+        .unwrap_or(&"termite")
```
VuiMuich, line 217: Would we want to use the opportunity and replace the orphaned `termite` with something alive?

mautamu: Sounded like a good idea to me, but apparently some AUR folks are trying to keep a somewhat-maintained version around (https://aur.archlinux.org/packages/termite/). Maybe should just move it down list?

VuiMuich: Nice, didn't know that!
Yeah, moving it down the list makes sense.

VuiMuich: I reconsidered this, and I given our statement that the list is ordered uncommon -> common, it is arready in a good position and it is actually `alacritty` that seems a bit out of place.
But I quess we can save this discussion for another day.

I'll go ahead and merge this. Thanks @vrmiguel!

PR #690, opened 2022-01-30
Reduce xlib calls and other optimisations

AethanFoot:
This reduces the xlib calls by 1: removing calls that don't add anything, 2: focus events are self contained and no longer makes moving/resizing calls, 3: only make the windows move/resize when needed (this mainly involves returning false in command_handler/display_event_handler when the windows don't actually move/resize, not returning true/false depending on whether it succeeded). Maybe more to come.

Thanks.

### DIFF
--- a/leftwm-core/src/display_servers/xlib_display_server/mod.rs
+++ b/leftwm-core/src/display_servers/xlib_display_server/mod.rs
@@ -83,30 +83,14 @@ impl DisplayServer for XlibDisplayServer {
             }
         }
 
-        match self.xw.mode {
-            // Only process minimal events when moving/resizing.
-            Mode::MovingWindow(_)
-            | Mode::ResizingWindow(_)
-            | Mode::ReadyToMove(_)
-            | Mode::ReadyToResize(_) => {
-                let xlib_event = self.xw.get_mask_event();
-                let event = XEvent(&mut self.xw, xlib_event).into();
-                if let Some(e) = event {
-                    log::trace!("DisplayEvent: {:?}", e);
-                    events.push(e);
-                }
-            }
-            Mode::Normal => {
-                let events_in_queue = self.xw.queue_len();
+        let events_in_queue = self.xw.queue_len();
 
-                for _ in 0..events_in_queue {
-                    let xlib_event = self.xw.get_next_event();
-                    let event = XEvent(&mut self.xw, xlib_event).into();
-                    if let Some(e) = event {
-                        log::trace!("DisplayEvent: {:?}", e);
-                        events.push(e);
-                    }
-                }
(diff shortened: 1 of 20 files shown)


leftwm-core/src/display_servers/xlib_display_server/xatom.rs
```diff
@@ -120,143 +119,57 @@ impl XAtom {
     #[allow(clippy::too_many_lines)]
     // TODO: Use a match statement.
```
VuiMuich, line 120: Nice!
I guess the TODO isn't needed any longer then.

VuiMuich:
Awesome work as always, Aethan!

Wondering how much is there to come, bevore exiting draft status :joy:

AethanFoot:
Some mundane tasks are done and other random pieces (most in an effort to debug/improve/fix ClickTo not really successfully, it may be better). Main thing for this pr is checking for regressions (haven't found any in the few days I have been using it).
Thanks!

leftwm-core/src/display_servers/xlib_display_server/mod.rs
```diff
@@ -101,137 +103,33 @@ impl DisplayServer for XlibDisplayServer {
...
-            }
-            DisplayAction::ReloadKeyGrabs(keybinds) => {
-                self.xw.reset_grabs(&keybinds);
-                None
-            }
-            DisplayAction::ConfigureXlibWindow(window) => {
-                self.xw.configure_window(&window);
-                None
-            }
+                previous_window,
```
mautamu, line 128: Thank you for taking care of this!

leftwm-core/src/display_servers/xlib_display_server/xatom.rs
```diff
@@ -117,146 +116,58 @@ impl XAtom {
...
+            a if a == self.NetDesktopViewport => "_NET_DESKTOP_VIEWPORT",
+            a if a == self.NetNumberOfDesktops => "_NET_NUMBER_OF_DESKTOPS",
+            a if a == self.NetCurrentDesktop => "_NET_CURRENT_DESKTOP",
+            a if a == self.NetDesktopNames => "_NET_DESKTOP_NAMES",
+            a if a == self.NetWMDesktop => "_NET_WM_DESKTOP",
+            a if a == self.NetWMStrutPartial => "_NET_WM_STRUT_PARTIAL",
+            a if a == self.NetWMStrut => "_NET_WM_STRUT",
+
+            a if a == self.UTF8String => "UTF8_STRING",
+            _ => "(UNKNOWN)",
```
mautamu, line 169: Much cleaner!

mautamu (approved):
Appears to work, although I'm not sure exactly what to test for.

PR #693, opened 2022-01-31
fix(416): focus empty workspace on click

### LINKED ISSUE #416: Click on empty screen and then spawn windows there

### LINKED ISSUE #426: Feature Request: Set margins conditionally for each workspace or screen
Leftwm currently places windows above my polybar instance on only the second screen. I have addressed this "successfully" in two ways:
1) Define workspaces in `.config/leftwm/config.toml` for each screen with the second screen offset vertically by `y = 10` or whatever number is required.
2) Define margins in `.config/lefwm/themes/my_polybar_theme/theme.toml` with `margin = [20, 8, 8, 8]` or whatever values you like. 

The first is probably the better solution but I suspect would lead to other issues I haven't discovered yet with the actual screen space not matching the representation of that screenspace. I also don't understand why this vertical offset only affects the windows and not the placement of the polybar instance or the background image from `feh`.
The second is easier to reason about but leads to a larger gap above the windows on the first screen, which I don't like.

My request is to have `theme.toml` accept values for `margin` by screen or workspace identifier OR to have `config.toml` accept a `margin` setting within the workspace definitions. I think the latter is preferable and might actually allow for interesting customization of specific workspaces for specific workflows.

kid:
Resolves #416

### DIFF
--- a/leftwm-core/src/display_event.rs
+++ b/leftwm-core/src/display_event.rs
@@ -8,7 +8,7 @@ pub enum DisplayEvent {
     Movement(WindowHandle, i32, i32),
     KeyCombo(ModMask, XKeysym),
     KeyGrabReload, // Reloads keys for when keyboard changes.
-    MouseCombo(ModMask, Button, WindowHandle),
+    MouseCombo(ModMask, Button, WindowHandle, i32, i32),
     WindowCreate(Window, i32, i32),
     WindowChange(WindowChange),
     WindowDestroy(WindowHandle),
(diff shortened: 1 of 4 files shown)


kid:
I'm not sure how to write tests for this one :thinking: 
Do we have any tests that uses the mouse that I could look into?

VuiMuich:
I am not aware of any mouse related tests and a quick rip-grep didn't bring up anything either.
But you find some inspiration  here:
```
leftwm-core/src/display_servers/xlib_display_server/mod.rs
26:mod xcursor;
118:                    let _ = self.xw.move_cursor_to_window(win);
123:                let _ = self.xw.move_cursor_to_point(point);
201:                let point = self.xw.get_cursor_point().ok()?;
253:    /// Creates a verify focus event for the cursors current window.
255:        let handle = self.xw.get_cursor_window().ok()?;
```
```
leftwm-core/src/display_servers/xlib_display_server/xwrap/window.rs
87:        let cursor = self.get_cursor_point().unwrap_or_default();
88:        Some(DisplayEvent::WindowCreate(w, cursor.0, cursor.1))
145:                    let _ = self.move_cursor_to_window(handle);
```

PR #694, opened 2022-02-03
FocusWindow command

Melesar:
This command focuses a window with the given `WM_CLASS`. 

- If the window is not present, it does nothing
- If the window is visible in any workspace, it will gain focus
- If the window is not visible, the current workspace will switch to a tag with the window and give it focus
- If the window is hidden in the `Monocle` or `MainAndDeck` layout, it will scroll the layout so that the window becomes visible and focused
- If the window is already focused, it will focus the previousely focused one, following the same rules

Example configuration:

```toml
[[keybind]]
command = "FocusWindow"
value = "brave-browser" # The first or the second entry of WM_CLASS attribute
modifier = ["modkey"]
key = "b"
```
Another added possibility is to focus a visible window by its index on the screen. Example configuration:

```toml
[[keybind]]
command = "FocusWindow"
value = "1" # Will focus the first window on the screen
modifier = ["modkey", "Control"]
key = "1"
```
Indices are not displayed in any way yet. This feature is WIP for now and will not be documented

Issue #275, opened 2021-04-06
Intermittent: Polybar overwrites tags area with code from template.liquid


bmclean2:
Seemingly randomly, and more frequently since the last few commits, polybar will overwrite the tags module with code directly from template.liquid. So I'll get the end of a line like this `%{A1:$SCRIPTPATH/change_to_tag {{workspace.index}} {{tag.index}}:}` printed to the screen on the bar, but with the variables completed. What I see is `H/change_to_tag 0 6:}` overwriting tags 1 through 6. I originally wondered if there was a typo in the file, but I can't spot one. It happens intermittently, but enough to be an annoyance. It is caused by both key bindings and mouse clicks, and doesn't seem to be related to the tag index. Problem exists on community themes as well.

mautamu:
I’ve witnessed this as well. My suspicion is that it might be an unescaped parenthesis in the name of the window, as it seems to happen most often on tags with Firefox, Brave, or Thunderbird open.

bmclean2:
Yes, I usually have both Firefox and Brave open, so that would explain things. The increasing frequency could just be a coincidence.

VuiMuich:
As mentioned in #207 I observed something similar, but I believe that it is actually a different issue: polybar is faster with launching then leftwm-state can produce an parsable output and prints the returned error-message.
Also had this during crashes of leftwm-worker, for obvious reasons.

Maybe we should look for a way to make polybar to fallback to some default, if leftwm-state fails to return the correct output.

A possible solution should then be included in the troubleshooting and 'theming polybar' sections of readme and wiki.

VentGrey:
I commented this on the official discord when the `_NET_CURREN_DESKTOP` bug happened. An easy way to trigger it is to copy the polybar example and "spam" the workspace change keybind (Hold <kbd>Super</kbd> + <kbd>2</kbd> for example) and you'll see the same strange formatting referred in #575.

PR #697, opened 2022-02-07
Remove Chrono Dependency

mautamu:
Partially resolves RUSTSEC2020-0159.
 - Removes chrono 
 - Bumps time. 

 !!!DOES NOT REMOVE TIME 0.1.43 FROM ENTIRE DEPENDENCY TREE!!!

Cheers,
Mau

Tagging #594,

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,18 +13,18 @@ dependencies = [
 
 [[package]]
 name = "ansi_term"
-version = "0.11.0"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
 dependencies = [
  "winapi",
 ]
 
 [[package]]
 name = "anyhow"
-version = "1.0.48"
+version = "1.0.53"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "62e1f47f7dc0422027a4e370dd4548d4d66b26782e513e98dca1e689e058a80e"
+checksum = "94a45b455c14666b85fc40a019e8ab9eb75e3a124e05494f5397122bc9eb06e0"
 
 [[package]]
 name = "anymap2"
@@ -51,9 +51,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
+checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "bitflags"
@@ -70,7 +70,16 @@ dependencies = [
  "block-padding",
  "byte-tools",
  "byteorder",
- "generic-array",
+ "generic-array 0.12.4",
+]
+
+[[package]]
+name = "block-buffer"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4152116fd6e9dadb291ae18fc1ec3575ed6d84c29642d97890f4b4a3417297e4"
+dependencies = [
+ "generic-array 0.14.5",
 ]
 
 [[package]]
@@ -127,9 +136,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "2.33.3"
+version = "2.34.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "37e58ac78573c40708d45522f0d80fa2f01cc4f9b4e2bf749807255454312002"
+checksum = "a0610544180c38b88101fecf2dd634b174a62eef6946f84dfc6a7127512b381c"
 dependencies = [
  "ansi_term",
  "atty",
@@ -141,10 +150,19 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam-channel"
-version = "0.5.1"
+name = "cpufeatures"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "06ed27e177f16d65f0f0c22a213e17c696ace5dd64b14258b52f9417ccb52db4"
+checksum = "95059428f66df56b63431fdb4e1947ed2190586af5c5a8a8b71122bdf5a7f469"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "crossbeam-channel"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e54ea8bc3fb1ee042f5aace6e3c6e025d3874866da222930f70ce62aceba0bfa"
 dependencies = [
  "cfg-if",
  "crossbeam-utils",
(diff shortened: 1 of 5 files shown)


mautamu:
I haven't been able to locally check this for accuracy. @0323pin Does this work for you?

(MSRV check blocked on https://github.com/leftwm/leftwm/pull/694).

0323pin:
@mautamu It was sleep time where I am. Thanks for the quick work.
I need to get a few things done first (work) but, I'll kick a patched build as soon as I possibly can and report back.
Thanks you!

VuiMuich:
May I ask why you are using `UTC` in the `log` and `local` in `leftwm-state`?

Btw. nice one with the `if let Ok(..)` I only found `.expect()` to unwrap it in my commit on #508 yesterday.

0323pin:
@mautamu It builds fine but I get "TIME ERROR!" on the bar when having `{{ localtime }}` in `template.liquid`.
Should I be calling something else in `template.liquid`? I've looked at the changes but, I'm failing to see

If I use `{{ "now" | date: "%Y-%m-%d %H:%M" }}` time shows up in UTC as expected.

VuiMuich:
> @mautamu It builds fine but I get "TIME ERROR!" on the bar when having `{{ localtime }}` in `template.liquid`. Should I be calling something else in `template.liquid`? I've looked at the changes but, I'm failing to see how to call the new variable.
> 
> If I use `{{ "now" | date: "%Y-%m-%d %H:%M" }}` time shows up in UTC as expected.

`{{ "now" | date: "%Y-%m-%d %H:%M" }}` is 'a function' directly from the `liquid-core` crate and is only able to return UTC

This is weird, the following code works fine in the `fn setup_logfile()` but returns an error just as mautamu's code:
```rust
    let date = time::OffsetDateTime::now_local().expect("Localtime: ");
    return date
        .format(time::macros::format_description!(
            "[year]-[month]-[day] [hour]:[minute]"
        ))
        .expect("Formated localtime: ")
        .to_string();
```

0323pin:
> `{{ "now" | date: "%Y-%m-%d %H:%M" }}` is 'a function' directly from the liquid-core crate and is only able to return UTC

Yes, I know. You have linked the bug earlier.

So, you're seeing it also?

VuiMuich:
I built this little test project:
`src/main.rs`
```rust
fn get_localtime() -> String {
    // let now = chrono::Local::now();
    // now.format("%m/%d/%Y %l:%M %p").to_string()

    // log::info!("The `{{ localtime }}` feature is deprecated, please use another source like this little program from the epitaph theme https://github.com/VentGrey/Epitaph/blob/90a4a799001d8896beff5226143dddaf34e9b3f4/scripts/lemonbar/time.c");
    // "DEPRECATED_LOCALTIME".to_string()

    let now = time::OffsetDateTime::now_local().expect("My Localtime");
    return format!(
        "{}",
        now.format(time::macros::format_description!(
            "[year]-[month]-[day] [hour]:[minute]"
        ))
        .expect("My Formated localtime")
    );

    // if let Ok(now) = time::OffsetDateTime::now_local() {
    //     if let Ok(ret) = now.format(time::macros::format_description!(
    //         "[month]/[day]/[year] [hour]:[minute]:[second]"
    //     )) {
    //         return ret;
    //     }
    // }
    // String::from("TIME ERROR!")
}
```
`Cargo.toml`
```toml
[package]
name = "timetest"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
time = {version = "0.3.7", features = ["formatting", "local-offset", "macros"]}
```
And it runs as expected:
```shell
timetest >>> cargo r
   Compiling timetest v0.1.0 (/home/vuimuich/timetest)
    Finished dev [unoptimized + debuginfo] target(s) in 0.53s
     Running `target/debug/timetest`
The one: 202202071212
and the other: 2022-02-07 12:12
```
So something seems to be weird in the context of `leftwm-state`. Maybe it is from conflcting `time` crates, as `liguid-core` uses `0.1.44` and we want `0.3.7`...

0323pin:
> So something seems to be weird in the context of leftwm-state. Maybe it is from conflcting time crates, as liguid-core uses 0.1.44 and we want 0.3.7...

Indeed, it's strange. The code runs as expected when used stand-alone but fails when inside `leftwm-state`.
Maybe it's worthy to try to force `liquid-core` to update `time`, if it turns out to be conflicting crate versions?

VuiMuich:
But actually that would also be weird, as `liquid` only gets used locally in functions and I thought that should limit its life span and 'our' `time` is also limited to the `get_localtime` function.
And why is my variant calling `get_localtime` even with `{{ "now" | date: "%Y-%m-%d %H:%M" }}`?

0323pin:
> And why is my variant calling get_localtime even with {{ "now" | date: "%Y-%m-%d %H:%M" }}?

You're right, that's weird. This should be a liquid only thing and shouldn't be impacted. 😮

mautamu:
> May I ask why you are using `UTC` in the `log` and `local` in `leftwm-state`?
> 

That’s just the fallback, since now_local returns Result and now_utc returns the underlying object without a result.


Also that’s some extremely peculiar behavior  we’ve found… 🤨. I’ll try to take a look after work if y’all haven’t figured it out by then.

Also just saw the work on the logger default 🤦. Apologies I didn’t see it earlier. Will try to integrate proposed changes later too.

VuiMuich:
> That’s just the fallback, since now_local returns Result and now_utc returns the underlying object without a result.

:see_no_evil: I totally missed that it is in the `else` arm, now that absolutely makes sense as a fallback. Sorry for my blindness :bow:

> Also that’s some extremely peculiar behavior we’ve found… raised_eyebrow. I’ll try to take a look after work if y’all haven’t figured it out by then.
> 
> Also just saw the work on the logger default facepalm. Apologies I didn’t see it earlier. Will try to integrate proposed changes later too.

I tried to backtrace it but I am bad at this and don't understand the output :roll_eyes: 
I asked for an opinion on the rust discord and they think it is not possible to compile two versions of a crate into one binary, which is a pitty, as `chrono` is used as well in `liquid-core` as mentioned and in `slog-term`.
Right now I am tumbling down that rabbit whole:
- `liquid` are aware of the vulnerability and appear generally open to do the move to `time` – no idea though how big that effort would be for them
- `slog-term` have no (relevant) mention of `chrono` or `time` or the vulnerability, but as this is the only crate of the `slog` suite chances are the changes to do the switch are manageable
- `chrono` appears to be almost completely stalled (last commit in October 2021) at this point

0323pin:
> I asked for an opinion on the rust discord and they think it is not possible to compile two versions of a crate into one binary

:(

> solved this by `slog-term v2.8.1`

Nice, they just need to push a release now.

Which, leaves `liquid`.

mautamu:
Oh actually I think I have an idea for how to do it, but it’ll take a bit of effort. 

The tldr is this:

https://stackoverflow.com/questions/50999616/how-to-idiomatically-alias-a-crate-in-rust-2018

VuiMuich:
Another idea: as `leftwm-state` operates on updates on the state we could store the time in the state file and update it from the `worker`, so the two `time` crates end up in different binaries.

0323pin:
Hmm... sounds like a burden to maintain. Much cleaner if `liquid` and `slog-term` could keep `time` in sync.

Thoughts?!

VuiMuich:
:thinking: so 

> Oh actually I think I have an idea for how to do it, but it’ll take a bit of effort.
> 
> The tldr is this:
> 
> https://stackoverflow.com/questions/50999616/how-to-idiomatically-alias-a-crate-in-rust-2018

I tried this in the little `timetest` from above, but I believe it breaks macros :sob:

mautamu:
Ah there’s a way to do it without macros but will make the code less clean 😢.

mautamu:
I've gotten a bit further, but my solution can't obtain the current offset. We could pass through an offset? Not super familiar with liquid, but maybe localtime(-5)? 

Thoughts? @VuiMuich / @0323pin

0323pin:
> I've gotten a bit further, but my solution can't obtain the current offset. We could pass through an offset?

Do you mean it passes UTC?
Because `liquid` can do that on it's own. I also thought about adding one to the hour but couldn't find how to do that in `liquid`.
Unfortunately, `liquid` doesn't seem to do `date -d '+1 hour'

mautamu:
I think this is what's causing #275

0323pin:
> I'm now in favor of dropping chrono and time from leftwm-state entirely.

Looks like `liquid` is closing in on a fix to drop `chrono`, https://github.com/cobalt-org/liquid-rust/pull/460

VuiMuich:
> > I'm now in favor of dropping chrono and time from leftwm-state entirely.
> 
> Looks like `liquid` is closing in on a fix to drop `chrono`, [cobalt-org/liquid-rust#460](https://github.com/cobalt-org/liquid-rust/pull/460)

If they get the crate bumped as soon as they merged this I guess we could wait and bump `liquid` in our `Cargo.toml` as well.

PR #704, opened 2022-02-13
Helpers and cleanup

AethanFoot:
Fix focus_new_windows, and do some new clippys and cleanup with some helper functions.


mautamu (approved):
Lots of handy things in here to reduce verbosity! Looks good to me. I've been testing it for a while now and no major issues so far.

PR #709, opened 2022-02-17
Add configuration option to move new windows to top

lgmys:
This PR allows the user to explicitly say in the `config.toml` that
new windows should be moved to the top.

### DIFF
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -338,11 +338,14 @@ fn insert_window(state: &mut State, window: &mut Window, layout: Layout) {
         }
     }
 
-    // If a window is a dialog, splash, or scractchpad we want it to be at the top.
+    // If a window is a dialog, splash, or scractchpad,
+    // or move_new_windows_to_top is set in the config,
+    // we want it to be at the top.
     if window.r#type == WindowType::Dialog
         || window.r#type == WindowType::Splash
         || window.r#type == WindowType::Utility
         || is_scratchpad(state, window)
+        || state.insert_new_windows_on_top
     {
         state.windows.insert(0, window.clone());
         return;
--- a/leftwm-core/src/config/mod.rs
+++ b/leftwm-core/src/config/mod.rs
@@ -31,6 +31,8 @@ pub trait Config {
 
     fn focus_new_windows(&self) -> bool;
 
+    fn insert_new_windows_on_top(&self) -> bool;
+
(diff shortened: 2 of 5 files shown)


leftwm-core/src/handlers/window_handler.rs
```diff
@@ -56,6 +56,10 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
             self.state.focus_window(&window.handle);
         }
 
+        if self.config.move_new_windows_to_top() {
+            self.state.move_to_top(&window.handle);
+        }
+
```
AethanFoot, line 62: Rather than handling this here I would place it in insert_window. Also calling move_to_top may cause the tiled window to appear above floating windows, instead I would insert the window first in the workspace before the sorting happens. Placing your check in this block of insert_window should work:
```
        if matches!(layout, Layout::Monocle | Layout::MainAndDeck) || new_option { // Add check for new option
            // Extract the current windows on the same workspace.
            let mut to_reorder = helpers::vec_extract(&mut state.windows, for_active_workspace);
            if layout == Layout::Monocle || to_reorder.is_empty() || new_option { // Add check for new option
                // When in monocle we want the new window to be fullscreen if a window was
                // fullscreen.
                if was_fullscreen {
                    let act = DisplayAction::SetState(
                        window.handle,
                        !window.is_fullscreen(),
                        WindowState::Fullscreen,
                    );
                    state.actions.push_back(act);
                }
                // Place the window above the other windows on the workspace.
                to_reorder.insert(0, window.clone());
            } else {
```

AethanFoot (commented):
Thanks for adding this!

leftwm-core/src/handlers/window_handler.rs
lgmys: Cool I will improve it later today, thanks!

lgmys: Unfortunately, `insert_window` is not aware of config, so the config value I introduced would have to passed in somehow, via an argument maybe?

Also, shouldn't it be placed just above this?:
```
    // Past special cases we just push the winodw to the bottom.
    state.windows.push(window.clone());
```

lgmys: Insertion would be done like this `state.windows.insert(0, window.clone());` I imagine

AethanFoot:
Hmmm, this may be encompassed by #710.

lgmys:
Closing this in favor of #710

Issue #358, opened 2021-05-26
[Feature] New Windows as Master


jooooscha:
Is there a way to make a new window automatically the master?
If I missed something, please point me there :)

Thanks :)

AethanFoot:
Currently not, however this would easy enough to implement a switch for this.
For future reference I would say this could be an option per tag/layout once greater configuration is implemented.

VuiMuich:
> For future reference I would say this could be an option per tag/layout once greater configuration is implemented.

Agreed, having this option at least configurable per layout is quite important as there probably might be very different opinions on the preferred behavior.

nblxxx:
Just installed Leftwm and love it already! but this is the only confusing thing about it for me, coming from xmonad, where new windows spawn in the previously focused tile, so if you're in master they go there (so not always, like requested by @jooooscha here)

kid:
I think one feature that would be nice would be the option to insert the window relative to the current window, and not just the top/bottom of the stack

VuiMuich:
These are some great ideas.
So basically the the switch could be a `new_window_mode` with options `Top`, `Bottom`, `InsertBeforeCurrent`, `InsertAfterCurrent` and `Replace`.

nblxxx:
That would be perfect!

> So basically the the switch could be a `new_window_mode` with options `Top`, `Bottom`, `InsertBeforeCurrent`, `InsertAfterCurrent` and `Replace`.

kid:
Not sure what would be the behavior of `Replace` though, what did you have in mind @VuiMuich ?

VuiMuich:
I guess in most scenarios it would end up being the same as `InsertBeforeCurrent`. I thought there might be a corner case where they might be a bit different, but not sure anymore. :shrug:

VuiMuich:
Implemented in #710

PR #710, opened 2022-02-19
Configurable new window behavior

kid:
This implements #358 and allows configurable behavior for new windows.

### DIFF
--- a/leftwm-core/src/config/insert_behavior.rs
+++ b/leftwm-core/src/config/insert_behavior.rs
@@ -1,0 +1,9 @@
+use serde::{Serialize, Deserialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq)]
+pub enum InsertBehavior {
+    Top,
+    Bottom,
+    BeforeCurrent,
+    AfterCurrent
+}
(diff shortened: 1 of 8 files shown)


kid:
Probably needs a better name than `insert_behavior`, maybe `new_window_mode`?

kid:
Also, I'm not sure what to do with Monocle and `MainAndDeck`, as I'm not sure why they are special cases in `insert_window` (I can guess why for `Monocle`, but `MainAndDeck` is a bit of a mystery :thinking: )

kid:
Marking as ready now that it has tests, but I'm just not sure about the naming

VuiMuich:
> Also, don't know what to do with Monocle and `MainAndDeck`, as I'm not sure why they are special cases in `insert_window` (I can guess why for `Monocle`, but `MainAndDeck` is a bit of a mystery thinking )

Will get into this PR later, but want to leave this comment right now:
The deck behaves in most cases like the monocle, but further special, as the top window remains always separated from it.

I hope this makes sense to you.

kid:
does it makes sense to handle at least `InsertBehavior::Top` (and maybe InsertBehavior::BeforeCurrent` ?) as well for `MainAndDeck`

mautamu:
Was quietly force-pushed, should be ready to go now. Thanks for the quick update @kid! Probably wise to go after #594 but before #638

PR #713, opened 2022-02-19
Allow disabling swap for MoveWindowTop

cecton:
# Changes
- command `MoveWindowTop` in core allows disabling/enabling automatic swap
- changed `FocusWindowTop` to use struct variant instead of tuple variant (for clarity sake)
- turned `move_focus_common_vars` into a macro to allow flexibly pass any kind and number of parameters
- command in config toml `MoveWindowTop` now accepts a boolean value to disable/enable the swapping (enabled by default to avoid breaking compatibility)
- `MoveWindowTop` can be passed (optional) with a boolean argument in the command pipe. Defaults to swap: true (to avoid breaking change... though I think `false` would be better there imo)

No breaking change has been introduced (hopefully :grin:)

### DIFF
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -28,6 +28,21 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
     }
 }
 
+macro_rules! move_focus_common_vars {
+    ($func:ident ($state:expr $(, $arg:expr )* $(,)? )) => {{
+        let handle = $state.focus_manager.window(&$state.windows)?.handle;
+        let tag_id = $state.focus_manager.tag(0)?;
+        let tag = $state.tags.get(tag_id)?;
+        let (tags, layout) = (vec![tag_id], Some(tag.layout));
+
+        let for_active_workspace =
+            |x: &Window| -> bool { helpers::intersect(&tags, &x.tags) && !x.is_unmanaged() };
+
+        let to_reorder = helpers::vec_extract(&mut $state.windows, for_active_workspace);
+        $func($state, handle, layout, to_reorder, $($arg),*)
+    }};
+}
+
 fn process_internal<C: Config, SERVER: DisplayServer>(
     manager: &mut Manager<C, SERVER>,
     command: &Command,
@@ -44,9 +59,9 @@ fn process_internal<C: Config, SERVER: DisplayServer>(
         Command::SendWindowToTag { window, tag } => move_to_tag(*window, *tag, manager),
         Command::MoveWindowToNextWorkspace => move_window_to_workspace_change(manager, 1),
         Command::MoveWindowToPreviousWorkspace => move_window_to_workspace_change(manager, -1),
-        Command::MoveWindowUp => move_focus_common_vars(move_window_change, state, -1),
-        Command::MoveWindowDown => move_focus_common_vars(move_window_change, state, 1),
-        Command::MoveWindowTop => move_focus_common_vars(move_window_top, state, 0),
+        Command::MoveWindowUp => move_focus_common_vars!(move_window_change(state, -1)),
+        Command::MoveWindowDown => move_focus_common_vars!(move_window_change(state, 1)),
+        Command::MoveWindowTop { swap } => move_focus_common_vars!(move_window_top(state, *swap)),
 
         Command::GoToTag { tag, swap } => goto_tag(state, *tag, *swap),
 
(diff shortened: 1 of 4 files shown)


leftwm-core/src/handlers/command_handler.rs
```diff
@@ -557,28 +572,12 @@ fn toggle_floating(state: &mut State) -> Option<bool> {
...
-    func(state, val, handle, layout, to_reorder)
-}
-
 fn move_window_change(
     state: &mut State,
-    val: i32,
     mut handle: WindowHandle,
     layout: Option<Layout>,
     mut to_reorder: Vec<Window>,
+    val: i32,
```
VuiMuich, line 580: Just curious: Is this reordering meaningful and I miss the implications, or is it simply aesthetic?

VuiMuich:
Once again a very nice PR of yours! :clap:

leftwm-core/src/handlers/command_handler.rs
cecton: The ordering matters.

Some of the arguments are created by the wrapper function/macro `move_focus_common_vars`. With the function, the argument `val` was arbitrary and sometimes not used. So that one goes to `$arg` in the macro (which is multiple! very cool)

But the leading argument must be common for all the functions. That why I moved `val` to the end.

An alternative to the macro would be to use a generic type argument... tbh I think macros are nicer in this case because it allows having multiple custom arguments.

I'm also not super happy with this wrapper function (`move_focus_common_vars`)... but I'm not sure how to change it and it's not really my priority so... :woman_shrugging: good enough for now I guess.

VuiMuich: Ohh, the macro brought that, thanks for the elaborate explanation!

Iirc the `..._common_vars` were part of some performance optimisations with significant impact by @AethanFoot

Same as with macros mechanics these are hard for me to grasp with my layman greenhorn knowledge though :joy:

cecton: Performance? Oh that explains it!

Well now it's performant and flexible xD

cecton: @VuiMuich you can see in this code how the arguments are dynamically added depending on how the macro is called:

```rust
$func($state, handle, layout, to_reorder, $($arg),*)
```

- `$func` is your inner function.
- `$state` the state provided in argument of the macro
- `handle`, `layout`, `to_recorder` are computed in that macro
- `$( ... ),*` is like `$( ... )*`: expanding as many `$arg` as provided to the macro call, but the `,` indicates they will be joined with a `,` in between

This would work the same:

```rust
$func($state, handle, layout, to_reorder $(, $arg)*)
```

(It's slightly better because the `,` after `to_reorder` is only inserted if there are arguments. But that just compile-time cost so it really does not matter at all.)

VuiMuich:
I finally had some time to test this.
One thing that strikes me odd is that the `value` for `MoveWindowTop` has to be put in as `String` in `config.toml` and not as `boolean`, but I guess that's the toml parser to blame, right?

cecton:
Yup that's something we need to fix in the new/future config file.

For now, afaik, all commands parse the "value" from a string.

March 5, 2022 9:31:54 AM CET VuiMuich ***@***.***> wrote:

I finally had some time to test this.
One thing that strikes me odd is that the value for MoveWindowTop has to be put in as String in config.toml and not as boolean, but I guess that's the toml parser to blame, right?
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

VuiMuich:
@cecton I berieve I found some little config breaking thing:
When my config is (using compact syntax with `keybind = [ ... ]`):
```toml
{ command = "MoveWindowTop", modifier = ["modkey"], key = "Return" },
```
the parser breaks, I have to at least write it with an empty `value` field like this:
```toml
{ command = "MoveWindowTop", value = "", modifier = ["modkey"], key = "Return" },
```
If you can reproduce on your main as well, would you have time to send a patch PR?
If you don't experience this, it is probably related to #682 and we can fix it before merging this.

VuiMuich:
Interesting enough `leftwm-check` wont complain, but in `jouranlctl` I get this logged:
```
Mar 16 09:07:01 manjobook leftwm-worker[203559]: An error occurred while parsing the command: provided string was not `true` or `false`
```

cecton:
eww this code is terrible.... I'd better refactor it so we won't make the same mistake again

VuiMuich:
> eww this code is terrible.... I'd better refactor it so we won't make the same mistake again

Thanks.

Tbh I was a bit puzzled by it during review, but though it was just my poor knowledge of fancy syntax tricks :angel:

cecton:
No it's just I don't refactor everything at every PR so there are a few copy-paste from other parts of the code that are supposed to work and do the same job but if that part had a bug, I copy-pasted bugs with it.

PR #714, opened 2022-02-20
Fixed basic_polybar theme for multi-monitor setup

clay53:
Most advanced bash script I've written so far so if there are better ways of doing things, I'm fine with changing them.

I used this theme setup as the basis for mine. Worked fine on my single-monitor laptop but when I brought it to my desktop that has 3 monitors with the middle one being a different size, polybar was not spawned with the correct monitor set or in the right positions.

I changed the pre-loop section to instead of getting a single monitor, to get all the monitors by removing the grep and tac (which caused monitors to be in reverse order compared to `leftwm-state -q -n -t $SCRIPTPATH/sizes.liquid`). This is so I can set the monitor variable that's passed to the polybar command to the correct monitor (previously, it only got the first monitor and sent that to each). This fixes the highlighted tag being wrong. I got the correct monitor in-loop by using sed which was already a dependent of this script. However, sed expects it to be indexed at one so I created a new variable for that. I then set the 0-indexed variable at the end to that variable instead of recalculating index+1 (don't know if that's actually faster though). I also changed offsetx=$x to offset=$x since offsetx didn't work for my middle monitor which is index0 and a different size than the others. As far as I know, these changes are only an improvement and shouldn't break old setups.


VuiMuich:
Welcome adventurer!

This is looking really nice!
I have one little nitpick: maybe you could rename `index1` to `indextemp` or something, just to be more clear the `1` has nothing to do with the iteration. Of course this becomes clear when you read the script (and I have to admit I was slightly confused at first when reading the diff, with the blank code it was already better), but this might be a first contact with shell scripting for someone so, we don't have to make it harder as it has to be.

Also would you please ad a `README.md` with the explanation you added in the PR body as documentation.

Thanks!

clay53:
> Also would you please ad a README.md with the explanation you added in the PR body as documentation.

@VuiMuich Where would you like that README.md? Is there an example for something like that?

VuiMuich:
Just in the `basic_polybar` theme directory.
I believe the `basic_eww` has one for reference.

clay53:
> Just in the `basic_polybar` theme directory.
> I believe the `basic_eww` has one for reference. 

Is it just to explain how it works/how to use it or to just explain my changes?

VuiMuich:
> > Just in the `basic_polybar` theme directory.
> > I believe the `basic_eww` has one for reference.
> 
> Is it just to explain how it works/how to use it or to just explain my changes?

Sorry for my delayed response.
Well, I think basically what was in the last paragraph of your opening post. Sure some general doc on all the details would be nice to have eventually, but this has not to happen in this PR, except you are motivated, then go for it :joy:

clay53:
@VuiMuich Alright, I just added the last paragraph into a README. Might makes docs later but not rn.

PR #728, opened 2022-03-07
Fix spacing and else statements on liquid templates

rijkvp:
- The polybar template included three spaces instead of two for 'busy' tags, which caused an inconsistent layout.
- The else statements on all bar templates had a `tag.visible` condition (probably a copy/paste error).


mautamu (approved):
Looks good, thanks for catching this!

PR #737, opened 2022-03-17
CloseAllOtherWindows command

Areizen:
* Initial implementation of CloseAllOtherWindows command
* Given a window, close all other windows in the workspace

### DIFF
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -698,7 +698,10 @@ fn close_all_other_windows(state: &mut State) -> Option<bool> {
     let current_workspace = state.focus_manager.workspace(&state.workspaces);
 
     for window in &state.windows {
-        if window.handle.ne(&current_window?) && current_workspace?.is_displaying(window) {
+        if window.handle.ne(&current_window?)
+            && current_workspace?.is_displaying(window)
+            && window.r#type.ne(&WindowType::Normal)
+        {
             let act = DisplayAction::KillWindow(window.handle);
             state.actions.push_back(act);
         }
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -118,7 +118,8 @@ fn process_internal<C: Config, SERVER: DisplayServer>(
         Command::SetMarginMultiplier(multiplier) => set_margin_multiplier(state, *multiplier),
         Command::SendWorkspaceToTag(ws_index, tag_index) => {
             Some(send_workspace_to_tag(state, *ws_index, *tag_index))
-        }
+        },
+        Command::CloseAllOtherWindows => close_all_other_windows(state),
         Command::Other(cmd) => Some(C::command_handler(cmd, manager)),
     }
 }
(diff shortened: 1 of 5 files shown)


leftwm-core/src/handlers/command_handler.rs
```diff
@@ -691,6 +692,24 @@ fn focus_window_top(state: &mut State, swap: bool) -> Option<bool> {
...
+fn close_all_other_windows(state: &mut State) -> Option<bool> {
+    let current_window: Option<WindowHandle> = state.focus_manager.window(&state.windows).map(|w| w.handle);
+    let current_workspace = state.focus_manager.workspace(&state.workspaces);
+
+    if current_window.is_none() {
+        return Some(false);
+    }
+
+    for window in &state.windows {
+        if window.handle.ne(&current_window.unwrap()) && current_workspace.unwrap().is_displaying(&window) {
```
mautamu, line 704: change the &window to window. 
```
warning: this expression creates a reference which is immediately dereferenced by the compiler
   --> leftwm-core/src/handlers/command_handler.rs:706:57
    |
706 |             && current_workspace.unwrap().is_displaying(&window)
    |                                                         ^^^^^^^ help: change this to: `window`
    |
    = note: `#[warn(clippy::needless_borrow)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
```

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -691,6 +692,24 @@ fn focus_window_top(state: &mut State, swap: bool) -> Option<bool> {
     None
 }
 
+fn close_all_other_windows(state: &mut State) -> Option<bool> {
+    let current_window: Option<WindowHandle> = state.focus_manager.window(&state.windows).map(|w| w.handle);
+    let current_workspace = state.focus_manager.workspace(&state.workspaces);
+
+    if current_window.is_none() {
```
mautamu, line 699: Since we have an Option<bool> return type, we might consider just coercing this to return None like so:

```rs
fn close_all_other_windows(state: &mut State) -> Option<bool> {
    let current_window: Option<WindowHandle> =
        state.focus_manager.window(&state.windows).map(|w| w.handle);
    let current_workspace = state.focus_manager.workspace(&state.workspaces);

    for window in &state.windows {
        if window.handle.ne(&current_window?)
            && current_workspace?.is_displaying(&window)
        {
            let act = DisplayAction::KillWindow(window.handle);
            state.actions.push_back(act);
        }
    }
    Some(true)
}
```

This way we don't panic if something weird happens, and we can be a bit less verbose.

mautamu (commented):
This looks pretty good, just a couple of nitpicks. I think that the errors from GH are spurious. See above for recommendations code-wise.

I think it might be nice if there were a way to prevent polybar et al. from being killed, but I suspect that is outside the scope of this PR. Whenever this is merged, we should remember to add documentation to the Wiki.

Mau

VuiMuich:
> I think it might be nice if there were a way to prevent polybar et al. from being killed, but I suspect that is outside the scope of this PR. Whenever this is merged, we should remember to add documentation to the Wiki.

I think for the initial implementation we could simply apply to windows of type `Normal` only. Assuming most unwanted windows of other type like `Splash`, `Util` or `Dialog` etc. have a parent that gets killed and takes them down.
Later we could add an optional list of window types as an value (keybind) or argument (external command)  to specifically include types like `Dock` or `Desktop`.

If you would like to add the extended behaviour immediately I would like to mark this PR as `draft` for now.

PR #738, opened 2022-03-17
Add GoToPreviousTag feature to unconditionally go to last visited tag

sagittarius-a:
Similarly to GoToTag feature, I wanted a way to unconditionally go to
previous tag, without having to press the key for the current tag.

No default keybindings have been assigned for this feature.

I did not find an up to date list of available commands so I did not
know where to document this new command.

Tested in Xephyr. `cargo test` passes.

### DIFF
--- a/leftwm-core/src/command.rs
+++ b/leftwm-core/src/command.rs
@@ -18,7 +18,7 @@ pub enum Command {
         tag: TagId,
         swap: bool,
     },
-    GoToPreviousTag,
+    ReturnToLastTag,
     FloatingToTile,
     TileToFloating,
     ToggleFloating,
--- a/leftwm-core/src/command.rs
+++ b/leftwm-core/src/command.rs
@@ -18,6 +18,7 @@ pub enum Command {
         tag: TagId,
         swap: bool,
     },
+    GoToPreviousTag,
     FloatingToTile,
     TileToFloating,
     ToggleFloating,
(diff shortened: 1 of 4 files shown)


mautamu:
I'm rather hesitant about the moniker given the new command. The overlapping meanings of "Previous" in `FocusPreviousTag` and `GoToPreviousTag` might cause confusion, in my opinion. Potential recommendations for replacements: `GoToLastTag`, `ReturnToLastTag`, `ReturnToPreviousTag`. Thoughts?

Because this is both an external and internal command, after it has been merged into main, it should be documented in the wiki [here](https://github.com/leftwm/leftwm/wiki/Config) and [here](https://github.com/leftwm/leftwm/wiki/External-Commands) with the appropriate "NOTE: This is only available in LeftWM >=0.3.0." message.

VuiMuich:
I was thinking the same as Mau with my first idea being like `SwitchBackToPreviousTag`, but it's quite long and `ReturnToLastTag` is my favorite suggestion.

sagittarius-a:
I do agree this is quite ambiguous.

I solely chose this name to based on the `CHANGELOG` entry referencing this feature with:

```
GotoTag sends screen to previously viewed tag
```

I like the `ReturnToLastTag` so if you're okay with this I'll commit the changes with this name.

Regarding documentation, maybe we could link this 2 specific locations in the `CONTRIBUTING.md` file ?

VuiMuich:
Thanks for this addition, would you please add the info to the wiki sections.

A proposal for the `CONTRIBUTING.md` would be very welcome for sure.

PR #740, opened 2022-03-20
Feature(leftwm-state): Attempt to reintroduce `localtime` as a feature.

mautamu:
This attempts to add localtime back to leftwm-state. It's still yielding a very high error rate for me. Any ideas how to improve this?

### DIFF
--- a/leftwm/Cargo.toml
+++ b/leftwm/Cargo.toml
@@ -45,6 +45,7 @@ tempfile = "3.2.0"
 [features]
 default = ["slog-term"]
 journald = ["slog-journald", "slow-dm-fix"]
+localtime = []
 
 # Sleep on restart
 slow-dm-fix = []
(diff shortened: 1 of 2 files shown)


VuiMuich:
Have you had a look at the `time-fmt` crate, or do you suspect it to be breaking anywhere else besides parsing the format?

mautamu:
It's breaking when it tries to pull the local time, which is I think the same issue we were running into previously. Formatting works, and it works most of the time when you use UTC rather than local time.

0323pin:
https://www.reddit.com/r/rust/comments/tqzeek/what_lightweight_datetime_library_to_use_2022/
https://github.com/chronotope/chrono/commits/main

Someone has taken over maintenance of `chrono` and a new release to fix the CVE is on-going.

mautamu:
The bigger issue is the performance, which I don’t think chrono solves? But I suppose if it’s still feature gated we’re alright as folks can choose to take the leftwm-state performance hit or not as they need to.

0323pin:
> The bigger issue is the performance

oh, I didn't think about that. Honestly, I was only pointing out that it appears to be back under active development.
Personally, I'm simply using `date` from `liquid` on my status bar and have moved the clock to my shell prompt.

mautamu:
Closing this as I’m no longer actively working on it.

Issue #685, opened 2022-01-29
Scratchpad breaks configuration


lieskjur:
**Describe the bug**
config does not pass `leftwm-check` after adding a scratchpad to it.

**To Reproduce**
1. Add lines to `config.toml`
```
[[scratchpad]]
name = "Alacritty"
value = "alacritty"
x = 0.46
y = 0.20
height = 0.80
width =  0.54

[[keybind]]
command = "ToggleScratchPad"
value = "Alacritty" # Name set for the scratchpad
modifier = ["modkey"]
key = "t"
```
2. save `config.toml`
3.
```
~> leftwm-check
:: LeftWM version: 0.2.11
:: LeftWM git hash: 399345b
:: Loading configuration . . .
Configuration failed. Reason: duplicate field `scratchpad` at line 332 column 1
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK
```

**Environment:**
 - OS/Distro: Manjaro

AethanFoot:
What else is in your config, adding it to mine doesn't seem to produce the same error.

lieskjur:
Here is the full config [https://github.com/lieskjur/leftwm-config/blob/44751498083908afa9cbd02b55c23253cd7f71e8/config.toml](url)

AethanFoot:
Ah it is because of `scratchpad = []` at line 8.

lieskjur:
just noticed it as well when I fixed the link -_-. Thanks for bearing with me

PR #747, opened 2022-03-28
add the `leftwm-config` command

blackdragon2447:
I've added the `leftwm-config` command as mentioned in #685.

### DIFF
--- a/leftwm/src/bin/leftwm-config.rs
+++ b/leftwm/src/bin/leftwm-config.rs
@@ -1,0 +1,81 @@
+use anyhow::Result;
+use clap::{App, Arg};
+use leftwm::Config;
+use std::env;
+use std::fs::File;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use xdg::BaseDirectories;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let matches = App::new("LeftWM Check")
+        .author("Lex Childs <lex.childs@gmail.com>")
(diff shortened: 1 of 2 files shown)


VuiMuich:
Thanks @blackdragon2447 this is really nice!

One thing I was wondering, though, if we possibly want to handle this more like `leftwm-theme` and don't 'ship' it by default.
Main reasoning behind this:
- We would keep the base compilation as quick and slick as it is
- More features for config convenience like some interactive/tui configuration could be added easily without the risk of starting to bloat the base implementation of the wm
- At some point I was considering if `leftwm-check` and `leftwm-command` should become 'degraded' from being a individual binaries to really be just some moduled subcommands the of `leftwm` bin.

blackdragon2447:
Removing the seperate binairies could be interseting as it'd allow them to be more complex withoit bloating the main binairy(as you stated), tho the reason that leftwm-config came up was as a way to generate the confg as an install hook for package managers/maintainers.
On the other hand, the idea of having a(n) (n)curses leftwm-config tool sounds quite appealing since it would make leftwm more accessible, plus one could have leftwn-check part of leftwm-config.
leftwm-command should (in my opinion) be kept part of the main package in one form or another tho since having that seperate would defenitly complicate things like theme stuff or the use of exteral keybind daemons.

AethanFoot:
What we could do is make a repo like leftwm-utils that could store them. Then make is so the packages can be built individually or all together.

blackdragon2447:
that would be a good idea, I currently have a repo with the current code (link on discord), but I can move it to an official leftwm-utils repo.

Issue #561, opened 2021-11-01
Allow disabling the window-snap feature


hertg:
I generally don't use the window-snap feature, mostly because of its rough behavior described in #486.
Sometimes I'd like to position floating windows at the edge or even beyond the edge of the screen, the snap feature makes that a bit awkward because all windows jump around when I move the window over the edge.

My proposal is to add a config key to disable that feature, similar to `disable_current_tag_swap`, eg. `disable_window_snap` or something similar.

akosfekete:
I also had similar problems with snapping, I was surprised to see that there is no option to disable it. I created a fork that adds the `disable_window_snap` config key, the change seems trivially simple (unless I'm overlooking something), it works fine for me. Could there be a reason for not simply adding this config key (like avoiding bloated config)? Or are you planning to add it in #569?

hertg:
No you probably didn't miss anything, it should be fairly straight forward. Maybe open a PR with your change? It will most probably get merged :)

*My PR #569 came to a halt, because I wasn't able to figure out the position where the window would end up after it snaps, so the snapping-display is still not very user-friendly with my changes. It fixes the snap flickering, but while it is in "snapping" mode, the window gets displayed over the whole workspace, rather than where it actually ends up if you let go of it. I'm currently working on a new layout-system that should allow us to determine future window positions without re-positioning any actual windows.

akosfekete:
Thanks for the heads-up! I created a PR ( #754 )

hertg:
Was implemented in #754

PR #754, opened 2022-04-23
Allow disabling window snapping

akosfekete:
Add disable_window_snap config option as mentioned in #561 

I would update the config wiki page as below.

# Snapping Behaviour
If snapping is enabled, floating windows are snapped (repositioned to a tile) when the window  is dragged close to a Workspace edge. This might lead to undesired behaviour when dragging windows near the edge of the screen, so it is disabled by default.

Snapping can be disabled with `disable_window_snap`:

Default: `disable_window_snap = true`

Example: `disable_window_snap = false` (enables snapping)

### DIFF
--- a/leftwm-core/src/config/mod.rs
+++ b/leftwm-core/src/config/mod.rs
@@ -54,6 +54,7 @@ pub trait Config {
     fn get_list_of_gutters(&self) -> Vec<Gutter>;
     fn max_window_width(&self) -> Option<Size>;
     fn disable_tile_drag(&self) -> bool;
+    fn disable_window_snap(&self) -> bool;
 
     /// Attempt to write current state to a file.
     ///
@@ -176,6 +177,9 @@ impl Config for TestConfig {
         None
     }
     fn disable_tile_drag(&self) -> bool {
+        false
+    }
+    fn disable_window_snap(&self) -> bool {
         false
     }
     fn save_state(&self, _state: &State) {
--- a/leftwm-core/src/handlers/window_move_handler.rs
+++ b/leftwm-core/src/handlers/window_move_handler.rs
@@ -18,11 +18,12 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
             Some(w) => w.margin_multiplier(),
             None => 1.0,
         };
+        let disable_swap = &self.config.disable_window_snap();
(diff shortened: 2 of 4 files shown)


hertg:
Thanks for submitting :)
Fairly straight-forward PR, I don't have any objections to the changes.
I would even be in favor of defaulting the setting to `true`, because I don't see anyone really using the feature. It is more of an annoyance in its current state. But maybe I'm wrong, are people using it?

@VuiMuich @AethanFoot Thoughts?

VuiMuich:
I agree with defaulting to `true` in this case even though it is an actual breaking change, but I would expect the support requests for this change would still be outnumbered by the requests to have this feature at all.

@akosfekete would you mind pushing this little change? When we get no veto against this new default I'll merge tomorrow.

akosfekete:
I updated the default value and changed the doc template. I also think defaulting to true makes sense, I was confused by this feature when I started using leftwm (maybe this is more useful when using multiple workspaces on ultrawide monitors?)

juanCortelezzi:
The `disable_window_snap` option is enabled by default but I still have some snapping action going on when a floating window is near the border of the screen. I believe this [line](https://github.com/leftwm/leftwm/blob/7b2a1b79ba43484d64383ea6de52ae32cb97bf24/leftwm-core/src/handlers/window_move_handler.rs#L26) might be my issue. Since I use a single monitor, therefore a single workspace I would like to disable this feature and prevent the snapping. Is there any way to do this at the moment? thanks!

VuiMuich:
Which build hash are you on? I cant reproduce with my local branch that i forked from main last week (to be precise I am slightly ahead of #793, built with the `ron-config` feature) and I am also on a single monitor.

juanCortelezzi:
`LeftWM 0.3.0, Git-Hash: eaf531d` - downloaded from the AUR, also this is [my config](https://github.com/juanCortelezzi/Dotfiles/blob/main/leftwm/.config/leftwm/config.toml).


https://user-images.githubusercontent.com/57237705/180614156-e427ecd8-5752-4c30-b0d3-66c8c8b42a15.mp4

VuiMuich:
NP, afaik the AUR "release" only gets bumped, when we update the crate as well. But I believe there is a `leftwm-git` package as well.

PR #757, opened 2022-05-01
Add moving window to next/previous tag (#751)

### LINKED ISSUE #751: Move to next/previous tag
There's no way to move a window to the next or previous tag.
Something like MoveWindowToNextTag or MoveWindowToPreviousTag instead of workspace would be very useful.

Thanks in advance.

Zwo1in:
# Description

Add moving window to the next or previous tag.

When executing while working on first/last tag, the command will move window wrapping around tag list.

Fixes #751

## Type of change

- [x] This change requires a documentation update (wiki / Keybind Commands)

### Wiki

# Keybind Commands

...

## MoveWindowToNextTag

Takes the window that is currently focused and moves it to the next tag. If `value` is `true`, the focus will follow the window. Next tag is brought to the current workspace and focused.

Example:

```toml
[[keybind]]
command = "MoveWindowToNextTag"
value = "true"
key = "l"
```

## MoveWindowToPreviousTag

Takes the window that is currently focused and moves it to the previous tag. If `value` is `true`, the focus will follow the window. Previous tag is brought to the current workspace and focused.

Example:

```toml
[[keybind]]
command = "MoveWindowToPreviousTag"
value = "true"
key = "h"
```

## MoveToLastWorkspace
...

---

### command_handler

I've been struggling a bit with my command not taking an effect when running with `leftwm-command` so I've updated section about where to look when adding/renaming commands.

See [CONTRIBUTING.md User Documentation section](../CONTRIBUTING.md#user-documentation) for further details. (Seems broken link)

---

# Checklist:

- [ ] Manual page has been updated accordingly (Not needed unless we want this as default)

### DIFF
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -17,7 +17,10 @@ use crate::utils::{child_process::exec_shell, helpers};
 use crate::{config::Config, models::FocusBehaviour};
 
 impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
-    /* Please also update src/bin/leftwm-check if any of the following apply after your update:
+    /* When adding a command
+     * please update src/utils/command_pipe and leftwm/src/command if:
+     * - a command is introduced or renamed
+     * please also update src/bin/leftwm-check if any of the following apply after your update:
      * - a command now requires a value
      * - a command no longer requires a value
      * - a new command is introduced that requires a value
@@ -57,6 +60,9 @@ fn process_internal<C: Config, SERVER: DisplayServer>(
         Command::ToggleSticky => toggle_state(state, WindowState::Sticky),
 
         Command::SendWindowToTag { window, tag } => move_to_tag(*window, *tag, manager),
+        Command::MoveWindowToNextTag { follow } => move_to_tag_relative(manager, *follow, 1),
+        Command::MoveWindowToPreviousTag { follow } => move_to_tag_relative(manager, *follow, -1),
+        Command::MoveWindowToLastWorkspace => move_to_last_workspace(state),
         Command::MoveWindowToNextWorkspace => move_window_to_workspace_change(manager, 1),
         Command::MoveWindowToPreviousWorkspace => move_window_to_workspace_change(manager, -1),
         Command::MoveWindowUp => move_focus_common_vars!(move_window_change(state, -1)),
@@ -68,7 +74,6 @@ fn process_internal<C: Config, SERVER: DisplayServer>(
 
         Command::CloseWindow => close_window(state),
         Command::SwapScreens => swap_tags(state),
-        Command::MoveWindowToLastWorkspace => move_to_last_workspace(state),
         Command::NextLayout => next_layout(state),
         Command::PreviousLayout => previous_layout(state),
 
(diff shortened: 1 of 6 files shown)


Zwo1in:
I'm wondering whether it would be useful to add an argument to this command whether focus should follow

AethanFoot:
For the additional argument I would say whether to follow makes sense, then if it goes to a hidden tag it brings that tag to the current workspace and focuses (similar to wmctrl commands). Great work!

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -276,6 +281,21 @@ fn move_to_tag<C: Config, SERVER: DisplayServer>(
...
+fn move_to_tag_relative<C: Config, SERVER: DisplayServer>(
+    manager: &mut Manager<C, SERVER>,
+    delta: i32,
+) -> Option<bool> {
+    // Map indexing from 1..len to 0..(len - 1)
+    let current_tag = manager.state.focus_manager.tag(0).unwrap_or_default() - 1;
+    // apply euclidean division reminder to the result of offseting to wrap around tags vector
+    // and add 1 to remap back to 1..len indexing
+    let tags_len = manager.state.tags.normal().len() as isize;
+    let desired_tag = (current_tag as isize + delta as isize).rem_euclid(tags_len) + 1;
```
AethanFoot, line 294: Really like this btw (does it handle negatives as expected).

Zwo1in: thanks :) (yes it does, dunno if that was a question)

Zwo1in:
> For the additional argument I would say whether to follow makes sense, then if it goes to a hidden tag it brings that tag to the current workspace and focuses (similar to wmctrl commands). Great work!

So I've added the additional argument whether focus should follow the window. I've enabled it as a default as it feels convenient but this may be subjective.

I use single monitor setup at home, will compile and play with this on 2 monitors tomorrow at work.

I've also discovered an issue when testing this one, when you have selected a layout which doesn't equally size windows (e.g. RightWiderLeftStack) and you change tags back and forth, windows gets resized to being equal until you change layout back and forth again. Will describe that in a day and can dig it further if nobody minds.

Zwo1in:
I've updated Wiki documentation proposal in first comment but just realized I'm also missing external commands section

Zwo1in:
So yesterday's implementation had one bug. When focus was `Sloppy` then there was a case when moving window to the other tag where is some other window. If mouse ended up being over the other window, it gained the focus instead of the moved window.

I'm not sure how could I write a test versus such a case but I've just pushed a version with fix for it

VuiMuich:
> So yesterday's implementation had one issue. When focus was `Sloppy` then there was a bug, if after moving window to the occupied tag the mouse ended up being over the other window, it gained the focus instead of the moved window.
> 
> I'm not sure how could I write a test versus such a case but I've just pushed a version with fix for it

Sorry for the delayed reply.
I am not certain I follow the bug description correctly, how is it possible to move a window from another tag to the focused one? The current focused window always 'enforces' the focused tag. But in sloppy the mouse gets centred on the focused window so it refocuses and I see how that can lead to an unintended behaviour.
Maybe just leave a `TODO` commend that a test for this situation is desired.

Other then that this looks merge-ready to me. @AethanFoot what do you say?

Zwo1in:
> I am not certain I follow the bug description correctly, how is it possible to move a window from another tag to the focused one? 

The case was that when moving window to the next/prev tag and then instantly also focusing this tag, current cursor position was deciding factor what window from the new tag got focused. Eg. you had one window on a tag next to the current. You moved the current window to the next tag and brought focus to this tag. If you had the cursor on the left side of the screen, the left window got focused, if on the right side then the right one. Didn't matter where the original window was placed. It resulted in scenarios where when moving window a few tags next or previous, you could start moving other window by accident. Not sure if you'd test this, I can add TODO if you wish

Zwo1in:
Huh, I think that the best explanation of this would be to just show the diff, however it seems like I cannot choose to see diff between last two commits (latter being the former amended). I've never used github much as I work with gitlab on daily basis but does github dislikes amends and rebases? I've always thought this is how git is meant to be used...

Nevertheless, here is probably close to how diff looks like:
```diff
-    if follow {
-        manager.state.goto_tag_handler(desired_tag);
-    }
+    if follow {
+        let moved_window = *manager.state.focus_manager.window_history.get(1)?;
+        manager.state.goto_tag_handler(desired_tag);
+        manager.state.handle_window_focus(&moved_window?);
+    }

VuiMuich:
After a long wait I finally had some time to put it to a test and every thing seems to be fine.
Thanks!

PS: would you please update the wiki? Thanks.

Zwo1in:
Sure

Zwo1in:
done

VuiMuich:
Thanks again!

PR #763, opened 2022-05-11
Correctly assign main width percentage when switching layouts

### LINKED ISSUE #762: Issues regarding main width percentage and `LeftWiderRightStack` layout
**Describe the bug**
So there are two problems that I'm aware of.

First one:
Main window width percentage set for `LeftWiderRightStack` layout and it's sibling does not persist tag change when using `LayoutMode::Workspace`.

When using `LayoutMode::Workspace` and switching layout to `LeftWiderRightStack`, main window's width is set to 75% . When you then switch to a different tag, all tags will now have equal windows (50%) even tho the layout hasn't changed.

Second one:
Custom main window width percentage behavior is not consistent when switching layouts.

After increasing/decreasing main window width, switching layout to any other that supports main window resizing keeps my overridden main window width. Switching to and from to the layouts that don't support main window resizing doesn't make my setting forgotten when switching back. However switching to and from `LeftWiderRightStack` and it's sibling behaves like for layouts not supporting main resizing but this layout allows resizing.
After increasing/decreasing main window width when on `LeftWiderRightStack` and then switching to other layout however keeps the width change but it looks like the width change was done relative to the base 50% width, not the `LeftWiderRightStack`'s 75%. So e.g. if you decrease main window width from 75% to 50% on `LeftWiderRightStack` and then switch to `MainAndVertStack`, the main window width will be 25%, not 50% as I'd expect.

**Solutions**
I see two possible ways to make this behavior consistent.

First one:
When switching to another layout, always determine whether user has changed `main width` by checking if current main width equals layout's default main width. If user has default main width, then set workspace's main width to the default value of layout. If user has changed main width, then always keep it, even for `LeftWiderRightStack` and it's sibling.

Second one:
Disallow changing main width percentage for `LeftWiderRightStack` and it's sibling and ignore value of workspace's main width within layout's `update` method.

First solution makes `LeftWiderRightStack` basically a `MainAndVertStack` with starting default main width set to 75%, but after changing main width anywhere those two will be identical.

Second solution much more separates `LeftWiderRightStack` from `MainAndVertStack` and makes it kinda a static, predefined layout.

I'm not sure which one I like more but I'd probably favor the latter as the first feels like killing this layout, one could just resize `MainAndVertStack` and has the exact same behavior afterwards. Kind of shame as I already implemented the former while digging this topic, but I'm okay with implementing the second one also :upside_down_face: 

**Environment:**
 - OS/Distro: Arch 5.17.5-zen
 - Output of `leftwm-check`
 
```
❯ leftwm-check
:: LeftWM version: 0.3.0
:: LeftWM git hash: c0fcf84
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK
```

Zwo1in:
# Description

This is an implementation of first proposed solution in #762:

When switching to another layout, always determine whether user has changed main width by checking if current main width equals layout's default main width. If user has default main width, then set workspace's main width to the default value of layout. If user has changed main width, then always keep it, even for LeftWiderRightStack and it's sibling.

Fixes resetting main width then switching tags on `LayoutMode::Workspace` and `LeftWiderRightStack` as well as makes its behavior of main width percentage while switching layouts consistent with other layouts.

Fixes #762.

## Type of change

None

# Checklist:


Zwo1in:
Will probably update this PR and switch to the second proposed solution in #762.

VuiMuich:
Hey @Zwo1in, finally found a bit of time to go through the open PRs.
I guess we might put mark this as draft, if you are still interested in implenenting the second solution as stated, right?

Zwo1in:
Sure! I remember that I had some doubts about the second idea but will need to recall/rethink it. Kinda forgot about those PRs for a white too ;)

VuiMuich:
Hey @Zwo1in may I nudge you in a friendly way and ask if you might find a bit of time to git this done?

In order to make a return easier for you I went ahead and pushed a fix for the merge conflicts that occurred in the meantime.

Zwo1in:
Hi, thanks for a reminder. To be honest I can't ensure you that I'll get back to it at least during the next few weeks as I'm a bit a bit busy. I've also got some MacBook (;_; tiling so bad there) so I'm currently not really using left on daily basis. By the way, what's the state of @hertg work on new layout system? It could be an alternative solution

hertg:
Yeah, i think it's fine leaving this in draft or closing it. The layout rework will replace that code anyway I think.

VuiMuich:
Ok, then lets close this.
Thanks for your efforts anyways @Zwo1in, hope to see you back and hopefully being able to merge you next contribution.

PR #765, opened 2022-05-14
Added option to stop cursor from following focused windows

Syudagye:
# Description

A little feature that adds the `mouse_follows_focus` option that will decide if the cursor will follow windows as they're focused or not.
Currently when you change the focused window, the cursor will be placed on the center of the newly created window, and that can sometimes be a bit annoying.
I can also lead to some problematic behaviors: for example when a window spawns a popup sub-window, and focusing the parent window forces you to focus the popup, the cursor would always teleport in the center of the popup when you try to escape it, and so your cursor would be stuck in the popup. This happened a lot to me with kde applications, such as kdenlive and krita.
I've discussed this feature idea briefly on discord a while ago, but didn't really went to try to implement it since then.

I couldn't find any related issue for this feature, let me know if one exists so i can reference it here.

This PR is just a draft, i couldn't finish it as fast as i would have liked it, so there is still stuff to do, and also I need others to see if the implementation is ok or needs more work.

The cursor will not teleport when:
- [x] Changing focus
- [ ] Spawning/Destroying a window
- [ ] Resizing the main window
- [x] Changing layout*
- [x] Changing tag/workspace*

<sub>
* Those looks like they are working, but still need to find if there is no issue
</sub>

## Type of change

*Will add later*

# Checklist:

- [ ] Enhanced review is performed with `cargo clippy -- -W clippy::pedantic -A clippy::must_use_candidate -A clippy::cast_precision_loss -A clippy::cast_possible_truncation -A clippy::cast_possible_wrap -A clippy::cast_sign_loss -A clippy::mut_mut`

### DIFF
--- a/leftwm-core/src/config/mod.rs
+++ b/leftwm-core/src/config/mod.rs
@@ -55,6 +55,7 @@ pub trait Config {
     fn max_window_width(&self) -> Option<Size>;
     fn disable_tile_drag(&self) -> bool;
     fn disable_window_snap(&self) -> bool;
+    fn mouse_follows_focus(&self) -> bool;
 
     /// Attempt to write current state to a file.
     ///
@@ -195,6 +196,9 @@ impl Config for TestConfig {
         } else {
             false
         }
+    }
+    fn mouse_follows_focus(&self) -> bool {
+        true
     }
 }
 
--- a/leftwm-core/src/display_servers/mock_display_server.rs
+++ b/leftwm-core/src/display_servers/mock_display_server.rs
@@ -1,6 +1,7 @@
 use super::Config;
 use super::DisplayEvent;
 use super::DisplayServer;
+use crate::models::FocusManager;
 use crate::models::Screen;
 
 #[derive(Clone)]
@@ -26,7 +27,7 @@ impl DisplayServer for MockDisplayServer {
         unimplemented!()
     }
 
-    fn generate_verify_focus_event(&self) -> Option<DisplayEvent> {
+    fn generate_verify_focus_event(&self, _: &mut FocusManager) -> Option<DisplayEvent> {
         unimplemented!()
     }
 }
--- a/leftwm-core/src/display_servers/mod.rs
+++ b/leftwm-core/src/display_servers/mod.rs
@@ -1,5 +1,6 @@
 use crate::config::Config;
 use crate::display_action::DisplayAction;
+use crate::models::FocusManager;
 use crate::models::Window;
 use crate::models::WindowHandle;
 use crate::models::Workspace;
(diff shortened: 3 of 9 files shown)


Syudagye:
For now what i wanted for my personal use case have been addressed, but i'm looking forward to improve this.
First, i don't know if there is a better way to implement this option in the config (would it be better to have a completely new behavior?).

Also regarding the feeling of this addition, i made what was the most natural for me: the focus is recalculated when the cursor moves. But after looking at other window manager i noticed that it often changes focus only when a window is enter with the mouse. So i would like some feedback on which one could be the better option, and the most natural.

Also i have no idea why, but changing to a tag will focus the window under the cursor, even if i modified this in `focus_handler.rs`. I don't know what is forcing the focus on the mouse here.

VuiMuich:
Just started to get into this topic, but while answering questions on discord a thought came up: the reason why we have this is that in `Sloppy` focus mode when you drive focus away with the keyboard you would get unintentional refocus as soon as you move the mouse. In `ClickTo` and `Driven` this should not happen.
Of course with the necessary awareness, you could even build a workflow upon this refocus 🙃

VuiMuich:
> For now what i wanted for my personal use case have been addressed, but i'm looking forward to improve this. First, i don't know if there is a better way to implement this option in the config (would it be better to have a completely new behavior?).

The more I get into this, the more I get the feeling this should actually be a variant of the `Sloppy` mode, as `ClickTo` and `Driven` are not effected by the initial issue at all, right?
 
> Also regarding the feeling of this addition, i made what was the most natural for me: the focus is recalculated when the cursor moves. But after looking at other window manager i noticed that it often changes focus only when a window is enter with the mouse. So i would like some feedback on which one could be the better option, and the most natural.

Personally I think recalculating focus, when a window border is crossed is the most natural way, and I would expect it to be possibly a little less expensive.

> Also i have no idea why, but changing to a tag will focus the window under the cursor, even if i modified this in `focus_handler.rs`. I don't know what is forcing the focus on the mouse here.

:thinking: I would have guessed, that the same as you have in `focus_handler.rs:80`

Syudagye:
> The more I get into this, the more I get the feeling this should actually be a variant of the `Sloppy` mode, as `ClickTo` and `Driven` are not effected by the initial issue at all, right?

Yes, it's only with `Sloppy`. Perhaps renaming the `mouse_follows_focus` option to something that evokes that this is to activate a variant of the behaviour would be better. Maybe something like `sloppy_static_mouse` or `sloppy_mouse_dont_follow_focus` (the last one seems a bit too long tho...).

> Personally I think recalculating focus, when a window border is crossed is the most natural way, and I would expect it to be possibly a little less expensive.

Ok this will be the way then.

Thank you for the feedback, i'll implement this asap !

Syudagye:
I finished tweaking the little things and i think it is now ready for merge.
I updated the wiki and the man page, so check it to see if it's ok like this of if it needs modifications

VuiMuich:
Reading on mobile this seems good to me.
I hope I can have a proper look at it before or at the weekend.

PR #768, opened 2022-05-22
Add commands `ReleaseScratchPad`, `AttachScratchPad`, `NextScratchPadWindow` and `PrevScratchPadWindow`

SamClercky:
# Description

This is for an issue that I had when using a scratchpad: If I wanted to move the window in the scratchpad to one of my tags, Leftwm would still remember the window as being in the scratchpad.

This PR removes the moved window from the active scratchpad with a new command `ReleaseScratchPad`.

~~There are no extra dependencies or configs (in my opinion not needed) added.~~

In the meantime this PR adds 3 additional commands on top of `ReleaseScrachPad`: `AttachScratchPad`, `NextScratchPadWindow` and `PrevScratchPadWindow`. It also adds the ability for a scratchpad to contain more than 1 window, but restricts its use to a stack of windows that can be paged with the added next/prev commands.

Fixes #(issue): I couldn't find one, but if needed, I can create one for better visibility.

## Type of change

# Keybind

## ReleaseScratchPad

Moves the window in a scratchpad to a tag and removes the empty scratchpad.

The value can be one of the following:
* Empty or omitted: Use the currently focused scratchpad and the currently focused tag.
* `<tag id>`: Move the currently focused scratchpad to the tag with `<tag id>`
* `<scratchpad name>`: Move the window in the scratchpad with the name `<scratchpad>` to the current tag even if the scratchpad is not currently visible.

Example:

IN TOML
```toml
[[keybind]]
command = "ReleaseScratchPad"
value = "Alacritty"
modifier = ["modkey", "Control", "Shift"]
key = "p"
```

IN RON
```rust
(command: ReleaseScratchPad, value: "Alacritty", modifier: ["modkey", "Control", "Shift"], key: "p"),
```

Another use case is to move the window in a scratchpad to a specific tag and removing the empty scratchpad
in 1 keybinding:

IN TOML
```toml
[[keybind]]
command = "Execute"
value = "leftwm-command \"ReleaseScratchPad\" \"SendWindowToTag 1\""
key = "1"
```

IN RON
```rust
(command: Execute, value: "leftwm-command \"ReleaseScratchPad\" \"SendWindowToTag 1\"", modifier: ["modkey", "Shift"], key: "1"),
```

## AttachScratchPad

Attaches the currently selected window to a scratchpad. The value needs to be the name of the scratchpad to which the selected window will be attached. This command makes it possible to have more than 1 window in a scratchpad.

Example:

IN TOML
```toml
[[keybind]]
command = "AttachScratchPad"
value = "Alacritty"
key = "1"
```

IN RON
```rust
(command: AttachScratchPad, value: "Alacritty", modifier: ["modkey", "Shift"], key: "1"),
```

## NextScratchPadWindow/PrevScratchPadWindow

Cycles through the scratchpad windows. The value needs to be a valid scratchpad name. Note that this command will be ignored if the given scratchpad is not visible.

Example:

IN TOML
```toml
[[keybind]]
command = "NextScratchPadWindow"
value = "Alacritty" # Name set for the scratchpad
modifier = ["modkey", "Shift", "Control"]
key = "comma"

[[keybind]]
command = "PrevScratchPadWindow"
value = "Alacritty" # Name set for the scratchpad
modifier = ["modkey", "Shift", "Control"]
key = "period"
```

IN RON
```rust
(command: NextScratchPadWindow, value: "Alacritty", modifier: ["modkey", "Shift", "Control"], key: "comma"),
(command: PrevScratchPadWindow, value: "Alacritty", modifier: ["modkey", "Shift", "Control"], key: "period"),
```

# External Commands

| Command | Arguments (if needed) | Notes |
| -------------- | ----------------------------- | -------- |
| ReleaseScratchPad | `scratchpad name`, `tag id` or no argument | Move the currently focused scratchpad or the scratchpad with `scratchpad name` to `tag id` or the current tag if nothing provided as argument |
| AttachScratchPad | `scratchpad name` | Attach the currently selected window to the given scratchpad |
| NextScratchPadWindow | `scratchpad name` | Cycle to the next scratchpad window only if the given scratchpad is visible |
| PrevScratchPadWindow | `scratchpad name` | Cycle to the previous scratchpad window only if the given scratchpad is visible |

# Checklist:

EDIT 1: added documentation for new command suggestion
EDIT 2: added documentation for `AttachScratchPad`-, `NextScratchPadWindow`- and `PrevScratchPadWindow` commands + suggestions
EDIT 3: Translate keybind config to the new RON format + extra info about this PR in the description


SamClercky:
Hi VuiMuich,

Not entirely: When I move the window in the scratchpad to a tag and then try to toggle the scratchpad again, I expect the scratchpad to create a new window, but the moved window gets hidden instead.

This is because Leftwm still remembers the moved window as being a scratchpad. This PR fixes this edge case by checking if the 
moved window is not a scratchpad and if it is, removing it from the `active_scratchpads`.

SamClercky:
Works for me. When I have time, I will make the adjustments.

VuiMuich:
Hey, finally had some time to test this a bit.

The only thing I can foresee that this creates a demand for some `AttachScratchPad` command :grin: 
If you are intrigued to extend this PR by this related feature it would be great. But we can also open a feature request issue for this and add it later. That would be also fine, and we could merge this immediatelly.

VuiMuich:
LGTM and works great. Thanks :pray: 

One thing though: when multiple windows get attached to a single scratchpad only the last attached window can be accessed. This should be mentioned in the documentation.

In the future we should add the possibilty to focussing and moving commands to decide via an optional value if scratchpads should be included or not (maybe even a variant to effect scratchpads exclusivlys?).

VuiMuich:
Sorry just found one more thing, that I believe es better fixed before merge:
`AttachScratchPad` should do a little check if it is already attached to the destination scratchpad. Otherwise it is possible to nest it one layer deeper in itselfe. This might be another feature for a future extension, but as of right now I deem it 'unexpected behavior' and might lead to confusion.

SamClercky:
Some notes about the `AttachScratchPad` command: The current implementation only supported 1 window at a time in a scratchpad. With this new command it would be possible for the user to add more than 1 window to a scratchpad. A big part of the new code is to accomodate the possibility of more than 1 window in a scratchpad.

The last few commits (not the last one) added some extra checks to the scratchpad related commands so a window can not be added twice to the same scratchpad and that every time a window is shown, the window is first checked if it is still active by looking it up the pid in the still managed windows.

The last commit adds 2 more commands that I really wanted now that the scratchpads can have more than 1 window. These are `NextScratchPadWindow` and `PrevScratchPadWindow`. These make it easy to cycle through the scratchpad windows without having to attach and then release them.

I hope you like these new additions as much as I liked making them :smile:.

VuiMuich:
Awesome work.
I've just scrolled through the latest commits on mobile so far, but this looks all good so far.
I'll see if I can outing to test tomorrow or on Thursday.

I was a bit reluctant about the Prev/Next commands, as generally I would probably like an option to the existing moving and focusing commands a bit better. But I totally get, why you went this route and I guess for now it's absolutely ok to do it this way.

Maybe in the future we will be able to support the sofisticated keybind vision (or delusion, who knows) I sport from time to time...

VuiMuich:
Hey,
I tested a bit and noticed a couple of focus related things:
- it seems the `NextScratchPadWindow` and `PrevScratchPadWindow` commands toggle the scratchpad on, when it is hidden, is this indended?
- when the scratchpad is visible,  `FocusUp` is not working anymore
- when the scratchpad is visible,  `FocusDown` is only working to switch focus with the tags `Main` window
- when the scratchpad is visible,  `FocusTop` is behaving really weird, just play around it would be quite verbose to describe

Otherwise the latest look really good. Sorry its "one more thing" every time :grin:

SamClercky:
Sorry for the late response. My pc had some problems so I couldn't really test immediatly the different focus commands.

I tried the different focus commands and found that the behavior is not perfect with the scratchpad. I tried to find what was the original behavior without my code, but found it had the same behavior. I could not find a regression (maybe something I have misted?).

I can see that the current behavior is not ideal, but would like a better description of the desired behavior. Maybe this should be moved to its own PR? If you would still want me to implement it in this PR, could you point me to a place in the code that solves a similar problem that I could use as an example. I am not really familiar with the parts concerning the focus behavior and its expected invariants.

Concerning the scratchpad becoming visible when cycling the scratchpad is not intensional and I am looking for a fix :smile:. There is a check at the beginning of `cycle_scratchpad_window` so it would not execute if the scratchpad is not visible (aka should not become visible), but somehow it gets bypassed and the scratchpad becomes visible.

Thank you for guiding me through this :smile:!

VuiMuich:
I think a good reference would be the `fn move_window_change` and `fn focun_window_by_class` with the special treatment for `Layout::Monocle` and `Layout::MainAndDeck`,
If possible I would prefer to merge this without any kinks, but if you would like to get this checked and move on I can relate and we can move the ironing to a new PR indeed.

Thanks anyways for your endurance and patience!

SamClercky:
Hi VuiMuich, 

I will look into this and try to fix these last issues. Real life is becoming really buzy at the moment, so it might take me some time to finish this.

For now I updated the PR with the new changes from `main`, so the code will not get to much behind :smile:.

Thank you for your patience.

VuiMuich:
Polite and humble *bump* and request if you have a prospect on when you might have another stab at this one.

SamClercky:
I finally have some time on my hands to work on this. I hope to be done by the end of the week :smile:.

SamClercky:
Hi VuiMuich,

I was able to fix the issue with the cycle scratchpad commands and some focus related problems not yet discussed that I discovered while fixing some tests :tada:.

For the focus up/down/top commands, I can see that there is an issue, but cannot pin it down to a certain section in the code. I wrote some tests in the last commit to make it easier to find the root cause. The tests however turn out to be green.

I suspect that the focussing issue is a specific interaction between certain parts of the code (outside of the command handler), but I have no clue where I should look. Given enough time the problem could probably be fixed, but due to the nearing merge of #682, it might be better to let it be for the time being and merge this PR.

Thank you for all your patience and support. I really appreciate all the work you guys do to push this project forward!

VuiMuich:
> Hi VuiMuich,
> 
> I was able to fix the issue with the cycle scratchpad commands and some focus related problems not yet discussed that I discovered while fixing some tests tada.
> 
> For the focus up/down/top commands, I can see that there is an issue, but cannot pin it down to a certain section in the code. I wrote some tests in the last commit to make it easier to find the root cause. The tests however turn out to be green.
> 
> I suspect that the focussing issue is a specific interaction between certain parts of the code (outside of the command handler), but I have no clue where I should look. Given enough time the problem could probably be fixed, but due to the nearing merge of #682, it might be better to let it be for the time being and merge this PR.
> 
> Thank you for all your patience and support. I really appreciate all the work you guys do to push this project forward!

Thank you for all the work you put into this.
Do you expect to have time to handle the merge conflicts yourself? Otherwise I can have a look at those as well.

Was the new undiscussed issue you are referring to, that in some cases we could end up without any focused window at all, and when no window has focus the scratchpad wont unhide?

SamClercky:
Hi VuiMuich,

> Do you expect to have time to handle the merge conflicts yourself?

I just pushed the merge from main, but I cannot test it on my machine since I seem not to be able to get my keybindings working with the new lefthk integration. Do I need to upgrade to the new RON format or is there some documentation I overlooked? I first started `lefthk-worker` and then `leftwm-worker` with the same config that I use normally (`toml`).

> Was the new undiscussed issue you are referring to, that in some cases we could end up without any focused window at all, and when no window has focus the scratchpad wont unhide?

The issue had to do with the change that I made to the scratchpads (`Option` -> `VecDeque`). There was a piece of code that would check if the current selected window was a scratchpad and if not change focus to the second window in the focus history without checking if it was a scratchpad window or not. The problem with this was that the focused tag could become the hidden NSP tag. This was the reason the hiding of the scratchpads did not always work and could reappear if hidden previously.

VuiMuich:
Did you restart your session? The `lefthk` change requires a restart of the `leftwm` bin, wich doesn't happen during a regular reload. This is independent of the `ron` update. Also, when no `config.ron` is found, we fall back to `config.toml` automatically at the moment. Only if that fails as well, we fall back to the internal defaults.

Running this here right now and everything looks good so far.

I think the merge artefacts I removed with the previous push, where just an oversight.

Now that I started to read your code more thoroughly, I think it might be worth to refactor some (basically all the relevant fn and the tests in `comman-handler.rs` and the `ReleaseScratchPadOption` definition in `command.rs) to a new `scratchpad-handler.rs.
I am sorry that this comes so late in this PR, but is just now that I realised how much code this adds.

I love your extensiv tests, btw! :heart:

SamClercky:
I got it working again on my pc :smile:! In the mean time I did the refactor by moving the `command_handler` to its own directory with a `scratchpad_handler.rs`. This makes it possible to keep most command handling functions private for the rest of the application and still have the separation.

> I love your extensiv tests, btw! :heart:

:smile:

> btw: I can edit your docs to ron if you like.

I would like to do that myself so I can learn the new format better. Thanks anyway for the suggestion :smile:

VuiMuich:
> I would like to do that myself so I can learn the new format better. Thanks anyway for the suggestion smile

Basically it condenses to using `rust` as syntax highlight, removing the `[[keybind]]` line, removing quotes from the `command` value, appending commata after each value, replacing equal signs with colon and wrapping all in parens. I also like to put it all in a single line for compactness.

VuiMuich:
Tbh. I have not read the `scratchpad_handler.rs` in all detail and it runs very well.
So from my side it looks all good. As this is a fairly large PR I'd like to have someone else to have a look, as not to miss something.

AethanFoot:
I should have time to look through today

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+                .windows
+                .iter()
+                .find(|window| Some(window.handle) == **handle)
+                .map(Window::visible)
+            {
+                visibilty
+            } else {
+                false
+            }
+        })
```
AethanFoot, line 76: instead of if-else, maybe something like
```
manager
.state
.windows
.iter()
.find(|window| Some(window.handle) == **handle)
.map_or(false, |window| window.visible())

File: leftwm-core/src/handlers/command_handler/scratchpad_handler.rs [hunk 2]
```diff
@@ -0,0 +1,1242 @@
...
+        None
+    };
+    if let Some(handle) = handle {
+        manager.state.handle_window_focus(&handle);
+    }
+
+    Ok(())
+}
+
+/// Inverse of `hide_scratchpad` and makes the scratchpad visible
```
AethanFoot, line 103: This comment should relate to itself rather than another function. Just say it shows a scratchpad

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+    if let Some(previous_tag) = previous_tag {
+        manager
+            .state
+            .focus_manager
+            .tags_last_window
+            .remove(&previous_tag);
+    }
+    // Show the scratchpad.
+    window.tag(current_tag);
+    window.set_visible(true);
```
AethanFoot, line 134: Again this shouldn't be necessary.

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+
+    // send tag changement to X
```
AethanFoot, line 136: Same as before

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+        log::info!(
+            "Dead window in scratchpad found, discard: window PID: {}",
+            window
+        );
+    }
+
+    None
+}
+
+/// Inverse of `next_valid_pid_test`
```
AethanFoot, line 168: Again not a fan of documenting based on another function

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+            return Some(window);
+        }
+
+        log::info!(
+            "Dead window in scratchpad found, discard: window PID: {}",
+            window
+        );
+    }
+
+    None
```
AethanFoot, line 185: Could this be abstracted in some way, a macro or a separate function

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+/// If the name of the scratchpad is incorrect, false will be returned
+fn is_scratchpad_visible<C: Config, SERVER: DisplayServer>(
+    manager: &Manager<C, SERVER>,
+    scratchpad_name: &str,
+) -> bool {
+    let current_tag = if let Some(tag) = manager.state.focus_manager.tag(0) {
+        tag
+    } else {
+        return false;
+    };
```
AethanFoot, line 198: This isn't necessary, but a nifty macro could be made for these two as the same thing is being done

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+        if let Some(first_in_scratchpad) = next_valid_scratchpad_pid(id, &manager.state.windows) {
+            if let Some((is_visible, window_handle)) = manager
+                .state
+                .windows
+                .iter()
+                .find(|w| w.pid == Some(first_in_scratchpad))
+                .map(|w| (w.has_tag(current_tag), w.handle))
+            {
+                if is_visible {
+                    // window is visible => Hide the scratchpad.
```
AethanFoot, line 238: Capitals

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+                    if let Err(msg) = hide_scratchpad(manager, &window_handle) {
+                        log::error!("{}", msg);
+                        return Some(false);
+                    }
+                } else {
+                    // window is hidden => show the scratchpad
```
AethanFoot, line 244: Capitals

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+    if let Some(windows) = manager.state.active_scratchpads.get_mut(&scratchpad) {
+        log::debug!("Scratchpad {} already active, push scratchpad", &scratchpad);
+        let previous_scratchpad_handle = manager
+            .state
+            .windows
+            .iter()
+            .find(|w| w.pid.as_ref() == windows.front())
+            .map(|w| w.handle);
+
+        // check if window already in scratchpad
```
AethanFoot, line 330: Capitals

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+        if windows.iter().any(|pid| *pid == window_pid) {
+            return Some(false);
+        }
+
+        windows.push_front(window_pid);
+        if let Some(previous_scratchpad_handle) = previous_scratchpad_handle {
+            hide_scratchpad(manager, &previous_scratchpad_handle).ok()?; // first hide current scratchpad window
```
AethanFoot, line 337: Capitals

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+/// none to select the current window.
+pub fn release_scratchpad<C: Config, SERVER: DisplayServer>(
+    window: ReleaseScratchPadOption,
+    tag: Option<TagId>,
+    manager: &mut Manager<C, SERVER>,
+) -> Option<bool> {
+    let destination_tag =
+        tag.or_else(|| manager.state.focus_manager.tag_history.get(0).copied())?;
+
+    // if None, replace with current window
```
AethanFoot, line 362: Capitals

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+    let window = if window == ReleaseScratchPadOption::None {
+        ReleaseScratchPadOption::Handle(
+            manager
+                .state
+                .focus_manager
+                .window_history
+                .get(0)?
+                .as_ref()
+                .copied()?,
+        )
```
AethanFoot, line 372: I think it would look nicer if this is assigned to a variable first

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+    } else {
+        window
+    };
+
+    match window {
+        ReleaseScratchPadOption::Handle(window_handle) => {
+            // check if window is in active scratchpad
```
AethanFoot, line 379: Capitals

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1242 @@
...
+                .find(|(_, id)| window.pid.as_ref() == id.front())
+                .map(|(name, _)| name.clone())?;
+
+            log::debug!(
+                "Releasing scratchpad {} to tag {}",
+                scratchpad_name,
+                destination_tag
+            );
+
+            // if we found window in scratchpad, remove it from active_scratchpads
```
AethanFoot, line 399: Capitals, Ill leave the rest of the comment for u to find :laughing:

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -318,7 +318,7 @@ fn sane_dimension(config_value: Option<Size>, default_ratio: f32, max_pixel: i32
 }
 
 // Get size and position of scratchpad from config and workspace size.
-fn scratchpad_xyhw(xyhw: &Xyhw, scratch_pad: &ScratchPad) -> Xyhw {
+pub fn scratchpad_xyhw(xyhw: &Xyhw, scratch_pad: &ScratchPad) -> Xyhw {
```
AethanFoot, line 321: If we are using this outside of this file, maybe we should place it in scratchpad_handler

AethanFoot (changes requested):
Mostly comment nitpicks (saves me changing them later). I would say move the scratchpad_handler back a directory so it is with the rest (I say this a goto_tag_handler is similar as in that it is only used by command_handler, but is in the main directory), or you could move goto_tag_handler within the folder as well. Overall nice work!

leftwm-core/src/command.rs
```diff
@@ -1,3 +1,4 @@
+pub use crate::handlers::command_handler::ReleaseScratchPadOption;
```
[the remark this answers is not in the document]
SamClercky: I saw this as a nice to have, so the interface for the commands would still look the same and provide a more coherent api when specifying the options for the `ReleaseScratchPad` command.

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
SamClercky: This would be not possible as clippy would give the warning `clippy::let_and_return`

leftwm-core/src/handlers/window_handler.rs
SamClercky: I would argue against moving it to scratchpad_handler as it is in my opinion a place to handle the commands specific for the scratchpad and general properties of the windowing of the scratchpad should be placed somewhere else.

If you still want this, I will move it :smile:

AethanFoot: Hmm, true. Maybe we should move scratchpad.rs to models. And add this (and future function like that) to there

SamClercky:
Hi @AethanFoot,

I applied most of the your comments to the code and put all the ones I fixed on resolved. For the ones where I still had doubts or couldn't immediatly resolve, I added some comments :smile:

AethanFoot (approved):
Looks good, nice work! Not set up to test right now, but seems like @VuiMuich has tested, so we can merge anytime (we need to figure which order we a gonna merge things so might need to sort some conflicts)

VuiMuich:
I have few regression with the latest commits: 
- toggling SP on only works when there is a window that has focus
- toggling SP off doesn't return focus to any other visible window
- `NextScratchPad`/`PrevScratchPad` first toggle the curren SP window of only with the next execution they open the next SP window

SamClercky:
Hi VuiMuich,

I can't seem to reproduce these issues. Can you provide me more information on how to produce the undesired behavior?

This is what I see on my PC in Xephyr (last commit) but behavior is similar on real hardware.

In the first `gif` I first create a window, close it and open a new scratchpad (toggle command):
toggle_SP_no_focused_window

In the second `gif`, I first open a window, open a SP, kill it -> refocused original window, toggle SP multiple times -> refocused original window
toggle_SP_refocus

The last issue with Next/Prev commands, sounds like the same bug you mentioned the 23rd of july. Could it be it is the same? If so I should have been fixed in [`1488821`](https://github.com/leftwm/leftwm/pull/768/commits/1488821e0be1a8b0a8354eb6be3364e85d187f77). I seem not to be able to reproduce it, but I might be overlooking something.

Thanks in advance :smile:

VuiMuich:
This is my relevant config (in ron):
```rust
    scratchpad: [
        (name: "DynamicScratchpad", value: ""),
    ],

    keybind: [
        (command: ReleaseScratchPad, value: "", modifier: ["modkey", "Shift"], key: "d"),
        (command: AttachScratchPad, value: "DynamicScratchpad", modifier: ["modkey", "Control"], key: "d"),
        (command: NextScratchPadWindow, value: "DynamicScratchpad", modifier: ["modkey", "Control"], key: "comma"),
        (command: PrevScratchPadWindow, value: "DynamicScratchpad", modifier: ["modkey", "Control"], key: "period"),
        (command: ToggleScratchPad, value: "DynamicScratchpad", modifier: ["modkey"], key: "d"),

]
```

For 1): on empty tag open new window -> assign to `DynamicScratchpad` -> toggle `off` -> toggle `on` does not work
For 2): on same tag open new window -> toggle `DynamicScratchpad` `on` (focus is on SP) -> toggle SP `off` => tiled window does not get focused
For 3): focus tiled window -> assign to `DynamicScratchpad` -> cycle SP => only every second command shows the next SP, with all SP hidden inbetween

If you need more I can see if I have time to do a screencapture tomorrow.

SamClercky:
Hi VuiMuich,

I tried to reproduce the bug, but couldn't trigger it.

From looking at your config, I suspect the issue may be in that the scratchpad value is empty (shell command that does not generate a window). This creates a place in the scratchpad `VecDeque` without representing a real window.

It might be that the scratchpad code tries to show something, but actually only a non existing window is shown.

However this should not be possible as the cycle and toggle commands filter these invalid window pid's out. (see `next_valid_scratchpad_pid`)

Another option for me not being able to trigger the bug is that it might be a timing related issue and that my computer is a little slower/faster, so I won't get into an invalid state.

Could you maybe show me the `RUST_LOG=debug` logs as the filtering of the invalid pid's should be visible there as in `INFO: Dead window in scratchpad found, discard: window PID xxx`.

What do you think? I would realy like to have this fixed and get this PR merged :smile:

VuiMuich:
Just a quick update, as I started digging further into this: I reverted to `[5589fe7](https://github.com/leftwm/leftwm/pull/768/commits/5589fe7ff323b1c8481e49ea6d436ddbf70754d2)` and everything works normal. Now rebuilding with the latest again and see, if there maybe was some unrelated hickup fooling me.

BTW. part of the regression was also present for scratchpads configured in the 'classical' pre this PR way, I just omited them from the config example trying to be as minimal as possible.

Will edit this post with updates later.

VuiMuich:
Ok, when I `git revert 4e48f7f` everything is ok.
Maybe the manual `visibility` hanbling is actually helpful.

I'll push my revert, if @AethanFoot finds a different solution/approach, we still can improve in a future PR.

A very minor thing I noticed though, but don't consider this a blocking reason: when cycling through scratchpads, the focus jumps back to a tiled window for a fraction of a second (might indeed be my 2014 laptop though :frowning:).

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1211 @@
...
+            "Dead window in scratchpad found, discard: window PID: {}",
+            window
+        );
+    }
+
+    None
+}
+
+/// Check if the scratchpad is visible on the current tag.
+/// If the name of the scratchpad is incorrect, false will be returned
```
TornaxO7, line 173: > /// If the name of the scratchpad is incorrect (...)

What do you mean with "is incorrect"? Do you just mean "if the scratchpad is not visible"?

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1211 @@
...
+                    }
+                } else {
+                    // Window is hidden => show the scratchpad
+                    if let Err(msg) = show_scratchpad(manager, &window_handle) {
+                        log::error!("{}", msg);
+                        return Some(false);
+                    }
+                }
+
+                return Some(true);
```
TornaxO7, line 237: I think this can be summarized as:
```rust
    if let Some((is_visible, window_handle)) = manager.state
        .windows
        .iter()
        .find(|w| w.pid == Some(first_in_scratchpad))
        .map(|w| (w.has_tag(current_tag), w.handle))
    {
        let action_result = if is_visible {
            hide_scratchpad(manager, &window_handle)
        } else {
            show_scratchpad(manager, &window_handle)
        };

        match action_result {
            Ok(()) => Some(true),
            Err(msg) => {
                log::error!("{}", msg);
                Some(false)
            },
        }
```

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1211 @@
...
+    manager: &mut Manager<C, SERVER>,
+) -> Option<bool> {
+    // If `None`, replace with current window
+    let window_handle = window.or(manager
+        .state
+        .focus_manager
+        .window_history
+        .get(0)?
+        .as_ref()
+        .copied())?;
```
TornaxO7, line 277: Hm.... What do you think about
```rust
    let window_handle = {
        let current_window = manager
            .state
            .focus_manager
            .window_history
            .get(0)?
            .as_ref()
            .copied();

        window.or(current_window)?;
    };
```
?

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
TornaxO7: Hm... I'd say clippy to allow this in this block because it does make it more readable if this would be in a variable or what do you think?

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1211 @@
...
+            // Perform cycle
+            let front = scratchpad.pop_front()?;
+            scratchpad.push_back(front);
+        }
+        Direction::Backward => {
+            // Perform cycle
+            let back = scratchpad.pop_back()?;
+            scratchpad.push_front(back);
+        }
+    };
```
TornaxO7, line 488: I think we can make use of [`rotate_left`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.rotate_left) and [`rotate_right`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.rotate_right):
```rust
    match direction {
        Direction::Forward => scratchpad.rotate_left(1),
        Direction::Backward => scratchpad.rotate_right(1),
    };
```

leftwm-core/src/state.rs
```diff
@@ -23,7 +23,7 @@ pub struct State {
     pub mode: Mode,
     pub layouts: Vec<Layout>,
     pub scratchpads: Vec<ScratchPad>,
-    pub active_scratchpads: HashMap<String, Option<u32>>,
+    pub active_scratchpads: HashMap<String, VecDeque<u32>>,
```
TornaxO7, line 26: Would it be possible to create some new-types instead of using `String` and `u32`? Something like:
```rust
pub type TagName = String;
pub type ScratchPadId = u32;
// ...
pub active_scratchpads: HashMap<TagName, VecDeque<ScratchpadId>>,
```
(I don't know if those names are right but it would help to understand this attribute faster)

leftwm/src/config/keybind.rs
```diff
@@ -28,11 +28,36 @@ impl Keybind {
...
         config: &Config,
     ) -> Result<lefthk_core::config::Keybind> {
+        fn is_valid_scratchpad_name(config: &Config, scratchpad_name: &str) -> bool {
+            config
+                .scratchpad
+                .as_ref()
+                .and_then(|scratchpads| scratchpads.iter().find(|s| s.name == scratchpad_name))
+                .is_some()
+        }
+
```
TornaxO7, line 38: I'd move this function out because the logic of this function is only relevant for two base commands and not for the whole function `try_convert_to_lefthk_keybind`.

TornaxO7 (changes requested):
I left some suggestions :)

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
SamClercky: Incorrect as in not defined in the config. If the name is not in the config, it has no way of being initialized in the manager.

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1211 @@
...
+}
+
+/// Check if the scratchpad is visible on the current tag.
+/// If the name of the scratchpad is incorrect, false will be returned
+fn is_scratchpad_visible<C: Config, SERVER: DisplayServer>(
+    manager: &Manager<C, SERVER>,
+    scratchpad_name: &str,
+) -> bool {
+    // Like Try operator but returns false and only works on `Option`s
+    macro_rules! try_bool {
```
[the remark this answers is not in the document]
SamClercky: The function returns `true` if the SP is visible, `false` if not. The macro is a suggestion from AethanFoot to reduce code duplication. It functions as a shorthand return to handle the error cases the same way the `?`-operator is used for `impl Try`.

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1211 @@
...
+        ($cond:expr) => {
+            if let Some(value) = $cond {
+                value
+            } else {
+                return false;
+            }
+        };
+    }
+    let current_tag = try_bool!(manager.state.focus_manager.tag(0));
```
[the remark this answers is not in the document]
SamClercky: I like the idea of using build-in rust functions, but `unwrap_or` does not immediately return the outer function on failure, which the `try_bool!`-macro does.

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
```diff
@@ -0,0 +1,1211 @@
...
+    let scratchpad = try_bool!(manager.state.active_scratchpads.get(scratchpad_name));
```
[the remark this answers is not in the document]
SamClercky: Same as above

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
SamClercky: True, this looks cleaner

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
SamClercky: Didn't know this existed, will adjust :smile:

leftwm/src/config/keybind.rs
SamClercky: Where would you suggest to put this function?

SamClercky:
> Ok, when I git revert 4e48f7f everything is ok.
Maybe the manual `visibility` hanbling is actually helpful.

This is interesting. When trying on my machine with this commit reverted, I feel less jitter. Maybe the problem is in the delay between moving the SP to another tag (`NSP`) and the check in the loop which would normally set the visibility flag.

> A very minor thing I noticed though, but don't consider this a blocking reason: when cycling through scratchpads, the focus jumps back to a tiled window for a fraction of a second (might indeed be my 2014 laptop though :frowning:).

This is something I have noticed, but didn't think it would be considered a bug. The code for cycling, first hides the currently visible window and then shows the new window leaving a fraction of a second where there is no window visible. Could it be that the severity of the delay is caused by caching in `XLib` or `XCB`? I actually do not know enough about the internals of how X works to meaningfully measure this.

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
TornaxO7: Oh ok, that makes sense. I'd replace
> /// If the name of the scratchpad is incorrect, false will be returned

than with
```
/// Returns `false` if the scratchpad name isn't defined in the config
```

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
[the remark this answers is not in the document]
TornaxO7: Oh ok, I missunderstood the macro, I thought that `return false` would return to the variable.
Hm... and you've also tried this:
```rust
fn is_scratchpad_visible<C: Config, SERVER: DisplayServer>(
    manager: &Manager<C, SERVER>,
    scratchpad_name: &ScratchPadName,
) -> bool {
    if let Some(current_tag) = manager.state.focus_manager.tag(0) {
        if let Some(scratchpad) = manager.state.active_scratchpads.get(scratchpad_name) {
            // Filter out all the non existing windows (invalid pid) and map to window
            // Check if any of them is in the current tag
            return scratchpad
                .iter()
                .filter_map(|pid| manager.state.windows.iter().find(|w| w.pid == Some(*pid)))
                .any(|window| window.has_tag(&current_tag));
        }
    }

    return false;
```
?

TornaxO7 (changes requested):
Two little things

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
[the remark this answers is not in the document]
AethanFoot: I have to say I am not a massive fan of the double indentation, the current way is much more readable.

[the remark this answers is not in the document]
TornaxO7: Oh ok. I'd prefer a double-indentation than a macro, but I can live with it.

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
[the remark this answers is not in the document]
TornaxO7: discussion is resolved

leftwm-core/src/handlers/command_handler/scratchpad_handler.rs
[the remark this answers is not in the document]
TornaxO7: discussion is resolved

leftwm/src/config/keybind.rs
TornaxO7: This is a hard question.... I'd suggest to put it completely at the bottom of the file. :thinking:

VuiMuich:
> Strange you're not getting any logs from scratchpad related commands. Could it be because the log level was to low? I normally get the messages by starting `leftwm` on the command line and setting `DISPLAY` to the id I gave to `Xephyr`. Hope this helps :smile:

I was running this on a full on live session, as I had a couple of things lately that didn't work the same in Xephyr and a native session, and I was logging to systemd for convenience. I'll give it a shot to reproduce in Xephyr tomorrow.

VuiMuich:
@SamClercky just had an idea but can't verify right now:
I am using `focus_behaviou = ClickTo`, if you are on `Sloppy` that might explain why you can't reproduce my reported issues (at least some of them).
Also hobing to find time to do a quick screenrecording during the weekend.

TornaxO7 (approved):
LGTM now

VuiMuich:
Maybe we merge this now and open an issue to do some refactor to recreate https://github.com/leftwm/leftwm/commit/[4e48f7f](https://github.com/leftwm/leftwm/commit/4e48f7f15c281c1b5cfdfa9a9362006db51a098d)15c281c1b5cfdfa9a9362006db51a098d and at the same time avoid the mentioned regressions.

SamClercky:
@VuiMuich  

`ClickTo` did trigger the bug for me! I can confirm that with the 4e48f7f reverted, the bug is not visible anymore and that `Sloppy` focus behavior also does not trigger the bug.

Something not yet mentioned is that the bug for me triggers on an empty SP (`VecDeque` is empty), but not the first time. Every time I switch to a new tag, toggling works 1 time (show + hide) and then stops working like previously described.

I am ok with merging now and creating an open issue. Then other PR's won't be blocked by this one and I can try and start working on a new PR that fixes this last issue when I have some more time on my hands (school is restarting soon).

VuiMuich:
Thanks @SamClercky for all the work you put into this. Looking forward to any further contributions!
May I leave updating the docs and opening the issue for you?

Also thanks @TornaxO7 for your review, I guess next will be the logging PR.

PR #788, opened 2022-07-17
Refactor leftwm main

TornaxO7:
# Description

I refactored the `leftwm.rs` which should make it easier to read now (at least in my opinion).

Fixes #(issue)

## Type of change

- [x] Refactoring

No reason to update them.

# Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,6 +138,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "const_format"
+version = "0.2.26"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "939dc9e2eb9077e0679d2ce32de1ded8531779360b003b4a972a7a39ec263495"
+dependencies = [
+ "const_format_proc_macros",
+]
+
+[[package]]
+name = "const_format_proc_macros"
+version = "0.2.22"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef196d5d972878a48da7decb7686eded338b4858fbabeed513d63a7c98b2b82d"
+dependencies = [
+ "proc-macro2",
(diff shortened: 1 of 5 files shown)


VuiMuich (commented):
In line 236 the doc comment needs to reflect the renomer (line 272) of the former `reap` method.

> General question:
> Most of those methods of `pub struct Children` don't ever get called so, this seems to be a library thing, right?

Another little thing: maybe we could add some little expressions with some `unimplemented!()` and a `TODO` comment with the intention to later add a more dynamic process to find further subcommands in the `fn print_help_page()` and `fn is_subcommand(...)`. 
Besides that I like the set of `consts` to easy maintaining those items.

All in all some nice refactoring. Thanks!

TornaxO7:
> In line 236 the doc comment needs to reflect the renomer (line 272) of the former reap method.

So I just removed the comment. It should be relatively clear, that you can use the methods whenever you want so I don't think that it's necessary to have those comments.

> Another little thing: maybe we could add some little expressions with some unimplemented!() and a TODO comment with the intention to later add a more dynamic process to find further subcommands in the fn print_help_page() and fn is_subcommand(...).

Hm... I think that we should rather create an issue for this instead of adding a `TODO` or `unimplemented!` because:
- `unimplemented!` would raise a panic so the user might think first, that he/she reached a state which wasn't expected
- just adding a `TODO` can be easily ignored while a ticket/issue is more present in my opinion.

VuiMuich:
Oh, I totally forgot about the `panic` part of `unimplemented!` :flushed: 

Opening an issue, would be of course helpful to keep track. But one does not need to try very hard to ignore that either :stuck_out_tongue_winking_eye:

VuiMuich (approved):

PR #789, opened 2022-07-18
adding rust-toolchain.toml

TornaxO7:
# Description

Fixes #(issue)

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] This change requires a documentation update

# Checklist:

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "stable"


VuiMuich:
What do you think of adding `rust-toolchain.toml` to `.gitignore` so if someone wants to regularly build agains `nightly` or some specific version, or what ever can easily do without running into trouble when commiting, but we enforce `stable` as default.

TornaxO7:
Hm... I don't think that this would be useful, because it could happen that this person gets some errors on the nightly version while the stable works (ok, they could remove the `rust-toolchain.toml` but still...)

VuiMuich:
> Hm... I don't think that this would be useful, because it could happen that this person gets some errors on the nightly version while the stable works (ok, they could remove the `rust-toolchain.toml` but still...)

But that's basically the same as it is right now, isn't it? My reasoning would be that someone who decides to build with `nightly` against the default toolchain is aware enough of what they are doing to care to check if their commit runs well with `stable` as well.

TornaxO7:
Well, if you really insist in doing that, I can't complain, then.

PR #791, opened 2022-07-19
fix crash when serialising default config to toml

### LINKED ISSUE #692: leftwm panics on writing new default config.toml
Hi.
If I start leftwm with a missing config.toml (fresh install),leftwm is trying to write a new config file with default values, which fails with a panic

Error Msg.
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ValueAfterTable', leftwm/src/config/mod.rs:140:45
stack backtrace:
   0: rust_begin_unwind
             at /data/pkgmk/work/rust/src/rustc-1.57.0-src/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /data/pkgmk/work/rust/src/rustc-1.57.0-src/library/core/src/panicking.rs:100:14
   2: core::result::unwrap_failed
             at /data/pkgmk/work/rust/src/rustc-1.57.0-src/library/core/src/result.rs:1616:5
   3: leftwm::config::load
   4: leftwm_worker::main::{{closure}}
   5: leftwm_worker::main
```
leftwm-check output
```
:: LeftWM version: 0.3.0
:: LeftWM git hash: c244805
Configuration failed. Reason: No such file or directory (os error 2)
:: Checking environment . . .
    -> Environment OK (has XDG_RUNTIME_DIR) 
:: Checking theme . . .
```
The theme Error is expected because ./config/leftwm is empty

After diging around in leftwm/src/config/mod.rs I found a workaround 

 Patch:
```
--- leftwm/leftwm/src/config/mod.rs	2022-01-30 20:55:39.211958231 +0100
+++ leftwm-patch/leftwm/src/config/mod.rs	2022-01-30 22:29:32.664119177 +0100
@@ -98,9 +98,9 @@
     pub focus_new_windows: bool,
     pub keybind: Vec<Keybind>,
     pub state: Option<PathBuf>,
-    pub window_rules: Vec<WindowHook>,
 
     #[serde(skip)]
+    pub window_rules: Vec<WindowHook>,
     pub theme_setting: ThemeSetting,
 }
```

I'm not quiet sure if it is the root cause of the problem.

VuiMuich:
# Description

Fixes #692

This previously fixed issue was reintroduced by #765 by adding the default value for `sloppy_mouse_follows_focus` after the keybinds, which leads to a serialisation error with `toml` as it tries to add this value to the table of the last keybind entry.

There should be a test to avoid this in the future, but as this is somewhat critical for any new install I'd suggest to merge this asap and add the test later.

## Type of change

# Checklist:

### DIFF
--- a/leftwm/src/config/mod.rs
+++ b/leftwm/src/config/mod.rs
@@ -100,10 +100,11 @@ pub struct Config {
     pub disable_window_snap: bool,
     pub focus_behaviour: FocusBehaviour,
     pub focus_new_windows: bool,
+    pub sloppy_mouse_follows_focus: bool,
     pub keybind: Vec<Keybind>,
     pub state: Option<PathBuf>,
-    pub sloppy_mouse_follows_focus: bool,
-
+    // NOTE: any newly added parameters must be inserted before `pub keybind: Vec<Keybind>,`
+    //       at least when `TOML` is used as config language
     #[serde(skip)]
     pub theme_setting: ThemeSetting,
 }


stevensonmt:
Is this actually a bug in `toml` then that should be reported up to them?

VuiMuich:
Kind of, yeah.
But I guess it is somewhat systemic or some sort of design flaw: Toml has no explicit termination of tables, so when you put some value that ist supposed to be attached to the root-table behind some table in the notation like we use e.g. for keybinds:

```toml
[[some-table]]
Table-value1 = 1,
Table-value2 = 2

A-supposed-Root-table-value = false
```
All the following values are expected to be part of the last table. Luckily `serde` is smart enough to just ignore those values it has no fields to assign to, when deserialising.

I believe at one point I considered reporting this to `toml` or `toml-serde`, but learned about the above along the way and decided it probably would just be a "no fix".
And then we decided to move our config language to `ron` anyways.. so yesterday I spontaneously started some refactoring (#799) for this transition and turns out this might go along smoother then expected.

stevensonmt:
Why are `state` and `theme-setting` exempted from this issue?

VuiMuich:
Well, `theme_Setting` is explicitly skipped from `serde`.
And afaik is `state` also not serializing into `toml` (but in `ron` it does), but I would need to research, why this is. Maybe because it's default value is `None`?

PR #793, opened 2022-07-19
make `ron` the preferred conifg and theme language, add fallback to `toml`

VuiMuich:
# Description

This PR makes `RON` the preferred config and theming language.
If no `config.ron` or `theme.ron` are found their legacy `toml` counderparts are tried.
If no config at all is found the default config get serialized as `config.ron`.

For migration so far my best experience has been to let `leftwm-check` write a default file, by renaming/moving the `config.toml` so it is not found and then edit in the differences with both files side by side. Was less then 1/2 hour effort I guess.

~~TODO: Eventually a routine to automatically convert a valid `toml` config to `ron` should be added I guess. Probably as~~ There is a (temporary until toml is depraceted) command flag for `leftwm-check` now: `-m|--migrate-toml-to-ron`.

Here is an example `theme.ron`:
```rust
(
    border_width: 0,
    default_width: Some(400),
    default_height: Some(400),
    always_float: Some(true),
    margin: 5,
    workspace_margin: Some(5),
    default_border_color: "#222222",
    floating_border_color: "#005500",
    focused_border_color: "#FFB53A",
    on_new_window: Some("echo Hello World"),

    gutter: Some([Gutter (
        side: Top,
        value: 0,
        )]
    )
)
```

Working `RON` files can also be found here: https://github.com/leftwm/leftwm-contrib/tree/main/examples/config_ron

## Type of change

The update for the wiki is being prepared here and will be merged manually after this PR has been merged:
https://github.com/VuiMuich/leftwm_wiki/tree/ron_config

# Checklist:


VuiMuich:
:thinking: smh the theme does not load. will look into this tomorrow or thursday...

leftwm/src/bin/leftwm-check.rs
```diff
@@ -88,13 +92,22 @@ pub fn load_from_file(fspath: Option<&str>, verbose: bool) -> Result<Config> {
...
         }
 
+        #[cfg(feature = "toml-config")]
         let config = toml::from_str(&contents)?;
+        #[cfg(feature = "ron-config")]
+        let config = ron::from_str(&contents)?;
         Ok(config)
     } else {
         let config = Config::default();
+        #[cfg(feature = "toml-config")]
```
TornaxO7, line 102: A general suggestion: Would it make sense to abstract the way how we are handling config files? Because that would probably help to switch between config files. On the other hand, this would probably create some unnecessary code if we are switching only this time to another config-file type but it would definetely help, to avoid this `#[cfg(feature = "toml-config"]` for so many lines.

VuiMuich: Well, basically that arlready happend to some extent, with the librarification of `leftwm-core`, so it appears more or less the only places I had to touch where the actual interaction with the files, plus a couple of related tests.
I was surprised, how easy it actually was.

But if you have some ideas how to further generalise things, I am open to suggestions.

One thing that currently bugs me is that those feature-gates seem to fail, when building with `--all-targets --all-features`, which the CI does. Hence the failed checks on github.

AethanFoot: yeah, might want to change that CI now that we are adding mutually exclusive features. Could also add (incase both are added at the same time) an alternative path  ```#[cfg(all(feature = "toml-config", feature = "ron-config"))]
compile_error!("feature \"toml-config\" and feature \"ron-config\" cannot be enabled at the same time");``` or default to one over the other

TornaxO7: > But if you have some ideas how to further generalise things, I am open to suggestions.

I think that I could try to implement this. Like creating a trait for specifique config files.

## In general
I could imagine to have a directory in `leftwm-core` to have it like this:

`config/mod.rs`
```rust

pub fn get_config() {
    #[cfg(feature = "toml")]
    TomlConfigParser::new().get_config()

    #[cfg(feature = "yaml")]
    YamlConfigParser::new().get_config()
}

pub struct Config {
    /// we'd need to discuss which values are necessary for leftwm
}

pub trait ConfigParser {

    /// parses its config file. For toml it would be `~/.config/leftwm/config.toml` or for yaml `~/.config/leftwm/config.yaml` etc.
    /// Each parser would be a separate struct (see below)
    Config get_config(&self) -> Config;

    /// ...
}
````

`config/filetype/toml.rs`
```rust
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct TomlConfigParser {
    /// the toml fields
}

impl ConfigParser for TomlConfigParser {
    /// ...
}
```

`config/filetype/yaml.rs`
```rust
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct YamlConfigParser {
    /// the yaml fields
}

impl ConfigParser for YamlConfigParser {
    /// ...
}
```

VuiMuich: Awesome idea.

That would also be a great place to do some tweaking on the parser output as far as possible, like adding some basic doc comments and structuring to the output config file. And in case of toml try to make the parser use the compact syntax for keybinds etc.
TBH I have no idea, if the respectable serialisers have such formatting capabilities, just throwing ideas.

VuiMuich: Just stumbled over this possibly helpful fella: `toml::ser::tables_last` in the docs of the toml crate.

@TornaxO7 if you like to get into this, I guess a PR to this branch on my fork would probably be the easiest way to go, right?

TornaxO7: > @TornaxO7 if you like to get into this, I guess a PR to this branch on my fork would probably be the easiest way to go, right?

Theoretically yes, but I don't know how long it takes for me to do that since the university kicks in at the moment.
I'd say that you can merge that first and I'll create a PR afterwards.

VuiMuich: Ok then I guess I will work to solve the CI issues and call this review-ready.

TornaxO7: I'd close this because we already discussed on discord that this feature is not needed :)

VuiMuich:
:tada: finally I got all the clippy lints ironed out.

I am a bit confused  that `make test` and `make test-full` still miss some lints that the CI does, I guess we should address this so local testing can match the CI actually.

TornaxO7:
> I am a bit confused that make test and make test-full still miss some lints that the CI does, I guess we should address this so local testing can match the CI actually.

What? That's strange, I entered the clippy command to `test-full` could it be that you're using a different clippy version?

blackdragon2447:
currently if more than one config language is enabled leftwm will just always use the last one in the code since they'll just overwrite each other:
```rust
    #[cfg(any(feature = "toml-config", feature = "all-features"))]
    let _config_filename = path.place_config_file("config.toml")?;
    #[cfg(any(feature = "ron-config", feature = "all-features"))]
    let _config_filename = path.place_config_file("config.ron")?;`
```
I assume that that is not really intended behavior, should there be some way to switch which language is used of would it be a better idea to deny building with more then one language?

VuiMuich:
> currently if more than one config language is enabled leftwm will just always use the last one in the code since they'll just overwrite each other:

> I assume that that is not really intended behavior, should there be some way to switch which language is used of would it be a better idea to deny building with more then one language?

In an earlier commit I had the two languages with mutually exclusive `#[cfg(...)]` like this:
```rust
    #[cfg(all(feature = "toml-config", not(feature = "ron-config")))]
    let config_filename = path.place_config_file("config.toml")?;
    #[cfg(all(feature = "ron-config", not(feature = "toml-config")))]
    let config_filename = path.place_config_file("config.ron")?;
```
But this didn't match well with the CI.
On the other hand I had to bring so much `#[allow(...)]` overrides it might even work again with the above build config.

VuiMuich:
Would we want to change anything up? 
From my perspective this might be merge ready.

One thing I would like to add eventually is some commentary header like this:
```ron
//  _        ___                                      ___ _
// | |      / __)_                                   / __|_)
// | | ____| |__| |_ _ _ _ ____      ____ ___  ____ | |__ _  ____    ____ ___  ____
// | |/ _  )  __)  _) | | |    \    / ___) _ \|  _ \|  __) |/ _  |  / ___) _ \|  _ \
// | ( (/ /| |  | |_| | | | | | |  ( (__| |_| | | | | |  | ( ( | |_| |  | |_| | | | |
// |_|\____)_|   \___)____|_|_|_|   \____)___/|_| |_|_|  |_|\_|| (_)_|   \___/|_| |_|
// A WindowManager for Adventurers                         (____/
// For info about configuration please visit https://github.com/leftwm/leftwm/wiki
```
But this can easily wait for a later PR

blackdragon2447:
Would it be better for readability and/or preventing bugs if instead of having the `#[cfg(...)]` all throughout the code one would have a file/config handler module which does all the config language dependent stuff (assuming that leftwm would add support for more languages)?

filehandler/mod.rs would look something like this in that case:
```rust
#[cfg(feature = "guess-config")]
mod guess;
#[cfg(feature = "guess-config")]
pub use guess::generate_new_config;
#[cfg(feature = "guess-config")]
pub use guess::get_config_file;
#[cfg(feature = "guess-config")]
pub use guess::load;
#[cfg(feature = "guess-config")]
pub use guess::load_from_file;
#[cfg(feature = "guess-config")]
pub use guess::save_to_file;

#[cfg(feature = "ron-config")]
mod ron;
#[cfg(feature = "ron-config")]
pub use self::ron::generate_new_config;
#[cfg(feature = "ron-config")]
pub use self::ron::get_config_file;
#[cfg(feature = "ron-config")]
pub use self::ron::load;
#[cfg(feature = "ron-config")]
pub use self::ron::load_from_file;
#[cfg(feature = "ron-config")]
pub use self::ron::save_to_file;
```
And the rest of the code can just use `filehandler::load` for example independent of the config language(which is how it currently is in leftwm-config).

blackdragon2447:
Ah, missed that, sorry.

VuiMuich:
Updated OP according to the latest commits turning this PR completely around...

leftwm/src/bin/leftwm-check.rs
```diff
@@ -279,6 +303,10 @@ fn check_theme_toml(filepath: PathBuf, verbose: bool) -> Result<PathBuf> {
                 if verbose {
                     println!("The theme file looks OK.");
                 }
+                println!(
+                    "\x1b[1;93mWARN: TOML as config format is about to be deprecated.
```
TornaxO7, line 307: this `\x1b[1;93` is for being printed out with a color right? Why not choosing a crate for doing this?

TornaxO7, line 307: this `\x1b[1;93` is for being printed out with a color right? Why not choosing a crate for doing this?

TornaxO7: Hm... I'm seeting that this is done in the other parts of this binary as well, I think an extra issue for this might be suitable.

TornaxO7 (commented):
Added some thoughts. What do you think about them?

leftwm/src/bin/leftwm-check.rs
VuiMuich: Also not sure if its worth pulling another crate for this niche use case. I tend to keep deps tight.

TornaxO7: > I tend to keep deps tight.

May I ask wht the reason is?

VuiMuich: I guess a slight tendency for minimalism. Not as dogmatic or strickt as the suckless guys, but you definitely want to keep an eye on the dep count.
If you carelessly add crates you easily might bloat your project. Maybe not the resulting binary size, but at least cargo updates and compilation time.

TornaxO7: > If you carelessly add crates you easily might bloat your project. Maybe not the resulting binary size, but at least cargo updates and compilation time.

I understand this point, but on the other hand I think that would be in contrast to the idea of having crates which should help you to reinvent the wheel.

VuiMuich:
I worked on updating the docs: https://github.com/VuiMuich/leftwm_wiki/tree/ron_config
If someone might like to have a proofread, that would be nice.

VuiMuich:
Eventually we should consider refactoring all the file handling to a dediceted `utility` module, but this is for anothe day and another PR.

hertg (approved):
Just looking through the code, everything looks good to me

TornaxO7 (approved):
Looks fine to me

leftwm/src/bin/leftwm-check.rs
```diff
@@ -41,6 +47,20 @@ async fn main() -> Result<()> {
         "\x1b[0;94m::\x1b[0m LeftWM git hash: {}",
         git_version::git_version!(fallback = option_env!("GIT_HASH").unwrap_or("NONE"))
     );
+    if matches.occurrences_of("migrate") >= 1 {
+        println!("\x1b[0;94m::\x1b[0m Migrating configuration . . .");
+        let path = BaseDirectories::with_prefix("leftwm")?;
+        let ron_file = path.place_config_file("config.ron")?;
+        let toml_file_path = path.place_config_file("config.toml")?;
```
AethanFoot, line 54: This is just a little OCD, can these be named similar (e.g. with _path or not) as they are the same (except the ron and toml)

AethanFoot (approved):
Looks good!

leftwm/src/bin/leftwm-check.rs
VuiMuich: did it the other way round and removed the `let toml_file_str` by moving the function calls to the variable move in the call of `load_from_file`

VuiMuich:
> Looks good! I will add is it worth adding the implicit_some automatically for generated files.

This should be done in the `PrettyConfig` in `write_to_file` in line 137. Or did you mean something else?

AethanFoot:
Oh, does PrettyConfig automatically use the implicit_some extension

VuiMuich:
> Oh, does PrettyConfig automatically use the implicit_some extension

No, but it has it as a config option:
```rust
    let ron_pretty_conf = ron::ser::PrettyConfig::new()
        .depth_limit(2)
        .extensions(ron::extensions::Extensions::IMPLICIT_SOME);
```

leftwm/src/bin/leftwm-check.rs
```diff
@@ -76,29 +95,65 @@ pub fn load_from_file(fspath: Option<&str>, verbose: bool) -> Result<Config> {
...
+            let ron_file =
+                BaseDirectories::with_prefix("leftwm")?.place_config_file("config.ron")?;
+            let toml_file =
+                BaseDirectories::with_prefix("leftwm")?.place_config_file("config.toml")?;
+            if Path::new(&ron_file).exists() {
+                ron_file
+            } else if Path::new(&toml_file).exists() {
+                println!(
+                    "\x1b[1;93mWARN: TOML as config format is about to be deprecated.
+      Please consider migrating to RON or contact the theme creator about this topic.\x1b[0m"
```
mautamu, line 108: Should probably change this to "Please consider migrating config.toml using leftwm-check -m" owtte

mautamu (approved):
Testing ~/.config/leftwm/config.{toml,ron} (from existing leftwm instance):
- [x]  Building by checking out 1052a8e works 
- [x] `leftwm-check` sees config.toml
- [x]  Soft reload works and produces same result on 1052a8e as master
- [x] `leftwm-check -m` generates RON
- [x] `leftwm-check -v` diffs look alright between RON and TOML; Diff:
```diff
4,5d3
< WARN: TOML as config format is about to be deprecated.
<       Please consider migrating to RON or contact the theme creator about this topic.
```
- [x] Soft reload works and produces same result as TOML

Testing ~/.config/leftwm/themes/current/theme.{toml,ron}
- [x] Error if theme.toml exists
- [x] Ok if theme.ron exists
- [x] Theme changes (MUST UPDATE THE UP SCRIPT!)

VuiMuich:
Thanks @mautamu this was a very nice review checklist.

Despite the copy paste mistake in the `config` migration message I added a note to remind for the necessary change in the `up` script.

PR #798, opened 2022-07-22
Move `TestConfig` to `tests` module

TornaxO7:
# Description

Since this struct seems be only used for `tests`, I thought that I could move it to the `tests` module then.

## Type of change

Not needed

# Checklist:


AethanFoot:
I think at some point we should move tests to labelled files in their own directory to make them easier to manage and find (then tests for specific bugs can easily be labelled). 

Looks good, Thanks

PR #800, opened 2022-07-24
Update README: Polybar no longer requires an AUR helper on arch

mfdorst:
Polybar is [in the archlinux community repository now][1], so there's no need to use `paru` or `yay` to install it.

[1]: https://archlinux.org/packages/community/x86_64/polybar/


VuiMuich:
Thanks!

PR #805, opened 2022-07-26
Add urgent flag to tags

jmattheis:
# Description

X windows can set the urgency hint to notify that something has happened, like a new message in a chat program.

xprops look like this:

    WM_HINTS(WM_HINTS):
        The urgency hint bit is set
    _NET_WM_DESKTOP(CARDINAL) = 8
    WM_STATE(WM_STATE):
        window state:
        icon window: <field not available>
    ...

This change allows the user of leftwm to highlight tags that have urgent windows. i3 has a similar setting called urgent_workspace to highlight urgent workspaces. https://i3wm.org/docs/userguide.html#_colors

urgent

## Type of change

[this page](https://github.com/leftwm/leftwm/wiki/Customizing-Themes-with-%60liquid%60-templates) has to be updated. Add this line:
```
"urgent"      <boolean is true, when the tag has at least one window with the urgent hint>
```

# Checklist:


leftwm-core/src/display_servers/xlib_display_server/event_translate_property_notify.rs
```diff
@@ -30,16 +30,10 @@ pub fn from_event(xw: &XWrap, event: xlib::XPropertyEvent) -> Option<DisplayEven
         xlib::XA_WM_NORMAL_HINTS => {
             build_change_for_size_hints(xw, event.window).map(DisplayEvent::WindowChange)
         }
-        xlib::XA_WM_HINTS => match xw.get_wmhints(event.window) {
-            Some(hints) if hints.flags & xlib::InputHint != 0 => {
-                let handle = event.window.into();
-                let mut change = WindowChange::new(handle);
-                change.never_focus = Some(hints.input == 0);
-                Some(DisplayEvent::WindowChange(change))
```
jmattheis, line 38: The old code would never update never_focus if the window lost the InputHint. I've fixed this in this PR too.

VuiMuich:
Mostly looking good.

I was testing with `xdotool search --name "man xprop" set_window --urgency 1` (obviously having a terminal window with `man xprop` as an pseudo identifier for `xdotool` on some hidden tag). The only "hickup" I could see is, that the hint – as far as I was able to reproduce – only gets reset if the `main window` changes its focus state.
I think it should unset, on making the relevant `tag` visible.
Maybe a nice bonus might be to set focus automatically to the `urgent` window, idk.

jmattheis:
> I think it should unset, on making the relevant tag visible.

I think no other window manager does this. On i3, sway, and herbstluftwm the urgent flag only gets unset if the window is manually focused. herbstluftwm has a utility to jump to the urgent window directly via `herbstclient jumpto urgent`. I think something like this would be useful for leftwm, but it should be done in a separate PR.

VuiMuich:
Thanks for this nice addition!

PR #821, opened 2022-08-31
improve logging

TornaxO7:
# Description

Replacing `slog` with `env_logger`.

## Type of change

Not needed.

# Checklist:


TornaxO7:
Hm... I don't receive any logs yet, I need some time to investigate.

TornaxO7:
Alright, it should work now!

leftwm/src/bin/leftwm-worker.rs
```diff
@@ -25,78 +28,12 @@ fn main() {
...
-
-    let logger = slog::Logger::root(slog_async::Async::default(envlogger).ignore_res(), o!());
-
-    slog_stdlog::init().unwrap_or_else(|err| {
-        eprintln!("failed to setup logging: {}", err);
-    });
+    let log_file = CACHER.get_file(PathBuf::from(LOGGING_FILE)).unwrap();
 
-    slog_scope::set_global_logger(logger)
+    Builder::from_env(Env::default().default_filter_or("debug"))
```
TornaxO7, line 36: Is it fine, if we use `debug` as the defualt log-level?

leftwm/src/utils/cacher.rs
```diff
@@ -0,0 +1,58 @@
+use std::{fs::File, path::PathBuf};
+use xdg::{BaseDirectories, BaseDirectoriesError};
+
+lazy_static! {
```
VuiMuich, line 4: Is there possibly a way to avoid introducing this new dependency?

VuiMuich (commented):
I would like to keep the possibility to log to `journald`. For developing logging to stdout or stderr is fine, but the whole idea of #508 was to make logging to journald the default for packages like in the AUR so troubleshooting for 'non dev users' gets another valuable tool.
Maybe adding `libsystemd` under the `journald` feature flag might be good, but I have no idea how big the effort is to add the target: https://docs.rs/libsystemd/0.4.1/libsystemd/logging/index.html

Do you see any further use for the `cacher` despite logging? I would suggest to refactor anything related to a `leftwm/src/logging.rs` module. And IMO that would be the place for the `cacher` if no other place might use it.

And finally, I don't want to generally block now dependencies, but I would like to have at least a little discussion about the necessity, to have some references for devs in the future.

TornaxO7:
> Do you see any further use for the cacher despite logging?

Yes I do! Currently we are using `/tmp` for the down script for example or the state, which isn't that good in my opinion, because everyone can read it which might be unwanted for some users.

leftwm/src/utils/cacher.rs
TornaxO7: Hm... I thought that having a singleton would be the best since it would be kinda weird to have multiple "cachers". So the answer is actually no sadly.... but I'll ask my friends if they've got any ideas.

VuiMuich:
> > Do you see any further use for the cacher despite logging?
> 
> Yes I do! Currently we are using `/tmp` for the down script for example or the state, which isn't that good in my opinion, because everyone can read it which might be unwanted for some users.

Valid point. :+1:

TornaxO7:
> Plus how would you access this cacher from a script?

Oh right.... I had in my head, that `leftwm` takes care of the themes.... damn...

VuiMuich:
> Hm... I understand this point but couldn't it be a disadvantage for non-systemd users? I'll rethink the logging system a bit

Absolutely, so the `journald` logging should not necessary be exclusive of logging to stdout or a logfile.
Ideally there would be a feature-flag that builds with `journald` and IMO this should be included in `default`.
Packaging scripts might check if `systemd` is being used and if not (or in case of the `void` package) build with `--no-default-features`.

leftwm/src/bin/leftwm-worker.rs
```diff
@@ -10,7 +10,7 @@ fn main() {
 
     log::debug!("starting leftwm-worker");
 
-    let exit_status = panic::catch_unwind(|| {
+    let completed = panic::catch_unwind(|| {
```
VuiMuich, line 13: I guess this was inadvertedly reverted, right?
```suggestion
    let exit_status = panic::catch_unwind(|| {
```

leftwm/src/bin/leftwm-worker.rs
```diff
@@ -1,11 +1,14 @@
...
+
+const LOGGING_FILE: &str = "leftwm.log";
 
 fn main() {
-    //let _log_guard = setup_logfile();
-    let _log_guard = setup_logging();
-    log::info!("leftwm-worker booted!");
+    setup_logger();
+
+    log::debug!("starting leftwm-worker");
```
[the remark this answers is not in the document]
VuiMuich: I would be ok with setting the debug level in `dev` builds and info in `release` ones

AethanFoot:
For logging to journald (I am not 100% if this works), but could we use syslog. This then in theory would log to any syslog daemon, e.g. for void https://docs.voidlinux.org/config/services/logging.html .

TornaxO7:
> For logging to journald (I am not 100% if this works), but could we use syslog.

Syslog looks pretty promising for unix-based systems but it looks like that it doesn't create an opportunity for journald according to [this list](https://docs.rs/syslog/6.0.1/syslog/fn.unix.html). We could use the [`unix`](https://docs.rs/syslog/6.0.1/syslog/fn.unix.html) function which uses the `/var/run/log`.

TornaxO7:
Hm... or should be keep using `slog` but with a nice little wrapper to simplify the usage with it?

TornaxO7:
@VuiMuich @AethanFoot

AethanFoot:
I think journald picks up logs from those files (if you use `logger hello` it prints to journalctl, and logger uses syslog, same with `python -c 'import syslog; syslog.syslog("Hello World")'`). Yeah we could create a simpler use, as slog has an env_logger and syslog extensions.

AethanFoot:
We do need to overhaul error handling and messages at some point as well, which is related to this (if we make it its own crate at some point, lefthk can use it as well)

TornaxO7:
> Yeah we could create a simpler use, as slog has an env_logger and syslog extensions.

What do you mean with this? So it's fine if we keep slog?

AethanFoot:
Yeah we could, or we could the env_logger with a syslog crate. Either way would be okay.

TornaxO7:
@AethanFoot @VuiMuich A general question how the logging should work: Should the user be able to log to multiple places simultaneously like for example to journald + a log file at the same time or should we only enable logging only for *one* "type"?

AethanFoot:
I feel that for most users logging to journald (or their relevant syslog thing) should be default. Then logging to file can be enable for someone doi g dev work and disabling logging through features for users minimising.

TornaxO7:
@AethanFoot yes, but what's with logging to multiple places? Do think that's needed? Like logging to journald + to a file?

AethanFoot:
I would say yes, however we shouldnt need to worry too much if the two options are setup separately for their purpose. If they enable logging to file then yes both should be enabled. If they do nothing just syslog. And they can disable completely. (Sorry just trying to layout the cases clearly)

TornaxO7:
It would be nice if someone could test the logging with `syslog` and `journald` because I'm using `artix` without systemd (and so journald). I don't really understand how to wrtie to `/var/log/syslog.log`, I even tried to write with C into it:
```c
#include <syslog.h>

int main() {
    openlog("Artix btw", LOG_PID, LOG_USER);
    syslog(LOG_INFO, "YEEEEEEEEEE");
    closelog();
}
```
but that didn't work either.

The file works fine at my last test.

VuiMuich:
Just a quick feedback, a more in-depth review might follow this evening:
- Personally I'd prefer the inline imports over the imports over the whole file for mainly the reason, that with the inline imports it is easy to rip-grep for all "logging" calls no matter their level in one go
- I merged this over the `lefthk` branch, the merge conflicts are easy to solve and it runs well so far. It logs to journald, by default only at level info, which is fine but I don't see yet how to raise verbosity. Maybe we should besides later documenting this in the wiki leave a doc comment at the setup-logging invocation in `leftwm-worker` how to do this

Besides that great work and awesome find with the `tracing` crate, iirc the previous crate count we are now down by ten crates in total.

leftwm/src/bin/leftwm-worker.rs
```diff
@@ -5,7 +5,6 @@ fn main() {
     leftwm::utils::log::setup_logging();
 
     tracing::info!("leftwm-worker booted!");
-    tracing::debug!("bitte");
```
VuiMuich: I loved how you could see the struggle that went into getting there :heart_eyes:

TornaxO7:
> Personally I'd prefer the inline imports over the imports over the whole file for mainly the reason, that with the inline imports it is easy to rip-grep for all "logging" calls no matter their level in one go

So should I change it back @VuiMuich ?

VuiMuich:
@0323pin my I invite you to test this branch built with `cargo b --release --no-default-features -features=lefthk,sys-log`
Also would you see any issues with bumping msrv to `1.57.0`?

VuiMuich:
this closes #508

0323pin:
@VuiMuich Sure, no worries.
Ok, if I build this tomorrow?

0323pin:
@VuiMuich 
> Also would you see any issues with bumping msrv to 1.57.0?

Nope, NetBSD is on 1.62.1 and I'm on 1.63.0

VuiMuich:
> @VuiMuich Sure, no worries. Ok, if I build this tomorrow?

Sure



> @VuiMuich
> 
> > Also would you see any issues with bumping msrv to 1.57.0?
> 
> Nope, NetBSD is on 1.62.1 and I'm on 1.63.0

Nice. I think we still bump to the minimum we need to suffice our deps.

0323pin:
Ok! I'll give this a go tomorrow.

I hope none of you guys take this wrong but, ...
please, **no** systemd and **no** polkit. I'm fine with dbus but, that's as far as I can go.

VuiMuich:
> Ok! I'll give this a go tomorrow.
> 
> 
> 
> I hope none of you guys take this wrong but, ...
> 
> please, **no** systemd and **no** polkit. I'm fine with dbus but, that's as far as I can go.

Of course we don't intend to only support systemd, we just want it as default for systems running on systemd to be able to better troubleshoot user issues.

For users on non-systemd systems the 'default' should be syslog and for development we keep logging to a file around as an option.

Right now we do not automatically default to syslog if systemd is not available. AFAIK cargo is not able to do such checks and logic, but the Makefile should and package maintainers ofc can set those flags for their systems packages.

0323pin:
@VuiMuich Just built from 3bbbc94843294f4af15c2a5ee51eb973e57d4e76, all good.
Just need to convert my config to `ron`

> package maintainers ofc can set those flags for their systems packages

Yes, I need to look into this though.

0323pin:
@VuiMuich Oh! How do I assign the old layout_mode to Tag instead of workspace in the new ron config file?

VuiMuich:
I have this (giving layouts and scratchpad for context):
```rust
layouts: [
    MainAndDeck,
    Monocle,
    CenterMainFluid
],
layout_model: Tag,

scratchpad: <...>
```

0323pin:
@VuiMuich Yeah, I found it.

What about theme.toml? Should that also be converted to ron?
I have a very simple theme.toml file but, how to convert this to ron?

```
# leftwm theme.toml

border_width          = 0
margin                = 4
default_border_color  = "#282828"
floating_border_color = "#C0C5CE"
focused_border_color  = "#C0C5CE"
```

0323pin:
Fixed 😄 

```
(
    border_width: 0,
    margin: 4,
    default_border_color: "#282828",
    floating_border_color: "#C0C5CE",
    focused_border_color: "#C0C5CE",
)
```

VuiMuich:
> border_width          = 0
> margin                = 4
> default_border_color  = "#282828"
> floating_border_color = "#C0C5CE"
> focused_border_color  = "#C0C5CE"

This should do:
```rust
#![enable(implicit_some)]
(
border_width:          0,
margin:                4,
default_border_color:  "#282828",
floating_border_color: "#C0C5CE",
focused_border_color:  "#C0C5CE",
)
```

0323pin:
> I guess a migration guide for themes should be added to the wiki as well.

Indeed.
But, I got it working anyway.

0323pin:
> I guess the mention of polket was just a precaution, as I don't remember any discussion about it so far, right?

Yes, it was just a precaution.

0323pin:
@VuiMuich and @TornaxO7 Feel free to merge this whenever you guys think it's ready.
I have made up my mind on _how  to_ switch-off `journald-log`.

TornaxO7:
> I have made up my mind on how to switch-off journald-log.

@0323pin so you understood how to change it?

0323pin:
@TornaxO7 Yes, I'm running a build from your branch right now, no worries.
I maintain about 80 rust packages on NetBSD, so I need to find my way around things like this 😄 

Whenever you guys are ready, I'm ready.

TornaxO7:
@VuiMuich @AethanFoot @hertg Review can start I think (I guess only 2 reviews are fine as well?)

TornaxO7:
@0323pin you can review it as well, if you want to of course :)

0323pin:
@TornaxO7 Thanks for the offer :)

> Right now we do not automatically default to syslog if systemd is not available. AFAIK cargo is not able to do such checks and logic, but the Makefile should and package maintainers ofc can set those flags for their systems packages.

My first though was to use `[target.'cfg(target_os = "linux")'.dependencies]` and `[target.'cfg(all(unix, not(target_os = "linux")))'.dependencies]` in `Cargo.toml`. But, this is ofc not possible as it would exclude all non-systemd linux systems.

So, I will just do,

```
CARGO_NO_DEFAULT_FEATURES=	YES
CARGO_FEATURES+=		lefthk sys-log
```

If someone is using pkgsrc on a systemd-linux system, ... well, bad luck.

VuiMuich:
> My first though was to use `[target.'cfg(target_os = "linux")'.dependencies]` and `[target.'cfg(all(unix, not(target_os = "linux")))'.dependencies]` in `Cargo.toml`. But, this is ofc not possible as it would exclude all non-systemd linux systems.
> 
> 
> 
> So, I will just do,
> 
> 
> 
> ```
> 
> CARGO_NO_DEFAULT_FEATURES=	YES
> 
> CARGO_FEATURES+=		lefthk sys-log
> 
> ```

Fair enough. 

> If someone is using pkgsrc on a systemd-linux system, ... well, bad luck.

Well in that case that's just another quest in leftwm-adventure-land then 😁

TornaxO7:
Well, why aren't we using [`build.rs`](https://doc.rust-lang.org/cargo/reference/build-scripts.html) then? We could simply use rusts `cfg(windows)`, `cfg(unix)` etc. to set the feature-flags.

0323pin:
> Well, why aren't we using [build.rs](https://doc.rust-lang.org/cargo/reference/build-scripts.html) then? We could simply use rusts cfg(windows), cfg(unix) etc. to set the feature-flags.

Hmm ... https://doc.rust-lang.org/reference/conditional-compilation.html says:
```
target_os

Example values:

"windows"
"macos"
"ios"
"linux"
"android"
"freebsd"
"dragonfly"
"openbsd"
"netbsd"
```

How would you define Linux with systemd and Linux without systemd?

0323pin:
Ok, been running a build from 5fcb5f6a5d92684c49175be81f793c39fad1945f for the last two days, no issues found.

VuiMuich:
> Ok, been running a build from [5fcb5f6](https://github.com/leftwm/leftwm/commit/5fcb5f6a5d92684c49175be81f793c39fad1945f) for the last two days, no issues found.

Great. Thanks for testing!

TornaxO7:
May I ask what we're waiting for? :eyes:

VuiMuich:
We thought we'd merge #768 first and solve the resulting conflicts here, rather then the other way round and we almost merged the other two days ago, but then a few regressions put that on hold.

If you want to proceed to the next topic, don't feel obligated to wait for the merge of this here. The expected conflicts are not expected to be something major, we just wanted to spare the other contributor the chores of another 'merge main', as their branch had to undergo this already a couple of times.

TornaxO7:
Understandable! Thank you for the information!

VuiMuich:
btw:
Closes #508

VuiMuich (approved):

PR #825, opened 2022-09-04
add new-issues workflow

hertg:
**Description**
Adds a workflow that automatically adds all new issues and PRs to the roadmap project board.


hertg:
just tested, works great

PR #826, opened 2022-09-04
Issue forms

hertg:
This changes the `bug_report` issue template to an actual bug report form. In addition, two other issue forms are added for `Request a new feature` and `Enhance an already existing feature`. In the form selection, the Github Discussions tab and the Discord Server are promoted also.

> *Screenshot*
> image

On all forms, there is a notice that we recommend the discussions tab for questions and unspecific proposals.

---

While defining the forms, I used a personal public repository to preview what it looks like, see [here](https://github.com/hertg/gh-issue-templates/issues/new/choose).

- [Bug Report](https://github.com/hertg/gh-issue-templates/issues/new?assignees=&labels=bug&template=bug_report.yml)
- [Enhance an already existing feature](https://github.com/hertg/gh-issue-templates/issues/new?assignees=&labels=enhancement&template=feature_enhancement.yml)
- [Request a new feature](https://github.com/hertg/gh-issue-templates/issues/new?assignees=&labels=feature&template=feature_request.yml)

---

I made as few fields required as possible so we don't scare away people from actually reporting bugs and feature requests. Still, the bug report form has more fields and requires more information (ie. OS/Distro and the output of `leftwm-check`).

Both the feature proposal and enhancement proposal form only contain one markdown description field, but at the bottom also require the user to specify how important the feature is to them. I think this can provide us with some additional information to be considered when prioritizing and planning issues. The second dropdown about "how important they feel the feature is to other users" provides a kind of "reality check" that helps us in interpreting their selection of those answers.

---

What do you guys think about it? I feel that it could be a bit overwhelming, but on the other hand, we want users to file good reports... Opening a blank issue is still possible, this might be preferable in some circumstances. Especially when we ourselves want to open quick issues for ongoing things.

@VuiMuich @AethanFoot @mautamu 
Let me know if you have any feedback.


hertg:
As a side note:
If the bug report form is too much, we can also just shorten it back to one single textarea and populate it with the same content as the bug report has right now. Then the only difference to the current issue template would be the new notice above which may steer some users to the discussions.

VuiMuich (commented):
I love it.

The only thing that might be nice if it would be possible to have a placeholder in the `Title` to remind people to be brief but descriptive.

I like `config copy.yml` better :smile:

AethanFoot (approved):
Looks awesome!!

VuiMuich (approved):

PR #832, opened 2022-09-09
Refactor leftwm crash management

TornaxO7:
# Description

Currently `leftwm` just creates a new session if the current session exits which might be undesired because we are currently ignoring the reason *why* the session exited. The session might crash if the worker paniced which means that there's an error/a bug in the code of `leftwm`.

A little more context: I noticed this issue when I tried to run my logger. It crashed but leftwm kept creating new sessions.

## Type of change

Not needed

# Checklist:

- [X] Wiki pages have been updated accordingly (to perform **after** merge)


TornaxO7:
Eh, I don't understand the failing check o.O @AethanFoot @VuiMuich @hertg  can anyone of you explain what I'm doing wrong?

hertg:
> Eh, I don't understand the failing check o.O @AethanFoot @VuiMuich @hertg can anyone of you explain what I'm doing wrong?

You are not doing anything wrong, you can ignore the issue.
Problem is that the token used in this workflow is not passed to the action because this PR comes from a fork. But it is an error anyway that this action tries to run on PRs. I'll fix that later :)

hertg:
I just removed it

TornaxO7:
> I just removed it

Is it possible that you restart my checks? Because otherwise I need to look for some changes to push again.

VuiMuich:
> > I just removed it
> 
> Is it possible that you restart my checks? Because otherwise I need to look for some changes to push again.

Just tried that, I guess this might just stick with this PR, unless there is a way to remove checks manually I am failing to find.
Sorry but we might just have to ignore this failed check.

TornaxO7:
> Just tried that, I guess this might just stick with this PR, unless there is a way to remove checks manually I am failing to find.
Sorry but we might just have to ignore this failed check.

could it be that I have to pull from upstream?

TornaxO7:
I just saw that I forgot to test it. I'll converting it back from a draft to a PR if I run it a bit.

leftwm/src/bin/leftwm.rs
```diff
@@ -149,6 +158,8 @@ fn start_leftwm() {
             }
         }
 
+        no_error_occured = evaluate_session(&mut leftwm_session) != SessionStatus::Error;
```
VuiMuich, line 161: You have a double negative logic here. Wouldn't it be nicer to read with:
```suggestion
        error_occured = evaluate_session(&mut leftwm_session) == SessionStatus::Error;
```

leftwm/src/bin/leftwm.rs
```diff
@@ -159,10 +170,14 @@ fn start_leftwm() {
             std::thread::sleep(delay);
         }
     }
+
+    if !no_error_occured {
```
VuiMuich, line 174: follow up from above:
```suggestion
    if error_occured {
```

TornaxO7:
Alright, works fine! @AethanFoot @hertg @VuiMuich ready for review!

leftwm/src/bin/leftwm.rs
```diff
@@ -138,9 +146,10 @@ fn start_leftwm() {
 
     let flag = get_sigchld_flag();
 
-    loop {
+    let mut no_error_occured = true;
+    while no_error_occured {
```
VuiMuich, line 150: Ofc at some point we need to have some negative logic:
```suggestion
    let mut error_occured = false;
    while !error_occured {
```

leftwm/src/bin/leftwm.rs
```diff
@@ -196,3 +211,22 @@ fn get_sigchld_flag() -> Arc<AtomicBool> {
...
+fn evaluate_session(leftwm_session: &mut Child) -> SessionStatus {
+    if let Ok(exit_status) = leftwm_session.wait() {
+        if exit_status.success() {
+            SessionStatus::Success
+        } else {
+            SessionStatus::Error
+        }
+    } else {
+        SessionStatus::DidNotStart
+    }
```
VuiMuich, line 224: Can this be refactored in a match case with `Ok() if ..`
How are we acting on `DidNotStart` anyways? We maybe want to add some commentary/doc about the StessionStatus and its implications.

leftwm/src/bin/leftwm.rs
```diff
@@ -196,3 +211,22 @@ fn get_sigchld_flag() -> Arc<AtomicBool> {
...
+}
+
+fn print_crash_message() {
+    println!("Leftwm crashed due to an unexpected error.");
+    println!("Please create a new issue and post its log if possible.");
+    println!();
+    println!("NOTE: You can restart leftwm with `startx`.");
```
VuiMuich, line 231: Can you please refactor to a single invocation of `println!` macro?

VuiMuich (changes requested):

TornaxO7:
I adjusted the code a little bit @VuiMuich

TornaxO7:
@VuiMuich sorry, `no_error_occured` is now refactored as well!

VuiMuich (approved):

VuiMuich:
One thing I wanted to ask but forgot, as I can't test run this right now: does this return the exit codes, so the terminal recieve those?

TornaxO7:
> One thing I wanted to ask but forgot, as I can't test run this right now: does this return the exit codes, so the terminal recieve those?

No the `leftwm.rs` binary just looks

VuiMuich:
> > One thing I wanted to ask but forgot, as I can't test run this right now: does this return the exit codes, so the terminal recieve those?
> 
> 
> 
> No the `leftwm.rs` binary just evaluates the exit-status from `leftwm-worker` and exits normally (with code 0), but I can add this as well if you want.

Yes I think if we exit on error we should actually exit with 1

TornaxO7:
> > > One thing I wanted to ask but forgot, as I can't test run this right now: does this return the exit codes, so the terminal recieve those?
> > 
> > 
> > No the `leftwm.rs` binary just evaluates the exit-status from `leftwm-worker` and exits normally (with code 0), but I can add this as well if you want.
> 
> Yes I think if we exit on error we should actually exit with 1

Or should `leftwm` even return the exit code from the session?

VuiMuich:
Yes, when we exit because the worker exited on error, the main bin should not hide that and also exit on 1, imo.

TornaxO7:
> Yes, when we exit because the worker exited on error, the main bin should not hide that and also exit on 1, imo.

But couldn't it happen that the worker exited with even another code like 2 or 17 (for whatever reason)?

mautamu (approved):
Seems like this is a good way of handling crash-worthy errors; though perhaps in the future a notify-send and attempt to repair might also be in order.

TornaxO7:
@AethanFoot there's still this build error and I don't really know what I should do with it :sweat_smile: do you have any suggestions?

AethanFoot:
Looks like the starting and killing lefthk accidentally got refactored out (make sure to add the lefthk feature checks).

TornaxO7:
Looks like that it's fine now :thinking:

leftwm/src/bin/leftwm.rs
```diff
@@ -151,10 +151,8 @@ fn start_leftwm() {
             }
         }
 
-        #[cfg(feature = "lefthk")]
-        kill_lefthk_session(&mut lefthk_session);
```
VuiMuich, line 155: This is needed I believe, or did I simply miss, where it got moved to?

leftwm/src/bin/leftwm.rs
```diff
@@ -137,12 +137,12 @@ fn start_leftwm() {
     let mut children = Nanny::autostart();
 
     let flag = get_sigchld_flag();
-    loop {
-        let mut leftwm_session = start_leftwm_session(&current_exe);
-        #[cfg(feature = "lefthk")]
-        let mut lefthk_session = start_lefthk_session(&current_exe);
```
VuiMuich, line 143: Also needed. I think this is why CI fails.

VuiMuich (approved):

hertg (approved):
Have no time to test-run right now, but looks fine code wise

VuiMuich:
As there is a number of reports, that users get thrown out of their session, when only supposedly recoverable things crash I suspect this PR might be the most probable culprit.
Maybe we should indeed return to a more robust approach and only log the error state of the child.
@TornaxO7 would you have time to look into this?

TornaxO7:
@VuiMuich may I ask if you can provide me a link to try to reproduce the issues please?

VuiMuich:
Most of them got reported on discord, some even in dm, but I asked the latest one to report as an issue.

PR #835, opened 2022-09-10
Ignorting .toggletasks.json

TornaxO7:
# Description

This PR is only relevant for (neo-)vim users who use the
[`toggletasks.nvim`](https://github.com/jedrzejboczar/toggletasks.nvim) plugin
since it needs a json file where they can write their execution tasks. This
should be probably ignored for the most users, at least that's what I think.
It *could* be probably useful for other neovim users as well though.

## Type of change

Not needed

# Checklist:

- [X] Wiki pages have been updated accordingly (to perform **after** merge)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ result
 .envrc
 
 sweep.timestamp
+
+.toggletasks.json


TornaxO7:
Thank you!

PR #841, opened 2022-09-13
Updated `flake.nix` for lefthk support (also updated `flake.lock`)

Syudagye:
# Description

Updates the `flake.nix` file to include the `libxcb` build dependency needed by lefthk.
Using leftwm built with the recent lefthk implementation with nix flake resulted in no keyboard inputs being processed, because it was missing this dependency.

## Type of change


VuiMuich:
Thanks!

PR #850, opened 2022-09-21
refactor score window function

TornaxO7:
# Description

I just stepped over this while looking at the window creation in leftwm.
The score computation was a bit hard to understand so I splitted it into vars to
make it easier to understand the steps/calculation of the score.

## Type of change

Not needed

# Checklist:


carrascomj (approved):
It looks good to me: I'd bet that the compiler arrives to the same machine code and it is more readable.

AethanFoot (approved):

PR #853, opened 2022-09-22
fix building with no features

AethanFoot:
When building without lefthk we would get an error as `config.keybind` does not exist. For some reason having a `cfg(feature...)` is not allowed above setting an attribute in this situation so I made a function.

### DIFF
--- a/leftwm/src/bin/leftwm-worker.rs
+++ b/leftwm/src/bin/leftwm-worker.rs
@@ -14,10 +14,15 @@ fn main() {
         let rt = tokio::runtime::Runtime::new().expect("ERROR: couldn't init Tokio runtime");
         let _rt_guard = rt.enter();
 
+        #[cfg(feature = "lefthk")]
         let mut config = leftwm::load();
         // Clear the keybinds so leftwm is not storing them.
         // TODO: Make this more elegant.
-        config.keybind = vec![];
+        #[cfg(feature = "lefthk")]
+        config.clear_keybinds();
+
+        #[cfg(not(feature = "lefthk"))]
+        let config = leftwm::load();
 
         let manager = Manager::<leftwm::Config, XlibDisplayServer>::new(config);
         manager.register_child_hook();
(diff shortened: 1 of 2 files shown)


leftwm/src/config/mod.rs
```diff
@@ -42,7 +42,7 @@ const STATE_FILE: &str = "/tmp/leftwm.state";
 ///
 /// ```ron
 /// window_rules: [
-///     (window_class: "krita", spawn_on_tag: 3, spawn_floating: farse),   
+///     (window_class: "krita", spawn_on_tag: 3, spawn_floating: farse),
```
VuiMuich, line 45:
```suggestion
///     (window_class: "krita", spawn_on_tag: 3, spawn_floating: false),
```

leftwm/src/bin/leftwm-worker.rs
```diff
@@ -14,10 +14,15 @@ fn main() {
         let rt = tokio::runtime::Runtime::new().expect("ERROR: couldn't init Tokio runtime");
         let _rt_guard = rt.enter();
 
+        #[cfg(feature = "lefthk")]
         let mut config = leftwm::load();
         // Clear the keybinds so leftwm is not storing them.
         // TODO: Make this more elegant.
```
VuiMuich, line 20:
```suggestion
```
or do we want this even more elegant?

VuiMuich (commented):
Besides those two commentary nitpicks, lgtm.

leftwm/src/bin/leftwm-worker.rs
AethanFoot: even more elegant, preferably when we redo the config setup we will have something like:
```
pub struct Config {
    pub  settings: Settings,
    pub keybinds: Keybinds,
}
```
Then Settings will implemented the leftwm config trait and keybinds will implement lefthk config trait, then settings gets passed to leftwm and the keybinds to lefthk after loading the config

Issue #856, opened 2022-09-24
Incorrect behavior of `ToggleScratchpad` with focus behavior `ClickTo`


SamClercky:
### Describe the bug

This is a follow up of PR #768 where if the manual handling of the visibility in the scratchpad commands is not handled in the commands itself, the behavior is not as expected when focus behavior is set to `ClickTo`.

Observed behavior: (more in information in detail can be found in previously mentioned PR)
* Toggling SP on only works when there is a window that has focus
* Toggling SP off doesn't return focus to any other visible window
* `NextScratchPad` / `PrevScratchPad` first toggle the curren SP window of only with the next execution they open the next SP window

### Expected behavior (optional)

* `ToggleScratchPad` should toggle the scratchpad more than once without the handling of the visibility in the scratchpad commands.
* Toggling the scratchpad should return focus to another visible window
* `NextScratchPad` / `PrevScratchPad` should not toggle the current scratchpad and only cycle when there is a scratchpad visible + cycle immediatly

### Steps to reproduce (optional)

1. `git revert d598cbc`
2. Make sure that `focus_behavior` is set to `ClickTo`
3. Go to a tag with no open windows
4. Toggle twice the scratchpad with `ToggleScratchPad` (show + hide)
5. Try to toggle again -> nothing happens

Minimal config:
```rust
    focus_behavior: ClickTo,
    scratchpad: [
        (name: "ScratchpadTerminal", value: "alacritty", x: 0.055, y: 65, width: 0.89, height: 0.34),
        (name: "BTop", value: "alacritty -e btop", x: 10, y: 65),
        (name: "DynamicScratchpad", value: ""),
    ],
    keybind: [
        (command: ReleaseScratchPad, value: "", modifier: ["modkey", "Shift"], key: "d"),
        (command: AttachScratchPad, value: "DynamicScratchpad", modifier: ["modkey", "Control"], key: "d"),
        (command: NextScratchPadWindow, value: "DynamicScratchpad", modifier: ["modkey", "Control"], key: "comma"),
        (command: PrevScratchPadWindow, value: "DynamicScratchpad", modifier: ["modkey", "Control"], key: "period"),

        (command: ReleaseScratchPad, value: "", modifier: ["modkey", "Shift"], key: "d"),
        (command: AttachScratchPad, value: "DynamicScratchpad", modifier: ["modkey", "Control"], key: "d"),
[...]

PR #861, opened 2022-09-25
FIX: make leftwm-check -m migrate scratchpads correctly

### LINKED ISSUE #860: leftwm-check -m dont migrate scratchpads correctly
### Describe the bug

Migrating from config.toml to config.ron handle scratchpad elements incorrect

Having the following config.toml structure:
```
scratchpad = [
    { name = "Mousepad", value = "mousepad", x = 200, y = 300, height = 600, width = 800 },
]
```
and the running `leftwm-check -m` will generate this block in the config.ron file:
```
    scratchpad: [
        (name: ("Mousepad"), value: "mousepad", x: 200, y: 300, height: 600, width: 800),
    ],
```
Simply removing the tuple definition for name will fix the problem

_No response_

_No response_

### OS / Distro

Manjaro 22.0.0

_No response_

### LeftWM Check

```text
leftwm-check
:: LeftWM version: 0.3.0
:: LeftWM git hash: d4fae27
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
WARN: TOML as config format is about to be deprecated.
      Please consider migrating to RON or contact the theme creator about this topic.
      Note: make sure the `up` script is loading the correct theme file.
ERROR: File not found: theme.ron
```

SamClercky:
# Description

The newtype `ScratchPadName` did not define a way to convert to `String` (only from `String` to `ScratchPadName`). This gave the problem that the migration of `TOML` to `RON` introduced extra `(` and `)` around the scratchpad name and thus generating a bad config.

This PR also adds a test which tests that the (de)serialization can go in 2 ways. To make the test initially fail, a default scratchpad config was added to the `Default` impl of `Config`. The parameters were taken from the LeftWM wiki.

Fixes #860

## Type of change

# Checklist:

### DIFF
--- a/leftwm/src/config/default.rs
+++ b/leftwm/src/config/default.rs
@@ -1,3 +1,5 @@
+use leftwm_core::models::{ScratchPad, Size};
+
 #[cfg(feature = "lefthk")]
 use super::{default_terminal, exit_strategy, BaseCommand, Keybind};
 use super::{Config, Default, FocusBehaviour, LayoutMode, ThemeSetting, LAYOUTS};
@@ -199,6 +201,15 @@ impl Default for Config {
             .map(|s| (*s).to_string())
             .collect();
 
+        let scratchpad = ScratchPad {
+            name: "Alacritty".into(),
+            value: "alacritty".to_string(),
+            x: Some(Size::Pixel(860)),
+            y: Some(Size::Pixel(390)),
+            height: Some(Size::Pixel(300)),
+            width: Some(Size::Pixel(200)),
+        };
+
         Self {
             workspaces: Some(vec![]),
             tags: Some(tags),
@@ -206,7 +217,7 @@ impl Default for Config {
             layout_mode: LayoutMode::Workspace,
             // TODO: add sane default for scratchpad config.
             // Currently default values are set in sane_dimension fn.
-            scratchpad: Some(vec![]),
+            scratchpad: Some(vec![scratchpad]),
             window_rules: Some(vec![]),
             disable_current_tag_swap: false,
             disable_tile_drag: false,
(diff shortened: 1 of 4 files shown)


leftwm/src/config/default.rs
```diff
@@ -199,14 +201,23 @@ impl Default for Config {
...
+            width: Some(Size::Pixel(200)),
+        };
+
         Self {
             workspaces: Some(vec![]),
             tags: Some(tags),
             layouts: LAYOUTS.to_vec(),
             layout_mode: LayoutMode::Workspace,
             // TODO: add sane default for scratchpad config.
             // Currently default values are set in sane_dimension fn.
```
TornaxO7, line 219: Is this TODO still needed?

TornaxO7 (commented):
Added a little question.

leftwm/src/config/default.rs
VuiMuich, line 219: I think what was meant by this TODO was that `fn sane_dimension` is not a good place to set some default values.
Maybe we can clean this up with a fix to #856?

Good catch though.

SamClercky: Good question. I don't really know as this looks to me more like the question of what the defaults should be. I added these values because these came from the wiki and gave a case to test against.

This might create the need to open a new PR which discusses what the defaults should look like?

VuiMuich:
Thanks for the quick fix @SamClercky. 
LGTM

:heart: for the two-in-one test :smiley:

leftwm/src/config/default.rs
TornaxO7: > This might create the need to open a new PR which discusses what the defaults should look like?

I agree

TornaxO7 (approved):

PR #804, opened 2022-07-25
Floating window improvements

jmattheis:
# Description

I wanted to fix #700 by using the window size / location hints when spawn_floating is configured. I don't think I fully fixed this, because we probably want to have a better default than falling back to "fullscreening" the window if it doesn't have a max height/width. Some windows define min width / height, this could be used as default width / height. I'd say this can be done in a separate PR.

As a side effect #632 will also be ~~fixed~~ partly fixed because windows will receive the last location / size as hint. ~~I don't know if this is application specific or something that is built into X, so I don't know if this completely fixes #632. This behavior is similar to the one from i3 for floating windows.~~ This only applies to apps that manually save the location / size and then set them as hint on startup.

<details>
  <summary>xprop output from last position hint</summary>
  
  ```
_NET_WM_DESKTOP(CARDINAL) = 0
_NET_WM_ICON(CARDINAL) =        Icon (128 x 128):
        (not shown)

WM_STATE(WM_STATE):
                window state: Normal
                icon window: <field not available>
_NET_WM_STATE(ATOM) =
XdndAware(ATOM) = BITMAP
WM_NAME(STRING) =
_NET_WM_NAME(UTF8_STRING) = "Yubico Authenticator"
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x3e, 0x7e, 0x0, 0x0
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
_XEMBED_INFO(_XEMBED_INFO) = 0x0, 0x1
WM_CLIENT_LEADER(WINDOW): window id # 0x2800012
WM_HINTS(WM_HINTS):
                Client accepts input or input focus: True
                window id # of group leader: 0x2800012
WM_CLIENT_MACHINE(STRING) = "jannix"
_NET_WM_PID(CARDINAL) = 1146335
_NET_WM_SYNC_REQUEST_COUNTER(CARDINAL) = 41943057
WM_CLASS(STRING) = "yubioath-desktop", "Yubico Authenticator"
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, WM_TAKE_FOCUS, _NET_WM_PING, _NET_WM_SYNC_REQUEST
WM_NORMAL_HINTS(WM_SIZE_HINTS):
                user specified location: 3893, 604
                user specified size: 427 by 657
                program specified minimum size: 270 by 381
                window gravity: Static
  ```
  
</details>

<details>
[...]

PR #863, opened 2022-09-26
making centering working again

### LINKED ISSUE #824: floating windows spawn top-left instead of centered
Any app that requests a floating window or a floating prompt it opens in the (0,0) coordinates.
Peek 2022-09-02 14-34

This is an example in blueberry, also encountered in vs-code.

Before that these windows used to open in the middle of the screen.

**Environment:**
 - OS/Distro: ArcoLinux
 - image

TornaxO7:
# Description

Fixes #824

## Type of change

Not needed

# Checklist:

### DIFF
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -311,16 +311,12 @@ fn set_relative_floating(window: &mut Window, ws: &Workspace, outer: Xyhw) {
     let xyhw = window.requested.map_or_else(
         || ws.center_halfed(),
         |mut requested| {
+            requested.center_relative(outer, window.border);
             if ws.xyhw.contains_xyhw(&requested) {
-                requested.center_relative(ws.xyhw, window.border);
                 requested
             } else {
-                requested.center_relative(outer, window.border);
-                if ws.xyhw.contains_xyhw(&requested) {
-                    requested
-                } else {
-                    ws.center_halfed()
-                }
+                requested.center_relative(ws.xyhw, window.border);
+                requested
             }
         },
     );
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -314,8 +314,8 @@ fn set_relative_floating(window: &mut Window, ws: &Workspace) {
         |mut requested| {
             requested.center_relative(ws.xyhw, window.border);
             requested
-
-    });
+        },
+    );
     window.set_floating_exact(xyhw);
 }
 
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -135,7 +135,7 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
                 _ => None,
             };
 
-            tracing::debug!("WINDOW CHANGED {:?} {:?}", &window, change);
+            // tracing::debug!("WINDOW CHANGED {:?} {:?}", &window, change);
             changed = change.update(window, container);
             if window.r#type == WindowType::Dock {
                 update_workspace_avoid_list(&mut self.state);
@@ -315,18 +315,26 @@ fn set_relative_floating(window: &mut Window, ws: &Workspace, outer: Xyhw) {
     let xyhw = window.requested.map_or_else(
         || ws.center_halfed(),
         |mut requested| {
-            if ws.xyhw.contains_xyhw(&requested) {
-                requested
-            } else {
-                requested.center_relative(outer, window.border);
-                if ws.xyhw.contains_xyhw(&requested) {
-                    requested
-                } else {
-                    ws.center_halfed()
-                }
-            }
+            requested.center_relative(outer, window.border);
+            requested
+            // tracing::debug!("Requested: {:?}", requested);
+            // if ws.xyhw.contains_xyhw(&requested) {
+            //     tracing::debug!("YES Requested");
+            //     requested
+            // } else {
+            //     tracing::debug!("ADJUSTED requested: {:?}", requested);
+            //     if ws.xyhw.contains_xyhw(&requested) {
+            //         tracing::debug!("Requested finally in ws!");
+            //         requested
+            //     } else {
+            //         tracing::debug!("Requested still not in it");
+            //         ws.center_halfed()
+            //     }
+            // }
         },
     );
+    tracing::debug!("window.normal: {:?}", window.normal);
+    tracing::debug!("xyhw: {:?}", xyhw);
     window.set_floating_exact(xyhw);
 }
 
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -309,16 +309,16 @@ fn is_scratchpad(state: &State, window: &Window) -> bool {
         .any(|(_, id)| id.iter().any(|id| window.pid == Some(*id)))
 }
 
-fn set_relative_floating(window: &mut Window, ws: &Workspace, outer: Xyhw) {
+fn set_relative_floating(window: &mut Window, ws: &Workspace) {
     window.set_floating(true);
     window.normal = ws.xyhw;
     let xyhw = window.requested.map_or_else(
         || ws.center_halfed(),
         |mut requested| {
-            requested.center_relative(outer, window.border);
+            requested.center_relative(ws.xyhw, window.border);
             requested
-        },
-    );
+
+    });
     window.set_floating_exact(xyhw);
 }
 
@@ -375,7 +375,7 @@ fn setup_window(
             // This is currently for vlc, this probably will need to be more general if another
             // case comes up where we don't want to move the window.
             if window.r#type != WindowType::Utility {
-                set_relative_floating(window, ws, parent.exact_xyhw());
+                set_relative_floating(window, ws);
                 return;
             }
         }
@@ -385,7 +385,7 @@ fn setup_window(
             WindowType::Normal => {
                 window.apply_margin_multiplier(ws.margin_multiplier);
                 if window.floating() {
-                    set_relative_floating(window, ws, ws.xyhw);
+                    set_relative_floating(window, ws);
                 }
             }
             WindowType::Dialog => {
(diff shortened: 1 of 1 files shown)


leftwm-core/src/handlers/window_handler.rs
```diff
@@ -135,7 +132,7 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
                 _ => None,
             };
 
-            tracing::debug!("WINDOW CHANGED {:?} {:?}", &window, change);
+            // tracing::debug!("WINDOW CHANGED {:?} {:?}", &window, change);
```
TornaxO7: My opinion: This can be put aside for the time being since it's called *very* often I'd say, that we should rather create this for the necessary parts/functions otherwise the log file is gonna fill up very fast.

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -378,22 +367,12 @@ fn setup_window(
...
-        // Setup a child window.
-        if let Some(parent) = find_transient_parent(&state.windows, window.transient) {
-            // This is currently for vlc, this probably will need to be more general if another
-            // case comes up where we don't want to move the window.
-            if window.r#type != WindowType::Utility {
-                set_relative_floating(window, ws, parent.exact_xyhw());
-                return;
-            }
-        }
-
```
TornaxO7, line 390: I tried to find the use-case with `vlc` for that but even after removing it all dialogues were fine with `vlc`. Could anyone elaborate which case this if-clause catches?

TornaxO7 (commented):
Some general comments for discussion (or just my thoughts).

TornaxO7:
@M-Mabrouk1 can you test this one please?

TornaxO7:
@hertg @VuiMuich @AethanFoot (little ping)

leftwm-core/src/handlers/window_handler.rs
AethanFoot: The play pause popup that appears when vlc is in fullscreen. We obviously dont want to centre it

M-Mabrouk1:
> @M-Mabrouk1 can you test this one please?

It works the windows are centered back again. Although they are a little bigger than their normal size if I remember correctly not sure if this is handled by the app or the window manager.
image


Also, I didn't see the issue with the vlc, it used to happen every now and then before and I used to delete the config file for vlc.

Thank you for your effort.

leftwm-core/src/handlers/window_handler.rs
VuiMuich: Tested vlc specifically and it seems to work correct at first, but when you click any empty space of the play-controls window during fullscreen, it jumps to a position more or less centered.

VuiMuich:
One little nicety we lost compared to #804 is, that we don't use the `user specified position` any more. Would be nice if we could retain this.

leftwm-core/src/handlers/window_handler.rs
AethanFoot: Also to add to this, the parent clause is for any application and centres the dialog on the window

TornaxO7:
> One little nicety we lost compared to https://github.com/leftwm/leftwm/pull/804 is, that we don't use the user specified position any more. Would be nice if we could retain this.

Hm... may I ask what you mean with `user specified position`?

leftwm-core/src/handlers/window_handler.rs
AethanFoot: Yeah, in the case where there is a parent we want to centre on the parent. However if when centring, we go outside the workspace (like the edge of the window being off the screen, what `xyhw.contains_xyhw()` checks), we would want to centre on the workspace. Here is an example with android studio where the dialog is centred on the app, which is nicer:
example

TornaxO7: Hm.... ok I understand what you mean now. I'll fix that

AethanFoot:
It is part of the WM_SIZE_HINTS, but i think it is the reason the window werent centring as the program/user hints would place things like splashes and such in the top left very often

TornaxO7:
I'm unsure now what kind of feature is missing then, could you please give me a little example?

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -315,16 +311,8 @@ fn set_relative_floating(window: &mut Window, ws: &Workspace, outer: Xyhw) {
...
-                requested
-            } else {
-                requested.center_relative(outer, window.border);
-                if ws.xyhw.contains_xyhw(&requested) {
-                    requested
-                } else {
-                    ws.center_halfed()
-                }
-            }
+            requested.center_relative(outer, window.border);
```
AethanFoot, line 314: you might want to check that this is contained within the workspace so the window doesnt get cut off, e.g. if the parent is very narrow and at the edge of the screen the dialog is likely to spawn over the edge. Then if it is cut off then we should centre on the workspace so the whole window is visible. (if that makes sense)

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -315,16 +311,8 @@ fn set_relative_floating(window: &mut Window, ws: &Workspace, outer: Xyhw) {
...
-            if ws.xyhw.contains_xyhw(&requested) {
-                requested
-            } else {
-                requested.center_relative(outer, window.border);
-                if ws.xyhw.contains_xyhw(&requested) {
-                    requested
-                } else {
-                    ws.center_halfed()
-                }
-            }
```
AethanFoot, line 327: Basically this part, except you can do `requested.center_relative(ws.xyhw, window.border)` instead of `ws.center_halfed()`

AethanFoot:
I am not sure really, the only cases I can think of with using the program/user defined location is like with the vlc popup which is of type utility

TornaxO7:
@AethanFoot @VuiMuich alright, I added some of the old code again, this should work now.

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -316,6 +312,7 @@ fn set_relative_floating(window: &mut Window, ws: &Workspace, outer: Xyhw) {
         || ws.center_halfed(),
         |mut requested| {
```
AethanFoot, line 313: Sorry I meant something different in this section. E.g.:
```
requested.center_relative(outer, window.border);
if ws.xyhw.contains_xyhw(&requested) {
    requested
} else {
    requested.center_relative(ws.xyhw, window.border);
}
```

AethanFoot: center_relative alters the position of requested, so requested would actually need to be the final value in the second clause

AethanFoot: My code would centre the window on the parent, check whether this new position gets cut off, then if it doesnt returns the position when centred on the parent, if it is it returns the position centred on the workspace

TornaxO7: I updated the code, is that what you're suggesting?

AethanFoot: Yeah, does i work as expected? On mobile so cant verify things, but it should work.

TornaxO7: Well I don't have any issues, so I guess that it works for me. @VuiMuich @M-Mabrouk1 could you please try my latest commit?

VuiMuich: I am at work right now, but can check later when home.

leftwm-core/src/handlers/window_handler.rs
TornaxO7: I'll close this due to the latest commit.

leftwm-core/src/handlers/window_handler.rs
TornaxO7: I'll close this due to the latest commit.

TornaxO7:
Oh bruh, sorry for that many "requested review"'s.... I was wondering why the icon didn't change when I pressed on the "request review" button, so I pressed it a little bit more often...

VuiMuich:
No worries, I am an expert in ignoring pings :stuck_out_tongue_closed_eyes:

For most parts it works fine for me, but vlc still places the fullscreen controls centered.
I guess we should allow windows of type `utility` to strictly use the requested size and position.
This would be even more on `EWMH` specs imo:
> _NET_WM_WINDOW_TYPE_UTILITY indicates a small persistent utility window, such as a palette or toolbox. It is distinct from type TOOLBAR because it does not correspond to a toolbar torn off from the main application. It's distinct from type DIALOG because it isn't a transient dialog, the user will probably keep it open while they're working. Windows of this type may set the WM_TRANSIENT_FOR hint indicating the main application window.

So if we get this we are ready for `0.4.0` I think. The rest, like `user defined position` can be done in a later PR, possibly by adding a window rule as the use cases where this might apply, also aligns probably most of the with setting a `spawn_floating` rule anyways.

TornaxO7:
> For most parts it works fine for me, but vlc still places the fullscreen controls centered.

May I ask you what those `fullscreen controls` are?

VuiMuich:
2022-09-29-154747_2560x1440_scrot

TornaxO7:
I'm not able to reproduce this, could you please provide some reproduction steps?

VuiMuich:
I found where vlc stores the user defined position for the `fullscreen controller` (`~/.config/vlc/vlc-qt-interface.conf`) and reset it by removing the associated line. Works now as intended. I experience a little inconsistency of this utility window showing up, but I think this is not related to this PR, probably not even WM related at all.

BTW: those fullscreen controls show whenever you move the mouse while using vlc in fullscreen (hit `f` or `double-click` on the video area to enter fullscreen).

So I guess I fooled my self and `utility` windows are actually acting as intended by being `unmanaged` as they should. Sorry for sending you on the hunt for a ghost there.

With that sorted, I'd say this is ready. The only question now is, if we want to hold this back so it gets credited for Hacktober (does this even work this way, or do the commits need to have October time stamps)?

VuiMuich (approved):

AethanFoot (approved):

hertg (approved):

hertg:
I suspect I figured out what this line was for.

https://github.com/leftwm/leftwm/blob/78faaeae102403f70d6f89221c7199131e304e31/leftwm-core/src/handlers/window_handler.rs#L54

I think since this PR, some windows that were always on top of others aren't anymore. For example, rofi gets overlapped by tiled windows now, if the windows spawned **after** rofi.

VuiMuich:
> I think since this PR, some windows that were always on top of others aren't anymore. For example, rofi gets overlapped by tiled windows now, if the windows spawned **after** rofi.

Can confirm also putting this line back spawns new `normal` window in the expected stacking position: behind floating windows like rofi or a scratchpad.

PR #870, opened 2022-10-02
Update README.md

Sokom141:
# Description

Clarification on the use of systemd logging as the default for cargo builds under `Rebuilding the development installation`.


VuiMuich:
thanks.

PR #887, opened 2022-10-11
Fix regression that `wmctrl` and `rofi` could not request focus for a window

VuiMuich:
# Description

It seems like [commit 54ef2d91](https://github.com/leftwm/leftwm/pull/761/commits/54ef2d9142033ff1cef0667d156de84fcb52d010) broke the ability to act on focus requests by clients like `rofi -show windowcd` or `wmctrl -i -a <window id>`.

I am not sure if the line introduced `set_window_urgency` is now kind of obsolete now.
Regarding this line, wouldn't it actually make sense to send it a `false` as we want to reset a `_NET_WM_STATE_DEMANDS_ATTENTION` here, or do I get it wrong what this does?

Edit:
After re-reading the EWMH specs of `_NET_ACTIVE_WINDOW` I think there should be a config switch to decide if this moves the focus, or if it sets the `urgency`. I am not sure how to dub this though. Maybe like this:
```rust
pub enum active_window_request {
    focus,
    urgency,
}
```

Also would we want to refactor all these `if event.message_type` into a `match`?

## Type of change

# Checklist:


hertg:
may be a stupid question, but how do we display a window's urgency to the user? Do we draw attention to it in some way? If we don't, what other purpose does the urgency flag serve?

VuiMuich:
tags have a `urgent` flag and I suppose this becomes true if any window with that tag has the `_NET_WM_STATE_DEMANDS_ATTENTION` atom. But tbh I haven't checked the code for this.

But given you comment: a `urgent_border_color` might be a useful addition in that regard.

VuiMuich:
I am not certain, but it might be, this currently breaks the `urgent` flag in state output.

AethanFoot:
I found when setting focus with this, many windows would spam it and not allow u to leave a window in some cases. This is an issue with sloppy as the mouse takes the focus, but in past when moving the mouse as well with the focus caused this issue so I suspect this would be an issue with driven in the current situation.

VuiMuich:
:see_no_evil: sorry for the commit mess... brains not awake yet, also I guess amending pushed commits is not such a bright idea.

Anyways..
Do you recall what windows were spamming this message? I really would like to bring back the ability for `wmctrl` and `rofi` (and others) to focus a window.
Maybe we should evaluate the `source indication` to determine what we want to do?

VuiMuich:
Ok, I hope this is better now.
Just one thing I noticed: in `Monocle` and `MainAndDeck` the window does not get pulled to the top, when it is in the deck. But I haven't had a look into this.

AethanFoot:
Will need to figure out which application it is again. Will do some testing when I have time

AethanFoot:
Do we want to use `HandleWindowFocus` instead, as currently this would not do anything in sloppy mode.

VuiMuich:
Yeah, just after pushing the last commit it dawned on me, that it actually might be better to move all the logic to `leftwm-core` and let the event-translated just forward it to the right method, as this is what it actually says it does: translate...

AethanFoot (approved):
Just wanted to approve as this would be okay (I would change `WindowTakeFocus` to `HandleWindowFocus` so that the mouse is moved with the focus)

VuiMuich:
> Just wanted to approve as this would be okay (I would change `WindowTakeFocus` to `HandleWindowFocus` so that the mouse is moved with the focus)

Were you referring to the commit or to my last message?
Was also thinking of refactoring the `translation` part to a `match` as this might be more readable, what do you think?

charbelnicolas:
I'm waiting on this to give leftwm another try... Not that there is anything wrong with awesome wm :)

PR #891, opened 2022-10-11
feat: Add a `spawn_as_type` to the `window_rules` configuration

### LINKED ISSUE #766: Firefox Picture-In-Picture Issues
I use the [Picture-In-Picture](https://support.mozilla.org/en-US/kb/about-picture-picture-firefox) (will shorten it to PIP) feature of Firefox quite a lot, but it has some very weird behaviors. I'm not sure whether LeftWM or Firefox is to blame for that, but I will list them here. Sorted from most annoying to least annoying.

1. The PIP window always  jumps to first position when a new window opens.
   1.1 ~See the gif below, this is the `CenterMain` layout. When I open a new terminal, the PIP window jumps to the main position.~ -> This was fixed by #863 i think
     output
   
   1.2 *Btw. It also happens when you move a floating window, as soon as you let go.*
   output

2. The PIP window is always in front of everything, even when it is tiled.
    output

3. Window is larger than it should be, it seems to not respect the paddings.
    out

5. The PIP window shows no borders. It's hard to figure out whether it is focused.

**Additional context**
Slightly related to #220

Sokom141:
# Description

This PR adds a field to the window configuration in `window_rules`.
This field is called `spawn_as_type` and is a way for the user to set the [window type](https://specifications.freedesktop.org/wm-spec/latest/ar01s05.html#idm45374033166992) of a specific window with the usual rules via a config change.

Fixes #766 
Follow up to #879
Other references #631 

## Type of change

### Firefox Picture-in-Picture weird behavior fix:
Add this to your `config.ron` file in the `window_rules` section:
```ron
#![enable(implicit_some)]
(
    ...
    window_rules: [
        ...
        ( window_title: "Picture-in-Picture", spawn_as_type: Normal ),
        ...
    ]
    ...
)
```

# Checklist:

### DIFF
--- a/leftwm/src/config/mod.rs
+++ b/leftwm/src/config/mod.rs
@@ -15,7 +15,7 @@ use anyhow::Result;
 use leftwm_core::{
     config::{InsertBehavior, ScratchPad, Workspace},
     layouts::{Layout, LAYOUTS},
-    models::{FocusBehaviour, Gutter, LayoutMode, Margins, Size, Window},
+    models::{FocusBehaviour, Gutter, LayoutMode, Margins, Size, Window, WindowType},
     state::State,
     DisplayServer, Manager,
 };
@@ -64,6 +64,8 @@ pub struct WindowHook {
     pub window_title: Option<String>,
     pub spawn_on_tag: Option<usize>,
     pub spawn_floating: Option<bool>,
+    /// Handle the window as if it was of this `_NET_WM_WINDOW_TYPE`
+    pub spawn_as_type: Option<WindowType>,
 }
 
 impl WindowHook {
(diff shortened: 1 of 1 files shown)


VuiMuich:
lgtm
added @hertg as reviewer as he seems to be the one most interested in this :grin:

hertg (approved):
I haven't tested it due to lack of time currently, but the code looks good.

What a small and concise fix, and a neat feature addition actually.

Thank you for your contribution ❤️

hertg:
Just tested it locally and it works as expected.
We should also put a comment in #766 that the issue can be manually resolved by adding the window rule.

```ron
( window_title: "Picture-in-Picture", spawn_as_type: Normal ),
```

Also, do we have a wiki page for "Tips and Tricks" kind of stuff where we could point that out?

VuiMuich:
> Also, do we have a wiki page for "Tips and Tricks" kind of stuff where we could point that out?

That sounds like a good idea, also some FAQ might make sense, or would this be a better fit for a github.io website overhaul?

Eskaan (approved):
Exactly the changes as I made (and then reverted) in #900, so LGTM!

PR #892, opened 2022-10-12
No single border

thled:
# Description

This PR adds a configuration option `single_window_border`.
Its default is `true` but if set to `false` it removes the border of a window if it is the only one on its tag.

Motivation: I mostly only have one window open on a given tag. A border there just uses unnecessary space because it's obvious where the window ends.

## Type of change

On https://github.com/leftwm/leftwm/wiki/Config:

**Single Window Border**

This shows a border on windows even if they are the only normal window on their tags.

Default: `single_window_border: true`

# Checklist:

### DIFF
--- a/leftwm-core/src/config/mod.rs
+++ b/leftwm-core/src/config/mod.rs
@@ -27,7 +27,7 @@ pub trait Config {
 
     fn insert_behavior(&self) -> InsertBehavior;
 
-    fn no_single_border(&self) -> bool;
+    fn single_window_border(&self) -> bool;
 
     fn focus_new_windows(&self) -> bool;
 
@@ -93,7 +93,7 @@ pub(crate) mod tests {
         pub workspaces: Option<Vec<Workspace>>,
         pub insert_behavior: InsertBehavior,
         pub border_width: i32,
-        pub no_single_border: bool,
+        pub single_window_border: bool,
     }
 
     impl Config for TestConfig {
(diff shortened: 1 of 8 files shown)


VuiMuich:
Welcome adventurer!

Thanks for opening this PR! I had a brief look at it rn on my phone and it looks very good.
I have two little requests though:
- I think renaming the config option to `single_window_border` might describe better what it does. This would obviously require the bool to be inverted compared to your current implementation
- when you are already touching the use imports in `state.rs` would you mind to condense all the `crate::models::...` ones

Will have a deeper look at this later when I'm on my machine, not only on mobile.

thled:
Thanks for the wam welcoming.

I'm glad this configuration addition is considered.
Thank you for your fast feedback. I commited changes in regard of this feedback.

Looking forward to your deep look and the resulting feedback!

VuiMuich:
Been running it the last half hour, looking good so far.
One thing I noticed though is that scratchpads need a bit of extra attention: 
- when a scratchpad is visible during window creation, the border does not get toggled off when hiding the scratchpad
- when toggling a scratchpad on, the border remains off, which might be considered usefull but technically correct and cleaner to implement would probably to count scratchpads same as regular windows.

VuiMuich:
Yet another thing:
When a theme is applied all windows get borders turned on right now.

VuiMuich:
I guess most of this little things above could solved by essentially pairing each call of `handle_single_window` with a call of `sort_windows`.

Still no better idea for the method name and where to put it though. But as we are discussing a bunch of refactors anyway I would be ok leaving this for later.

VuiMuich:
After using this branch for a couple of days, I have to say I like it more then I initially expected.
But I think toggling scratchpads should actually toggle the borders on and off.

VuiMuich:
Thanks!
This is some nice addition.

thled:
Thank you @VuiMuich for the thorough testing and feedback!
I'm happy that it is part of the app now.

PR #896, opened 2022-10-13
No need to run clippy in release mode.

Eskaan:
# Description

There is no need to run clippy in release mode. Clippy is only a linter and this just slows down make test as the previous compiles all compile in development mode (all code has to be compiled again just for clippy, can't use cache).

Also, there is no mention of the --release flag in the [clippy readme](https://github.com/rust-lang/rust-clippy#step-3-run-clippy).

## Type of change

- [x] Development change (no change visible for user)

# Checklist:


VuiMuich:
Thanks

PR #905, opened 2022-10-15
Fix `--no-default-features` build failing

VuiMuich:
# Description

As brought up be Adàhy on discord dm building with `cargo b --no-default-features` failed.
With this PR it is now possible to build actually without a hotkey daemon.
If somobody would have a suggestion for a nicer way to work around blocking the `main` function in `lefthk_worker.rs` this would be very welcome.

## Type of change

# Checklist:

### DIFF
--- a/leftwm/src/bin/lefthk-worker.rs
+++ b/leftwm/src/bin/lefthk-worker.rs
@@ -3,8 +3,14 @@ use lefthk_core::{config::Config, worker::Worker};
 #[cfg(feature = "lefthk")]
 use xdg::BaseDirectories;
 
-#[cfg(feature = "lefthk")]
 fn main() {
+    // we need this little shenanigan to allow actually building with `--no-default-features`
+    #[cfg(feature = "lefthk")]
+    lefthk_worker_main()
+}
+
+#[cfg(feature = "lefthk")]
+fn lefthk_worker_main() {
     leftwm::utils::log::setup_logging();
 
     tracing::info!("lefthk-worker booted!");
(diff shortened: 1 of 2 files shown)


leftwm/src/bin/lefthk-worker.rs
```diff
@@ -3,8 +3,14 @@ use lefthk_core::{config::Config, worker::Worker};
...
 
-#[cfg(feature = "lefthk")]
 fn main() {
+    // we need this little shenanigan to allow actually building with `--no-default-features`
+    #[cfg(feature = "lefthk")]
+    lefthk_worker_main();
+}
+
+#[cfg(feature = "lefthk")]
+fn lefthk_worker_main() {
```
TornaxO7, line 13: Isn't that the same as just removing `#[cfg(feature = "lefthk")]` above the `fn main()`?

leftwm/src/config/keybind.rs
```diff
@@ -173,6 +173,7 @@ impl Modifier {
     }
 }
 
+#[cfg(feature = "lefthk")]
```
TornaxO7, line 176: may I ask, why we need it here?

TornaxO7 (changes requested):
Added some questions.

leftwm/src/config/keybind.rs
VuiMuich: Because this function is only used in keybinds, and without the feature flag `&Config` is not in scope, so compilation breaks.

leftwm/src/bin/lefthk-worker.rs
VuiMuich: Thanks this was exactly the hint I needed to finde a better solution.

PR #915, opened 2022-10-18
Feature: Multiple `up` scripts

### LINKED ISSUE #547: Search leftwm config directory for .up scripts instead of up script
To use *name*.up so users can use other names and split the scripts up for better clarity. Maybe related to #292.

Sokom141:
# Description

This PR gives the user the ability to break up the global script into an arbitrary number of scripts. Scripts must have a `.up` extension and will be executed in alphabetical order (based on the implementation of `PathBuf`).
It is also still possible for the user to have an `up` script that will be executed as the last script.

A `BinaryHeap` was chosen instead of a `Vec` as the emitted assembly [code was much shorter](https://godbolt.org/z/jo9q34xa6).

Fixes #547 

## Type of change

In [wiki](https://github.com/leftwm/leftwm/wiki/Config#autostart):

the same scripts can be placed in ~/.config/leftwm/ in order to be theme agnostic Note: there is no override rules between "theme-scripts" and "config-scripts", as they are agnostic of each other.

should be replaced with:

the same scripts can be placed in ~/.config/leftwm/ and broken up in multiple _name_.up scripts in order to be theme agnostic.
_i.e.:_ you could have four files:  `b.up`, `c.up`, `a.up` and `up`. The execution order will be: `a.up`, `b.up`, `c.up`, `up`.

**Note**: there is no override rules between "theme-scripts" and "config-scripts", as they are agnostic of each other.


leftwm-core/src/utils/child_process.rs
```diff
@@ -87,6 +88,31 @@ impl Nanny {
...
+                Err(_) => continue,
+            };
+            let file = entry.path();
+
+            if let Some(extension) = file.extension() {
+                if extension == "up" {
+                    scripts.push(Reverse(file));
+                }
+            }
+        }
```
TornaxO7, line 107: this code-part has its own logic, so I'd move it out to a function.

Also does your scripts-collection needs to be a Binary-Heap? If I'm seeing it right, you're just collecting the values, this can be done with a `Vec` which has (amortized) O(1) whichis definetely better than O(log(n)).

leftwm-core/src/utils/child_process.rs
```diff
@@ -87,6 +88,31 @@ impl Nanny {
...
         let mut path = Self::get_config_dir()?;
+        let config_dir = fs::read_dir(&path)?;
+
+        let mut scripts = std::collections::BinaryHeap::new();
+
+        for entry in config_dir {
+            let entry = match entry {
+                Ok(e) => e,
+                Err(_) => continue,
+            };
```
TornaxO7, line 99: I'd just use `if let Some(entry) = entry { ... }` also check if entry is a file before collecting it!

TornaxO7 (changes requested):

leftwm-core/src/utils/child_process.rs
Sokom141: > this code-part has its own logic, so I'd move it out to a function.
I agree.

As for the `BinaryHeap` vs `Vec` I choose the former because we want (as discussed on discord) to have an order in the executing of the scripts so if using a `Vec` we should `.sort()` it before executing the scripts.
To be fair most users won't have more than an handful of scripts so it doesn't really matter. I've chosen the BinaryHeap because of the emitted asm code size.
Here is the link to the example: https://godbolt.org/z/sMqocbEhM

What do you think?

leftwm-core/src/utils/child_process.rs
Sokom141: > I'd just use `if let Some(entry) = entry { ... }`

Right!

`Self::run_script` already check if the argument is a file before executing. Should we check two times?

leftwm-core/src/utils/child_process.rs
TornaxO7: > As for the BinaryHeap vs Vec I choose the former because we want (as discussed on discord) to have an order in the executing of the scripts so if using a Vec we should .sort() it before executing the scripts.
To be fair most users won't have more than an handful of scripts so it doesn't really matter. I've chosen the BinaryHeap because of the emitted asm code size.
Here is the link to the example: https://godbolt.org/z/sMqocbEhM
> 
> What do you think? 

Oh ok, I didn't knew that you discussed that on discord. Your argumentation sounds good to me.

leftwm-core/src/utils/child_process.rs
```diff
@@ -87,6 +88,31 @@ impl Nanny {
     /// Could be caused by inadequate permissions.
     pub fn run_global_up_script() -> Result<Option<Child>> {
```
TornaxO7, line 89: Is it by the way planned to have the multiple-up-script-feature only in the main directory?

leftwm-core/src/utils/child_process.rs
TornaxO7: > Self::run_script already check if the argument is a file before executing. Should we check two times?
 
Ok eh... this is odd... normally I'd expect the function to run immediately and return an error if it can't run (for example if it got a path to a directory instead of a file). But I think this can be fixed in another PR.

TornaxO7 (changes requested):

leftwm-core/src/utils/child_process.rs
Sokom141: Yes, as described in #547

leftwm-core/src/utils/child_process.rs
Sokom141: > Ok eh... this is odd... normally I'd expect the function to run immediately and return an error if it can't run (for example if it got a path to a directory instead of a file). But I think this can be fixed in another PR.

Yeah, I agree. Maybe I could code this to just return `up` scripts in an arbitary directory.

leftwm-core/src/utils/child_process.rs
TornaxO7: hm... ok

leftwm-core/src/utils/child_process.rs
TornaxO7: > Yeah, I agree. Maybe I could code this to just return up scripts in an arbitary directory.

I'd prefer this IMO, but maybe someone else has  a different opinion on this @VuiMuich what do you think?

Sokom141: > I'd prefer this IMO, but maybe someone else has a different opinion on this @VuiMuich what do you think?

Right now I've coded this solution to be generic over "files" and "extensions":

```rust
    fn get_files_in_path_with_ext(
        path: impl AsRef<Path>,
        ext: impl AsRef<OsStr>,
    ) -> Result<BinaryHeap<Reverse<PathBuf>>> {
        let config_dir = fs::read_dir(&path)?;

        let mut scripts = BinaryHeap::new();

        for entry in config_dir.flatten() {
            let file = entry.path();

            if let Some(extension) = file.extension() {
                if extension == ext.as_ref() {
                    scripts.push(Reverse(file));
                }
            }
        }

        Ok(scripts)
    }
```

TornaxO7: >  /// This function will return an error if `stf::fs::read_dir()` returns an error,
    /// for example when:
    /// * The provided `path` doesn't exist.
    /// * The process lacks permissions to view the contents.
    /// * The `path` points at a non-directory file.

I'd sum this up to:
```rust
    /// # Errors
    ///
    /// Comes from `std::fs::read_dir()`.
```
because the rest can be read there.

TornaxO7 (approved):

VuiMuich:
Thanks, great work.\nMay I ask you to up date the documentation for autostarting?

PR #922, opened 2022-10-24
Add `background_color` to theme config

Eskaan:
# Description

- Added `background_color` field to the theme config on request of @VuiMuich.
- Changed some theme handling, specifically printing the all error messages when failing to run a script.

## Type of change

Add `background_color: "#333333"` to the example in the to rename Border Colors section.

# Checklist

# Todo
- Make (all) colors optional?
- Which default color?

### DIFF
--- a/leftwm/src/config/mod.rs
+++ b/leftwm/src/config/mod.rs
@@ -540,7 +540,7 @@ impl leftwm_core::Config for Config {
         self.theme_setting
             .focused_border_color
             .clone()
-            .unwrap_or_else(|| "#FFD625".to_string())
+            .unwrap_or_else(|| "#FF0000".to_string())
     }
 
     fn on_new_window_cmd(&self) -> Option<String> {
--- a/leftwm/src/config/mod.rs
+++ b/leftwm/src/config/mod.rs
@@ -507,6 +507,10 @@ impl leftwm_core::Config for Config {
         self.theme_setting.floating_border_color.clone()
     }
 
+    fn background_color(&self) -> String {
+        self.theme_setting.background_color.clone()
+    }
+
     fn disable_window_snap(&self) -> bool {
         self.disable_window_snap
     }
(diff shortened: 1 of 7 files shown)


VuiMuich:
On those two TODOs:
- I welcome this suggestion, and think we might want to go all the way and make all theme fields optional then
- Should we go for a `#FFD625` like the `leftwm logo` for the default focused border color? `#333333` seems reasonable as the default background

Eskaan:
Just a fat fail, will do that tomorrow

VuiMuich:
Now that all colors are `Options` could you please make `border` and `margins` options as well? I think it would be just more consistent that way.

Eskaan:
Just to clarify, not all config options are `Option`s?
I changed all colors, margins and border width are other things.

VuiMuich:
Yes, only _all theme related_ config fields. It just feels silly, when all but two of the fields in `theme.ron` are optional, but those two are not special in any way so there would be a reason for them to be mandatory.

leftwm/src/config/mod.rs
```diff
@@ -469,11 +469,17 @@ impl leftwm_core::Config for Config {
...
-        self.theme_setting.border_width
+        self.theme_setting.border_width.unwrap_or(1)
     }
 
     fn margin(&self) -> Margins {
-        match self.theme_setting.margin.clone().try_into() {
+        match self
+            .theme_setting
+            .margin
+            .clone()
```
VuiMuich, line 479: Do we stell bed this clone? Might be with the `unwarp_or` we could loose this.

Eskaan: If I remove the .clone() it does not compile. `unwrap_or(...)` moves the object.

VuiMuich: Yeah, was just wondering why `border_width` has no clone.

Btw, can't we use `unwrap_or_default`?

Eskaan: It's an enum, so I would have to write a manual implementation. Also all other values also use `.unwrap_or()`, example in line 531 with false.

Eskaan: I think this is to clarify the defaults

PR #932, opened 2022-10-26
Some fixes for the `Makefile` and `flake.nix`

Syudagye:
# Description

This change just fixes little things in the `Makefile`:
- Made the `install-linked-dev` rule build the project in debug mode, since it's linking files in `./target/debug`
- Changed the install dir for the `install` rule from `/usr/bin` to `TARGET_DIR`, which is `/usr/local/bin`, so it gets cleaned up with the `uninstall` rule
- `clean` now removes the `./result` symlink, which appears after building with `nix build`
- removed the invocation of `sudo` in the `test-nix`. If you can't use it without `sudo`, you should check if you are in the `nix-users` group
- added the `--verbose` flag for `test-nix`, so we get more info when building

This also fixes some deprecated declarations on `flake.nix`, which were triggering warnings with `test-nix`.
There is one that remains tho, its the [overlay](https://github.com/Syudagye/leftwm/blob/2d0433ae71414834c14d8d5910c2417230677ca5/flake.nix#L73), I don't know of to fix this for now, but this shouldn't be complicated.

I think this can be "breaking" because of the install directory change i've made.
Since it will install to another location, maybe the leftover binaries from the old location will remain and could be executed instead of the new ones.

## Type of change

# Checklist:

  Note: `test-full-nix` still has a warning for now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,8 @@ test:
 	cd $(ROOT_DIR) && cargo clippy
 
 test-nix:
-	cd $(ROOT_DIR) && sudo NIX_PATH=nixpkgs=channel:nixos-unstable nix flake check --extra-experimental-features "nix-command flakes"
-	cd $(ROOT_DIR) && sudo NIX_PATH=nixpkgs=channel:nixos-unstable nix build --extra-experimental-features "nix-command flakes"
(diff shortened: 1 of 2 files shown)


Eskaan (approved):
Code changes look good to me (didn't test them).
He actually fixed mistakes of mine I didn't know of yet.

PR #935, opened 2022-10-27
Fix binary arguments (clap)

TornaxO7:
# Description

This should make `leftwm-check` working again.

Fixes #(993) (only half)

## Type of change

Not needed

# Checklist:

### DIFF
--- a/leftwm/src/bin/leftwm-check.rs
+++ b/leftwm/src/bin/leftwm-check.rs
@@ -41,11 +41,10 @@ async fn main() -> Result<()> {
 
     let config_file: Option<&str> = matches
         .get_one::<String>(INPUT_ARG)
-        .map(|input| input.as_ref());
+        .map(std::convert::AsRef::as_ref);
     let verbose = matches
         .get_one::<bool>(VERBOSE_ARG)
-        .map(|verbose| verbose.to_owned())
-        .unwrap_or(false);
+        .map_or(false, std::borrow::ToOwned::to_owned);
 
     println!(
         "\x1b[0;94m::\x1b[0m LeftWM version: {}",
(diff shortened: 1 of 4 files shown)


leftwm/src/bin/leftwm-command.rs
```diff
@@ -12,15 +12,10 @@ async fn main() -> Result<()> {
...
         .arg(
-            arg!("command").help("The command to be sent. See 'list' flag."), // .required(true)
-                                                                              // .multiple(true)
-        )
-        .arg(
-            arg!("list")
-                .help("Print a list of available commands with their arguments.")
-                .short('l')
-                .long("list"),
+            arg!(--"command" [COMMAND] "The command to be sent. See 'list' flag."), // .required(true)
```
Eskaan, line 15: This is wrong, we don't want to `leftwm-command --command "Command"` but `leftwm-command "Command"`. 
```suggestion
            arg!(name:command [COMMAND] "The command to be sent. See 'list' flag."),
```

leftwm/src/bin/leftwm-check.rs
```diff
@@ -10,37 +10,41 @@ use std::path::Path;
...
+                .help("Outputs received configuration file.")
         )
         .arg(
-            arg!("migrate")
+            Arg::new(MIGRATE_ARG)
                 .short('m')
                 .long("migrate-toml-to-ron")
-                .help("Migrates an exesting `toml` based config to a `ron` based one.\nKeeps the old file for reference, please delete it manually."),
+                .help("Migrates an exesting `toml` based config to a `ron` based one.\nKeeps the old file for reference, please delete it manually.")
         )
```
Eskaan, line 39: You can use `.args(&[<args>])` to make this more readable. Also in the other files.

Eskaan (changes requested):
Some things I wanted to mention to explain

leftwm/src/bin/leftwm-check.rs
TornaxO7: Can I assume that those problems don't exist anymore due to your PR?

PR #950, opened 2022-11-07
added lefthk feature checks and errors

OfekShochat:
# Description
add lefthk feature check to leftwm-check, and error reporting.

## Type of change

### DIFF
--- a/leftwm/src/bin/leftwm-check.rs
+++ b/leftwm/src/bin/leftwm-check.rs
@@ -46,7 +46,12 @@ async fn main() -> Result<()> {
         return Ok(());
     }
 
-    check_enabled_features();
+    match check_enabled_features() {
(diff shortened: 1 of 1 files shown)


VuiMuich:
Just one thing: if features is empty, i.e. when building with `--no-default-features`, we should probably state this explicitly.

OfekShochat:
Ok, what should it write instead?

OfekShochat:
Also, no default features doesn't say it wasn't built with any. So, should I return when the default feature is disabled?

OfekShochat:
ran test-full this time, sorry

VuiMuich (changes requested):
Despite those minor suggests, LGTM

VuiMuich (approved):

PR #956, opened 2022-11-16
Print `lefthk` warning only when no binary present

VuiMuich:
# Description

Only print the warning message about necessary session restart only when no binary for `lefthk-worker` is found.
To be precise, we don't print on any error when trying to launch `lefthk-worker` as this means the same for the session state I guess.

## Type of change

# Checklist:

### DIFF
--- a/leftwm/build.rs
+++ b/leftwm/build.rs
@@ -1,4 +1,4 @@
-use std::{env, process::Command};
+use std::env;
 
 fn main() {
     let mut features_string = String::new();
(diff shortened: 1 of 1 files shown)


Eskaan (approved):
LGTM, but as you are already on it, isn't there also a `warn!` macro?

VuiMuich:
> LGTM, but as you are already on it, isn't there also a `warn!` macro?

Afaik the `warn!` macro is for logging, but we want to print this message during the build process. If you know of a macro to drop a warning during compilation, I would be happy to consider it.

PR #980, opened 2022-12-08
Fix manual install instructions

TH3-S4LM0N:
### Description
As the manual install instructions say to use the `optimized` profile, cargo places the builds in `./target/optimized`, not `./target/release`.


Eskaan (approved):
Ah, I see.
So there was a line that I missed when I implemented it.
LGTM.

PR #985, opened 2022-12-09
Using args[0] instead of current_exe() under OpenBSD. current_exe() is not supported in this OS.

### LINKED ISSUE #983: LeftWM crashes because `current_exe()` is not supported in OpenBSD
### Describe the bug

It is a known issue that `current_exe()` is not supported in Rust under OpenBSD. The function crashes. 

The reason is that the OS itself does not provide any way to reliably know the current exe path of the binary. OpenBSD developers refuse to implement a function like this because they are not sure if they can implement it correct enough.

For the LeftWM use case, maybe we could user use `arg[0]` passed to `canonicalize()` or at least fall back to it if `current_exe()` fails or if the OS is OpenBSD.

_No response_

Run LeftWM under OpenBSD, redirecting the CLI log to a file. You will see a panic when calling `expect` after `current_exe()`.

_No response_

### OS / Distro

OpenBSD 7.2

_No response_

### LeftWM Check

```text
:: LeftWM version: 0.4.0
:: LeftWM git hash: de88c4a-modified
:: Built with no enabled features.
    -> Configuration loaded OK
WARN: Ignoring checks on keybinds as you compiled for an external hot key daemon.
:: Checking environment . . .
    -> Environment OK (has XDG_RUNTIME_DIR)
:: Checking theme . . .
```

igor-petruk:
# Description

As per discussion in https://github.com/leftwm/leftwm/issues/983, LeftWM currently cannot work in OpenBSD. It is due to known issue that a Rust function `current_exe()` is not implemented in OpenBSD and panics. For LeftWM purposes it is sufficient to just use `args[0]` instead, as this exe path is used only to find paths to other binaries and launch them. It is verified that it works in OpenBSD. The `current_exe()` approach is still more correct, so it is preserved for all other OS.

Fixes #983 

## Type of change

  - Fixes the bug on OpenBSD
  - Well, adds the support for OpenBSD in the first place.
  - There is no suitable place to say it now supports OpenBSD. There are installation instructions mentioning various OS, but they are about package manager use, not about the ability to run the WM. LeftWM is not yet packaged for OpenBSD.

# Checklist:

   -  This Makefile is not ready to run on OpenBSD, fails with syntax errors, but I have build a binary, run it and now I am typing from LeftWM under OpenBSD.
  - N/A
  - N/A


VuiMuich:
Looking good (besides me not really being able to confirm this works on `OpenBSD`).
Just one minor thing, could you add an `OpenBSD` section to the install instructions in the `README.md`, please?

igor-petruk:
Done. Please let me know if it looks good. I've added some useful tips that can save some users time.

igor-petruk:
Correction, my mistake. I think leftwm-theme is failing to build not because of nightly, but for other reason, related to my setup (My LibreSSL is too new for the latest openssl-sys). 

Let me for now remove any mention of leftwm-theme, because maybe it compiles fine. I will later verify this. Users will figure out to use manual theme management if they cannot compiles leftwm-theme

igor-petruk:
Hi, is there anything else I need to add here?

VuiMuich:
> Hi, is there anything else I need to add here?

Hey, I don't miss anything I guess, just was waiting for other PRs to be ready/confirmed running well to do the merges all at once. If we have a reason to proceed with this one immediately there is nothing blocking, but if not I'd like to wait.

igor-petruk:
No particular reason. No hurry, take your time, just wanted to check if
there is anything left.

igor-petruk:
Thank you very much 🙏

VuiMuich:
Thank you for your contribution! :pray:

If you could have a quick look if there is a place in the wiki that would be in need of update in regard of this PR, thanks.

PR #987, opened 2022-12-10
Swap for new copr repo

TH3-S4LM0N:
### Description
As the atim/leftwm copr has been broken since November I created a new one (`th3-s4lm0n/leftwm`). I would appreciate if anybody using Fedora could test this for a day or two because "it works on my machine" isn't especially reassuring to me. This is the first copr I've made so some testing can't hurt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,6 @@ paru -S leftwm
 ```sh
 sudo dnf copr enable th3-s4lm0n/leftwm -y && sudo dnf install leftwm
 ```
-**Important note: currently the copr package is broken due to missing `lefthk-worker` in the installation. We are on it to get the package fixed. In the meantime please look into cargo or manual installation.**
 
 ## NetBSD ([Official repositories])
 


VuiMuich:
Awesome, thanks!

tagging @NicTanghe for a test spin if possible.

NicTanghe:
Hello,
I have switched to RHEL8 and don't really have spare storage space.

VuiMuich:
I guess we proceed then, and react if there is some complaints.

Thanks a lot for the effort @TH3-S4LM0N!

VuiMuich:
Hey @TH3-S4LM0N, just wanted to ping you, to let you know we bumped to `0.4.1`. Not sure if that requries any manual work on the copr package. At least its no such severe things in this update as it was with the `0.4.0` release :joy:

TH3-S4LM0N:
I just have to bump the version number, thanks for notifying me though!

TH3-S4LM0N:
@VuiMuich is `0.4.1` going to have a github tag/release?

VuiMuich:
The tag is already published, but the release is in the making. 
I just learned, that the release has its own hash, so you might want to wait for it to be published.
Sorry for waking you early :grimacing:

Issue #971, opened 2022-11-28
SwapTags without having visited a previous tag causes error


linde12:
### Describe the bug

I'm getting some weird behavior when doing SwapTags without actually having visited another tag yet.

This swaps the main window with one of the windows in the right stack and gives some error message output above my polybar saying something like IoError ConnectionRefused.

### Expected behavior (optional)

No-op, should do nothing.

### Steps to reproduce (optional)

1. Start or reload LeftWM
2. Spawn some windows, e.g. browser and terminal
3. Do SwapTags (default modkey+w) without visiting any other tag first

### OS / Distro

Linux botnet 6.0.9_1 #1 SMP PREEMPT_DYNAMIC Fri Nov 18 16:45:48 UTC 2022 x86_64 GNU/Linux (Void Linux)

### Additional System Information (optional)

```text
Built from master v0.4.0 with --features=lefthk,syslog, running Void Linux. Added some traces to debug another issue (hence modified from leftwm-check) Using basic_polybar theme.
```

### LeftWM Check

```text
:: LeftWM version: 0.4.0
:: LeftWM git hash: a631cfb-modified
:: Enabled features: lefthk lefthk-core syslog-tracing sys-log
:: Checking feature dependencies . . .
    -> lefthk OK
:: Loading configuration . . .
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK (has XDG_RUNTIME_DIR)
:: Checking theme . . .
ERROR: Could not parse theme file: 1:16: Expected option
```

VuiMuich:
:thinking: actually this method starts with exactly a check if the history is at least two items long:
https://github.com/leftwm/leftwm/blob/de88c4a5d8463f187c8b2ac7b1cea0aff85c1178/leftwm-core/src/handlers/command_handler/mod.rs#L361-L383

Can you possibly grab the exact error from polybar logs?

linde12:
Ah, maybe its unrelated. I will try to find the logs for polybar once im home and post them here!

linde12:
Unable to reproduce the error message right now, but the weird behavior with windows switching places still happens. Maybe you can reproduce this issue?

VuiMuich:
Sorry for being a little late with my reply.
I forget to test this with a fresh session, but I just set up a new VR with Arch the other day, I can try to reproduce it there.

PR #984, opened 2022-12-09
replace `<module-name>/mod.rs` with `<module-name>.rs`

VuiMuich:
# Description

This replaces any `<module-name>/mod.rs` with `<module-name>.rs` and the acompanying `<module-name>` directory.
I saw this recently in a bunch of other repos and thought this is nice.
I allways found it odd, to use the `mod.rs` way in most scenarios.

This is difinitelly a controversial PR and I am happy to follow the common oppinion, so please vote with :+1: :-1: if you like or dislike.

## Type of change

- [x] Development change (no change visible to user)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation only update (no change to the factual codebase)
- [ ] This change requires a documentation update

## Updated user documentation:

Please insert user documentation that should be updated (as in the wiki).

See [CONTRIBUTING.md User Documentation section](../CONTRIBUTING.md#user-documentation) for further details.

**Note: Manual page changes must be performed in a commit, not in this PR section.**

# Checklist:

- [x] Ran `make test-full` locally with no errors or warnings reported
  Note: To fully reproduce CI checks, you will need to run `make test-full-nix`. Usually, this is not neccesary.
- [ ] Manual page has been updated accordingly
- [ ] Wiki pages have been updated accordingly (to perform **after** merge)

PR #988, opened 2022-12-10
Fix crash when swapping tag at startup

cecton:
# Description

When leftwm has just started and you swap tag (the shortcut that swaps between
the last and the current tag), it crashes on the `unwrap()`. This is because
there is no last tag in the history.

## Type of change

Bug fix

I guess none.

# Checklist:


VuiMuich:
Thanks!

This should close #971 then.

VuiMuich:
Thanks for fixing their merge conflict.

cecton:
There was none locally :woman_shrugging: I guess github is not using the same conflict resolution algorithm than git

VuiMuich:
:thinking: this is odd. Looking at the of the first commits' diff in the browser it looks like github was thinking this PR is based on a merged #984.

cecton:
GitHub works in mysterious ways sometimes :woman_shrugging: 

It was based on an older commit though. The parent commit is 2840e5ace1f7e293d9c2e37feebcc31b71b888e0

Well. Either way, it works and it's nice that I accidentally fixed some issue haha xD

VuiMuich:
Oh, now I get it, I guess. I the mean time there was some refactor on the commandhandler, that modularised it with the subdirectory structure and it was by pure chance, that almost at the same time you created this PR, I opened #984, wich renames all the `mod.rs` files, so I guess it got me confused for a moment.

PR #991, opened 2022-12-23
handle theme templates within Makefile

VuiMuich:
# Description

Recently in a thread on the discord, @mautamu asked whether the theme templates of the repo get installed to a general accessible path.
With this PR the makefile installs the theme templates from the repo directory `./themes` to `/usr/share/leftwm` (and removes the directory with the uninstall), same as both our AUR packages do.

Should we reflect this in the Docs somewhere?

## Type of change

# Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # - credit: https://stackoverflow.com/a/23324703/2726733
 ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
-SHARE_DIR := /usr/share/xsessions
+SHARE_DIR := /usr/share
 TARGET_DIR := /usr/local/bin
 
 # Set default profile if unset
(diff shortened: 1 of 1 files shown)


Eskaan (commented):
Two things I want to ask before testing this:
- `make install-linked` does not link but copy the theme files?
- Shouldn't we specify `--parents` for `sudo mkdir`?

VuiMuich:
Good points and with half a brain cell I had thought of that, too, but maybe I was wrong:
> - `make install-linked` does not link but copy the theme files?

Manpage and `.desktop` get copied as well, also not sure, how `leftwm-theme` would work with the link.
Maybe it's worth testing if we actually could link everything eventually...

> - Shouldn't we specify `--parents` for `sudo mkdir`?

Is it realistically possible the `/usr/share` doesn't exist?

PR #996, opened 2022-12-25
Apply changes suggested by clippy trough `make test-full`

blackdragon2447:
# Description

Simply apply the changes clippy suggests when running `make test-full`, which could be done mostly automatically.

## Type of change

# Checklist:

### DIFF
--- a/leftwm-core/src/models/tag.rs
+++ b/leftwm-core/src/models/tag.rs
@@ -287,9 +287,8 @@ impl Tag {
     /// ## Arguments
     /// * `delta` - increase/decrease main width percentage by this amount
     pub fn change_main_width(&mut self, delta: i8) {
-        self.main_width_percentage = (self.main_width_percentage as i8 + delta)
-            .max(0) // not smaller than 0
-            .min(100) as u8; // not larger than 100
+        self.main_width_percentage = (self.main_width_percentage as i8 + delta).clamp(0, 100) as u8;
+        // not smaller than 0 and not larger than 100
     }
 
     /// Sets the main width percentage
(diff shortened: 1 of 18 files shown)


leftwm-core/src/models/tag.rs
```diff
@@ -287,9 +287,8 @@ impl Tag {
     /// ## Arguments
     /// * `delta` - increase/decrease main width percentage by this amount
     pub fn change_main_width(&mut self, delta: i8) {
-        self.main_width_percentage = (self.main_width_percentage as i8 + delta)
-            .max(0) // not smaller than 0
-            .min(100) as u8; // not larger than 100
+        self.main_width_percentage = (self.main_width_percentage as i8 + delta).clamp(0, 100) as u8;
```
VuiMuich, line 290: Here I'd prefer to swap lines with the comment

leftwm/src/config/checks.rs
```diff
@@ -100,8 +100,8 @@ impl Config {
                 match error.0 {
                     Some(binding) => {
                         println!(
-                            "\x1b[1;91mERROR: {} for keybind {:?}\x1b[0m",
-                            error.1, binding
+                            "\x1b[1;91mERROR: {} for keybind {binding:?}\x1b[0m",
```
VuiMuich, line 103: Interesting, why the `error.1` is not pulled in. Is this because of the `.1`?

VuiMuich (commented):
Just the two nitpicks from my inner pedantic 😄

leftwm/src/config/checks.rs
VuiMuich: Just had a look at this, obviously the compiler does not process the tuple selection, when inside the curlies...

Issue #516, opened 2021-10-10
multiple master windows


hertg:
This feature request is inspired by the same feature from `dwm`.

**Feature**
Allow users to have multiple master windows instead of just a single one, and allow changing the amount of master windows via command/keybinding.

Right now, most of the layouts consist of a single master window and a "stack" with all the other windows (Monocle and CenterMain obviously being a little bit different, but they still have a single "master" window). The master window is always the one on top of the stack of all windows.

My suggestion is to allow to increase and decrease the amount of windows considered "master" windows via command/keybing.

![master-stack](https://user-images.githubusercontent.com/3533648/136689350-dc9997e8-73ac-453b-b281-00355823ca35.png)

**Reasoning**
Similar to the feature where the master window width can be changed dynamically, this introduces even more flexibility to all the layouts. A user could specify keybindings like <kbd>mod + i</kbd> (increase) and <kbd>mod + d</kbd> (decrease) to dynamically adjust how many windows appear in the "master zone" of the layout.

I only really understood how much I've actually used this feature in `dwm` after I switched to LeftWM completely :stuck_out_tongue_closed_eyes:. Below you see some examples on how that behaves in `dwm`.

*Example in CenterMain layout (master on the left)*
![output1](https://user-images.githubusercontent.com/3533648/136688933-7e8e0916-3556-4bef-a1dd-4352e86e46a2.gif)

*Example in VertAndStack layout (master in center)*
![output2](https://user-images.githubusercontent.com/3533648/136688944-60f90fa0-ccdc-444f-9ed2-a1b94735e970.gif)

AethanFoot:
Not quite, it doesn't do anything to aid the current layout files. It currently just holds the layout mode and selected layouts, and holds functions that require the layout list (just new_layout, and next/prevoius_layout). I am currently trying to make things a bit cleaner while waiting on lex to be able to it again.

hertg:
Implemented with #1000

PR #1000, opened 2022-12-26
Integrate leftwm-layouts library

### LINKED ISSUE #96: Changing main width doesn't work in fibonacci layout

### LINKED ISSUE #545: rename IncreaseMainWidth to IncreaseMainSize
hello I have not found a way to modify the height of the windows.
Something similar to IncreaseMainWidth.
Would there be some way

### LINKED ISSUE #762: Issues regarding main width percentage and `LeftWiderRightStack` layout

hertg:
> **Note** Not ready for review, just wanted to snatch #1000 :joy: 

## :sparkles: Features

- Extract layout calculations to [leftwm-layouts](https://github.com/leftwm/leftwm-layouts)
- **New commands** available `IncreaseMainCount` and `DecreaseMainCount` (#516)
- Any layouts' default values can be changed via configuration (`main_width`, `rotation`, `flip`, etc)
- New custom layouts can be defined via configuration

## :lady_beetle: Fixes

- Fixes #96 
- Fixes #545 
- Fixes #762 

## :boom: Breaking Changes

- The `layouts` (also `workspace.layouts`) in `config.ron` must be declared as Strings because they are no longer backed by an enum. (ie. `layouts: ["CenterMain"]` instead of `layouts: [CenterMain]`)
- The `max_window_width` was removed without direct replacement. The "always fill entire screen" behavior can now be addressed in layouts with the `Reserve` property.

## :warning: Deprecated

- The `IncreaseMainWidth` and `DecreaseMainWidth` commands have been deprecated, use `IncreaseMainSize` and `DecreaseMainSize` instead

## :ballot_box_with_check: Potential Follow-Ups

The following can be created as issues to address in the future. I do not think that those are important enough to keep this from being merged.

- [ ] Restore layouts on `SoftReload` (Currently, all layouts reset from config on `SoftReload`)
- [ ] Implement Command that allows to reset a layout to it's default properties (nice-to-have)
- [x] Make specifying custom layouts easier by [allowing to omit default values](https://github.com/leftwm/leftwm/pull/1000#issuecomment-1556296180) in the ron config


VuiMuich:
Looking very much forward whats to come from this PR.

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -369,9 +373,9 @@ fn swap_tags(state: &mut State) -> Option<bool> {
         std::mem::swap(&mut state.workspaces.get_mut(hist_a)?.tag, &mut temp);
         // Update dock tags and layouts.
         state.update_static();
-        state
-            .layout_manager
-            .update_layouts(&mut state.workspaces, state.tags.all_mut());
```
hertg, line 374: What is the point of the `update_layouts` method? Is it just so the `tag.layout` and `workspace.layout` are kept the same? I guess my refactoring makes this obsolete with the new layout manager.

hertg:
I updated the PR description with the things I still have on my mind as TODOs. Not sure if they must necessarily be included in this PR or if they could be addressed later. I think handling the `SoftReload` in a better way would probably be good to be included here, the other is more of an afterthought. I will try to address that and then get the PR ready.

A big part of this PR is the code inside [leftwm-layouts](https://github.com/leftwm/leftwm-layouts) which will be imported into leftwm. Giving the code there also a look as part of reviewing this PR would be appreciated.

hertg:
It seems Workspace IDs are currently not unique, which contradicts some assumptions I made in this PR. Waiting on feedback on [my comment](https://github.com/leftwm/leftwm/pull/848#issuecomment-1436081510).

leftwm-core/src/layouts/layout_mode.rs
```diff
@@ -0,0 +1,16 @@
+use serde::{Deserialize, Serialize};
+
+/// Leftwm now has 2 layout modes, Workspace and Tag. These determine how layouts are remembered.
```
VuiMuich, line 3:
```suggestion
/// Leftwm has 2 layout modes, Workspace and Tag. These determine how layouts are remembered.
```
This temporal reference makes no sense in doc comments, right?

VuiMuich:
Just did a first test build of this branch (merged current state of #1043 and main branch) and noticed those topics which can partially be considered regressions:
- `leftwm-state` no longer prints the current layout
- when setting up layouts in `config.ron` the Layouts need to be Strings, before they where their own types, which felt nicer imo

hertg:
I think this one is getting ready now. I fixed two of the larger things that were on my mind yesterday, namely to support layouts per workspaces and fixing a positioning issue that happened when a workspace had an offset.

@VuiMuich 
- The `layout` should now be included again in `leftwm-state`.
- Yep, the layouts config is now a list of Strings. The names of layouts can now even be chosen freely if one were to create a custom layout in the config, which is why it can no longer be a hardcoded enum. 

The layouts being Strings is one of the breaking changes this PR brings. Unless there is some flag that lets `ron` parse unquoted values as Strings, which I don't think there is, there will be no way around that.

VuiMuich:
Awesome 😎 

Maybe the `ron` [extension](https://github.com/ron-rs/ron/blob/master/docs/extensions.md) `unwrap_newtype` might work?

I'll give this a spin asap.

hertg:
> Awesome sunglasses
> 
> Maybe the `ron` [extension](https://github.com/ron-rs/ron/blob/master/docs/extensions.md) `unwrap_newtype` might work?
> 
> I'll give this a spin asap.

I mean, if it requires users to add the `#![enable(unwrap_newtypes)]` manually, there wouldn't be much of a point to it :stuck_out_tongue_closed_eyes:

Eskaan:
We can add it to the ron extensions that we activate by default.

Also I feel like @VuiMuich is trying to fing excuses to throw toml out, isn't that so? ;)

hertg:
I tried quickly to change the `layouts` type to a custom type, and it didn't seem to work. Maybe I did it wrong, but I won't go further down that path for now.

I tried something to the effect of:

```rust
pub struct LayoutName(String);

pub struct Config {
    ...
    pub layouts: Vec<LayoutName>,
}
```

Which still worked when the layout names are in quotes, but doesn't work after removing the quotes.

VuiMuich:
Yeah, activating the extension can be done at parser configuration, same as the `implicit_some`. 

Hey I am good with finding excuses 🤷🏼‍♂️, also moving a ax from toml to allow more complex configuration possibilities was the plan from the start on.

VuiMuich:
Well to be fair, I haven't had a look, how you are handling this matter in your lib, but from a quick glance at `leftwm-layout/src/layouts/default.rs` I suppose they are `&str` everywhere, right?
So to enable this it would be probably be a bigger change on the lib side, so maybe we go with the array of strings in the config now, and later bring it back as custom types...

leftwm-core/src/command.rs
```diff
@@ -69,10 +66,14 @@ pub enum Command {
...
     PreviousLayout,
-    SetLayout(Layout),
+    SetLayout(String),
     RotateTag,
-    IncreaseMainWidth(i8),
-    DecreaseMainWidth(i8),
+    IncreaseMainWidth(i32), // deprecated: use IncreaseMainSize instead
+    DecreaseMainWidth(i32), // deprecated: use DecreaseMainSize instead
+    IncreaseMainSize(i32),
+    DecreaseMainSize(i32),
```
hertg, line 74: *Note: Changed to `i32` because, although unlikely, this could now also be a pixel value. And we probably want to support pixel values higher than `255`*

leftwm-core/src/utils/command_pipe.rs
```diff
@@ -287,18 +288,16 @@ fn build_move_window_to_previous_tag(raw: &str) -> Result<Command, Box<dyn std::
...
 
-fn build_decrease_main_width(raw: &str) -> Result<Command, Box<dyn std::error::Error>> {
-    let headless = without_head(raw, "DecreaseMainWidth ");
-    let mut parts = headless.split(' ');
-    let change: i8 = parts.next().ok_or("missing argument change")?.parse()?;
-    Ok(Command::DecreaseMainWidth(change))
+fn build_decrease_main_size(raw: &str) -> Result<Command, Box<dyn std::error::Error>> {
+    let mut parts = raw.split(' ');
+    let change: i32 = parts.next().ok_or("missing argument change")?.parse()?;
+    Ok(Command::DecreaseMainSize(change))
```
hertg, line 300: *Note: The call to `without_head` was unnecessary because the `raw` parameter doesn't contain the command, it's already split out before this method is called.*

hertg:
@VuiMuich @AethanFoot @Eskaan I consider this ready :)

VuiMuich:
I've been running this branch for a couple of days now, and it's as stable as ever. 

@hertg could you give an example how a custom layout is configured, please? If there is one somewhere in this PR or in the lib I fail to see it.

I would really really like to once more investigate, if we could avoid the "layouts in the config are Strings" breakage, as I believe (hope?) this should be possible. I'll try to get a deeper look at this asap.

hertg:
@VuiMuich Theoretically, you should be able to add custom layouts with the `layout_definitions` config, but for some reason it doesn't seem to work.That's what it should look like:

```rust
layout_definitions: [
    (
        name: "Foo",
        flip: None,
        rotate: North,
        reserve: ReserveAndCenter,
        columns: (
            main: (count: 1, size: Ratio(0.5), flip: None, rotate: North, split: Horizontal),
            stacK: (flip: None, rotate: North, split: Vertical),
            second_stack: (flip: None, rotate: North, split: Horizontal),
        )
    )
]
```

I'm not sure why it doesn't work, and I currently don't have the capacity to track that down. The same goes for the "using strings for layout names" problem. I'd be more than happy if someone has some time on their hands and wants to track those two issues down.

VentGrey:
Tested this branch against Debian + Epitaph. Only had to change `IncreaseMainWidth` and `DecreaseMainWidth` as instructed. No further breakages reported :P

hertg:
I'm stupid, of course my config didn't work because I wrote `stacK` instead of `stack`, didn't notice the capital `K`. Using `leftwm-check` even would have helped me figure it out :man_facepalming: 

Here is a working example with a fully verbose definition:

```rust
layouts: [
    "Foo",
    "CenterMainFluid",
    "CenterMain",
    "Fibonacci",
    "MainAndDeck",
    "Monocle",
],
layout_definitions: [
    (
        name: "Foo",
        flip: None,
        rotate: North,
        reserve: ReserveAndCenter,
        columns: (
            flip: None,
            rotate: North,
            main: Some(Main(count: 1, size: 0.1, flip: None, rotate: North, split: Some(Horizontal))),
            stack: (flip: None, rotate: North, split: Some(Vertical)),
            second_stack: Some(SecondStack(flip: None, rotate: North, split: Horizontal)),
        )
    ),
]
```

Of course, some of the attributes are the default values and could be omitted. I kept them in this example on purpose to show all possible attributes. And I explicitly stated the `Some(..)` Option values to show which ones are `Option`s.  The definition and defaults have yet to be documented for end-users, but I plan on doing so later. That's why I communicated before that the new custom layouts is gonna be a "silent new feature" for now. The only available documentation for the layout definition is in the [leftwm-layouts source code](https://github.com/leftwm/leftwm-layouts/blob/main/leftwm-layouts/src/layouts/layout.rs) or on the [docs.rs page](https://docs.rs/leftwm-layouts/latest/leftwm_layouts/layouts/struct.Layout.html).

When one defines custom `layout_definitions`, it's important to notice that this does not automatically replace all default layouts (as can be seen in my example above, it simply adds the additional `Foo` layout). But it is still possible to actually overwrite the default layouts by defining a layout with the same name.

---

I also recommend giving the new "multiple main windows" a test-run. I'm using the following keybinds for those new commands.

```rust
(command: IncreaseMainCount, value: "", modifier: ["modkey"], key: "i"),
(command: DecreaseMainCount, value: "", modifier: ["modkey"], key: "d"),
```

---

I might do a short video in the future explaining some things. Would probably be of much help to new coming adventurers who might be overwhelmed by an overly technical documentation.

hertg:
> Of course, some of the attributes are the default values and could be omitted.

hmm, maybe that's actually not true..

VuiMuich:
The example works well (just changed `size: 0.4`, as this is a bit more reasonable  (at least on my 16:9 Display).
Didn't have time to look into the typing matter yet, but possibly next or the week after.

hertg:
I found some minor bugs in the leftwm-layouts library that I want to fix before this gets merged.
@VuiMuich Have you figured anything out about the layout name parsing? So we could prevent the necessity to enclose them in quotes?

Also, it would be nice if the fields on a custom `layout_definition` could be omitted if it is the default value anyway. It seems like that's not possible right now, so all struct fields have to be explicitly specified, which is very verbose. Is that possible? Basically an implicit `..Default::default()` for the struct and sub-structs.

---

eg. in my example from above
```rust
(
    name: "Foo",
    flip: None,
    rotate: North,
    reserve: ReserveAndCenter,
    columns: (
        flip: None,
        rotate: North,
        main: Some(Main(count: 1, size: 0.1, flip: None, rotate: North, split: Some(Horizontal))),
        stack: (flip: None, rotate: North, split: Some(Vertical)),
        second_stack: Some(SecondStack(flip: None, rotate: North, split: Horizontal)),
    )
)
```

it shall be possible to specify it like that (omitting anything that's the default anyway)
```rust
(
    name: "Foo",
    reserve: ReserveAndCenter,
    columns: (
        main: Some(Main(size: 0.1, split: Some(Horizontal))),
        stack: (split: Some(Vertical)),
        second_stack: Some(SecondStack),
    )
)
```

hertg:
I fixed the issues I wanted to address in `leftwm-layouts` with todays release `0.8.2`.
Now, I also enabled the `unwrap_newtypes` ron extensions by default, It doesn't help with the breaking change of layout names in quotes, but it makes specifying layout_definitions a little bit less cumbersome. I did not enable `unwrap_variant_newtypes` by default, because I think it could lead to unexpected behavior. See [extensions.md](https://github.com/ron-rs/ron/blob/master/docs/extensions.md) for more info.

hertg:
> Do you mean default as in config or as in unset- defaults? I suppose you mean the values in the config, but I am not sure about that.

When you change the `main` size, or change the amount of `main` windows, or when you rotate the layout at runtime, all those things directly change the layout properties (eg. `Layout::Main::size`). So what I mean is reset the layout back to its default properties, effectively resetting all user-made changes applied during runtime.

This is not really a big deal, because currently a `SoftReload` resets all layouts. But if a `SoftReload` ever does restore the modifications made during runtime, such a feature might be helpful. Just in case anybody morphs their layout into something unrecognizable from the original so that they want to reset :laughing:

Eskaan:
I want to note that the version of `leftwm-layouts` in `leftwm` is out-of-date whilst you updated `leftwm-core`

hertg:
You are right, good catch. Thanks @Eskaan :)
Merged your PR, published and updated the lib version here.

VuiMuich:
I am not sure yet if that't related to this PR but it happens that when creating a new window it gets not immediatelly rendered, only after performing a focus action it gets placed. Can somebody else reproduce this? I have to check if it is possibly related to `MainAndDeck` spevifically.

VuiMuich:
On the "list of strings in config" topic:
A naive idea would be to change the `name` in the `layout_description` to from a `Sting` to type `Layout` and see if ron-serde can figure out the rest?

VuiMuich:
Thank you so much @hertg for this milestone commit!!! :tada: :partying_face:

v3ss0n:
Awesome! gonna try.

PR #1002, opened 2022-12-29
Fix regression on calculating window score

guigot:
# Description

#986 (fe42fd0c0b6137823541510b3e118a52b28c673b) seems to introduce a small regression about calculating score window : for the class window, res_class and res_name was used and for the title window, name and legacy_name. Now, only res_class and legacy_name are used. This PR reintroduce the old behaviour.

## Type of change

# Checklist:

### DIFF
--- a/leftwm/src/config.rs
+++ b/leftwm/src/config.rs
@@ -106,8 +106,6 @@ impl WindowHook {
         let class_score = self.window_class.as_ref().map_or(0, |re| {
             u8::from(matches_any(re, vec![&window.res_class, &window.res_name]))
         });
-            _ => 0,
-        };
 
         let title_score = self.window_title.as_ref().map_or(0, |re| {
             u8::from(matches_any(re, vec![&window.legacy_name, &window.name]))
--- a/leftwm/src/config.rs
+++ b/leftwm/src/config.rs
@@ -97,13 +97,21 @@ impl WindowHook {
     /// Multiple [`WindowHook`]s might match a `WM_CLASS` but we want the most
     /// specific one to apply: matches by title are scored greater than by `WM_CLASS`.
     fn score_window(&self, window: &Window) -> u8 {
-        let class_score = match (&self.window_class, &window.res_class) {
-            (Some(wc_re), Some(res_class)) => u8::from(wc_re.replace(res_class, "") == ""),
+        let class_score = match (&self.window_class, &window.res_class, &window.res_name) {
+            (Some(wc_re), Some(res_class), Some(res_name)) => {
+                u8::from(wc_re.replace(res_class, "") == "" || wc_re.replace(res_name, "") == "")
+            }
+            (Some(wc_re), Some(res_class), _) => u8::from(wc_re.replace(res_class, "") == ""),
+            (Some(wc_re), _, Some(res_name)) => u8::from(wc_re.replace(res_name, "") == ""),
             _ => 0,
         };
 
-        let title_score = match (&self.window_title, &window.legacy_name) {
-            (Some(wt_re), Some(legacy_name)) => u8::from(wt_re.replace(legacy_name, "") == ""),
+        let title_score = match (&self.window_title, &window.legacy_name, &window.name) {
+            (Some(wt_re), Some(legacy_name), Some(name)) => {
+                u8::from(wt_re.replace(legacy_name, "") == "" || wt_re.replace(name, "") == "")
+            }
+            (Some(wt_re), Some(legacy_name), _) => u8::from(wt_re.replace(legacy_name, "") == ""),
+            (Some(wt_re), _, Some(name)) => u8::from(wt_re.replace(name, "") == ""),
             _ => 0,
         };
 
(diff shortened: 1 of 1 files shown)


hertg (approved):

leftwm/src/config.rs
```diff
@@ -97,13 +97,17 @@ impl WindowHook {
     /// Multiple [`WindowHook`]s might match a `WM_CLASS` but we want the most
     /// specific one to apply: matches by title are scored greater than by `WM_CLASS`.
     fn score_window(&self, window: &Window) -> u8 {
-        let class_score = match (&self.window_class, &window.res_class) {
-            (Some(wc_re), Some(res_class)) => u8::from(wc_re.replace(res_class, "") == ""),
+        let class_score = match (&self.window_class, &window.res_class, &window.res_name) {
+            (Some(wc_re), Some(res_class), Some(res_name)) => {
```
hertg, line 101: Without me knowing any better; Can we assume that `window.res_class` and `window.res_name` are always both present? Because if one of them doesn't, this match arm is not reached. This might potentially fix the regression but introduce another.

leftwm/src/config.rs
```diff
@@ -97,13 +97,17 @@ impl WindowHook {
...
+                u8::from(wc_re.replace(res_class, "") == "" || wc_re.replace(res_name, "") == "")
+            }
             _ => 0,
         };
 
-        let title_score = match (&self.window_title, &window.legacy_name) {
-            (Some(wt_re), Some(legacy_name)) => u8::from(wt_re.replace(legacy_name, "") == ""),
+        let title_score = match (&self.window_title, &window.legacy_name, &window.name) {
+            (Some(wt_re), Some(legacy_name), Some(name)) => {
```
hertg, line 108: Same as above

hertg (changes requested):

leftwm/src/config.rs
```diff
@@ -97,13 +97,21 @@ impl WindowHook {
...
     /// specific one to apply: matches by title are scored greater than by `WM_CLASS`.
     fn score_window(&self, window: &Window) -> u8 {
-        let class_score = match (&self.window_class, &window.res_class) {
-            (Some(wc_re), Some(res_class)) => u8::from(wc_re.replace(res_class, "") == ""),
+        let class_score = match (&self.window_class, &window.res_class, &window.res_name) {
+            (Some(wc_re), Some(res_class), Some(res_name)) => {
+                u8::from(wc_re.replace(res_class, "") == "" || wc_re.replace(res_name, "") == "")
+            }
+            (Some(wc_re), Some(res_class), _) => u8::from(wc_re.replace(res_class, "") == ""),
+            (Some(wc_re), _, Some(res_name)) => u8::from(wc_re.replace(res_name, "") == ""),
```
hertg, line 105:
```suggestion
        // returns true if any of the items in the provided `Vec<&Option<String>>` is Some and matches the `&Regex`
        let matches_any = |re: &Regex, strs: Vec<&Option<String>>| {
            strs.iter()
                .any(|str| str.as_ref().map_or(false, |s| re.replace(s, "") == ""))
        };

        let class_score = self.window_class.as_ref().map_or(0, |re| {
            u8::from(matches_any(re, vec![&window.res_class, &window.res_name]))
        });
```

leftwm/src/config.rs
hertg: Your updated code should work, the only downside is that there is some duplicate code and that the syntax is a bit hard to understand at a glance. But the syntax was hard to understand to begin with IMO. I will give you a suggestion on how I'd refactor the method, but there really is no right or wrong and it mostly comes down to personal preference.

leftwm/src/config.rs
```diff
@@ -97,13 +97,21 @@ impl WindowHook {
...
-        let title_score = match (&self.window_title, &window.legacy_name) {
-            (Some(wt_re), Some(legacy_name)) => u8::from(wt_re.replace(legacy_name, "") == ""),
+        let title_score = match (&self.window_title, &window.legacy_name, &window.name) {
+            (Some(wt_re), Some(legacy_name), Some(name)) => {
+                u8::from(wt_re.replace(legacy_name, "") == "" || wt_re.replace(name, "") == "")
+            }
+            (Some(wt_re), Some(legacy_name), _) => u8::from(wt_re.replace(legacy_name, "") == ""),
+            (Some(wt_re), _, Some(name)) => u8::from(wt_re.replace(name, "") == ""),
             _ => 0,
         };
```
hertg, line 116:
```suggestion
        let title_score = self.window_title.as_ref().map_or(0, |re| {
            u8::from(matches_any(re, vec![&window.legacy_name, &window.name]))
        });
```

hertg (changes requested):

hertg (approved):
@guigot I took the liberty and committed my syntax proposal to your branch. Let me know what you think and feel free to suggest another approach.

Would be great if you could confirm the code working on your machine. I will merge the PR after I get your ok.

guigot:
Your code is far more elegant than mine and works fine locally so let's go.

PR #1005, opened 2022-12-29
Add Command SwapWindowTop 

### LINKED ISSUE #1004: Add SwapWindowTop command
### Describe the feature

Running the command should swap the focused window with the current top window, without affecting the list position of any other windows.

I was always under the assumption that `MoveWindowTop` would swap the currently focused window with the top window and thought there was a bug that this sometimes reorders all the other windows. But after looking at the code I realized that these may be different use-cases? I don't like the `MoveWindowTop` behavior because I use it when I want to pay attention to a particular application and move it into the primary position. I just want it to go there, and not push everything else down the entire stack. Because that way all windows (ie. those between the focused and main) move to a different position and I lose track of where things are.

emiliode:
# Description

This PR introduces a new Command `SwapWindowTop`, which swaps the current window with the window on top of the stack, without changing the order of the other windows. 

Fixes #1004 

## Type of change

Add section to (https://github.com/leftwm/leftwm/wiki/Config#keybind-commands) : 
### SwapWindowTop 
Swaps the focused window with the window on top of the stack in the current workspace.  Takes an Optional argument swap, if set to false, calling the command while the top window is focused results in no action. Otherwise it will swap the top window with the one below it.  By default swap is set to true 

Add row to the External Commands table with the same describtion

# Checklist:

### DIFF
--- a/leftwm-core/src/command.rs
+++ b/leftwm-core/src/command.rs
@@ -41,6 +41,9 @@ pub enum Command {
     MoveWindowTop {
         swap: bool,
     },
+    SwapWindowTop {
+        swap: bool,
+    },
     FocusNextTag,
     FocusPreviousTag,
     FocusWindow(String),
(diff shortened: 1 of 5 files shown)


hertg (approved):
lgtm, did you testrun it on your machine?

emiliode:
Yes it works as expected on my machine

hertg:
great, already looks unproblematic codewise, just wanted to ask. I'll go ahead and merge.
Thanks a lot for your contribution :)

hertg:
@emiliode Can you update the wiki to include the new command you added?

emiliode:
> @emiliode Can you update the wiki to include the new command you added?

done

PR #1009, opened 2023-01-02
snap cursor to bottom-right of the window when resizing

### LINKED ISSUE #1003: Reposition cursor to bottom-right corner when resizing
### Describe the enhancement

When resizing a window, the cursor should automatically jump to the windows' bottom right corner (like in dwm).

This prevents the awkward situation where the cursor is no longer hovering the window that is being resized, and prevents the issue where a window can not be made any smaller if the cursor already reached the top-left corner of the screen during resize.

Would be nice to have

nemalex:
# Description

This PR makes the cursor move to the bottom-right corner of the window in the beginning of a resize

Fixes #1003 

## Type of change

# Checklist:

### DIFF
--- a/leftwm-core/src/handlers/mouse_combo_handler.rs
+++ b/leftwm-core/src/handlers/mouse_combo_handler.rs
@@ -19,9 +19,17 @@ impl State {
         if let Some(window) = self.windows.iter().find(|w| w.handle == handle) {
             if !self.disable_tile_drag || window.floating() {
                 let modifier = utils::modmask_lookup::into_modmask(&self.mousekey);
+                let win = window.clone();
                 // Build the display to say whether we are ready to move/resize.
                 let act = self.build_action(modmask, button, handle, modifier);
                 if let Some(act) = act {
+                    if let DisplayAction::ReadyToResizeWindow(_) = act {
+                        let move_act = DisplayAction::MoveMouseOverPoint((
+                            win.x() + win.width(),
+                            win.y() + win.height(),
+                        ));
+                        self.actions.push_back(move_act);
+                    }
                     self.actions.push_back(act);
                     return false;
                 }


leftwm-core/src/handlers/mouse_combo_handler.rs
```diff
@@ -19,9 +19,17 @@ impl State {
         if let Some(window) = self.windows.iter().find(|w| w.handle == handle) {
             if !self.disable_tile_drag || window.floating() {
                 let modifier = utils::modmask_lookup::into_modmask(&self.mousekey);
+                let win = window.clone();
```
hertg, line 22:
```suggestion
                let bottom_right = (window.x() + window.width(), window.y() + window.height());
```

leftwm-core/src/handlers/mouse_combo_handler.rs
```diff
@@ -19,9 +19,17 @@ impl State {
...
                 // Build the display to say whether we are ready to move/resize.
                 let act = self.build_action(modmask, button, handle, modifier);
                 if let Some(act) = act {
+                    if let DisplayAction::ReadyToResizeWindow(_) = act {
+                        let move_act = DisplayAction::MoveMouseOverPoint((
+                            win.x() + win.width(),
+                            win.y() + win.height(),
+                        ));
```
hertg, line 30:
```suggestion
                        let move_act = DisplayAction::MoveMouseOverPoint(bottom_right);
```

hertg (changes requested):
Your PR looks and works great, thank you!

The only thing bugging me is the `window.clone()`. I made a suggestion on how we could prevent that. Let me know what you think about it.

hertg:
@VuiMuich Should we implement the `disable_cursor_reposition_on_resize` config preemptively or wait and see whether people scream when it changes?

I'm quite happy with LeftWM changing this behavior without providing the config, sometimes less is more, but ofc I'm also the one who opened the issue :joy:

leftwm-core/src/handlers/mouse_combo_handler.rs
nemalex: Yeah, it is sensible. I prefer this solution too.

hertg:
I'm merging this as is. I opened #1011 to track the config option

PR #1017, opened 2023-01-09
Notify user when CommandPipe's Drop impl fails to unlock pipe file

vrmiguel:
# Description

Notifies stderr when CommandPipe's Drop impl fails to open pipe file with `O_NONBLOCK`. Also stops cloning `self.pipe_file` when doing so

## Type of change

# Checklist:


VuiMuich:
From a brief glance on mobile this looks good to me.

PR #1028, opened 2023-01-30
Update package address for NetBSD

0323pin:
pkgsrc.se is no more :(
After over 20 years it says goodbye.
Follow the current state on the official cvs repository. Not as good looking but, it does the job.

# Description

Simple URL change.


0323pin:
pkgsrc.se is back online under the control of the NetBSD foundation.

0323pin:
@VuiMuich thanks! All good with pkgsrc.se now, it's under NetBSD control now :)

Now, let's get LeftWM rolling.

PR #1038, opened 2023-03-02
Fix `leftwm-check` outputting default and dependencies

TH3-S4LM0N:
Currently when `leftwm-check` outputs the enabled features and dependencies:
```
Enabled features: default journald-log lefthk lefthk-core tracing-journald
```
When the only enabled features are `journald-log` and `lefthk`. Using `feature = ["dep:dependency"]` is the intended way to do feature dependencies and therefore `leftwm-check` won't print all the dependencies it uses for those features:
```
Enabled features:  journald-log lefthk
```
I also trim `default` in `build.rs`.
This just basically makes `leftwm-check` a bit more readable.

### DIFF
--- a/leftwm/Cargo.toml
+++ b/leftwm/Cargo.toml
@@ -62,12 +62,12 @@ tempfile = "3.2.0"
 
 [features]
 default = ["journald-log", "lefthk"]
-lefthk = ["lefthk-core"]
+lefthk = ["dep:lefthk-core"]
 
 # logging features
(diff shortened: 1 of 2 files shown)


mautamu (approved):

VuiMuich:
Good catch!

Lgtm as well, will try to get a bunch of PRs merged the next days. This will go along.

VuiMuich:
Again, thanks for taking care with these small but valuable improvements!

PR #1039, opened 2023-03-05
Chore: Update dependencies, bump MSRV

mautamu:
# Description

This commit updates all dependencies, resolves new clippy lint errors, and bumps the minimum MSRV to 1.65.0

## Type of change

MSRV is now 1.65.0

# Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
     - name: Install MSRV
       uses: actions-rs/toolchain@v1
       with:
-          toolchain: 1.60.0
+          toolchain: 1.65.0
           override: true
 
     - name: Run cargo check
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add `SwapWindowTop` command as an alternative to `MoveWindowTop` (via #1005 by @emiliode)
 - Reposition mouse cursor to bottom-right corner of window when resizing (via #1009 by @nemalex)
+- The currently supported MSRV is 1.65.0
 
 ### Fixed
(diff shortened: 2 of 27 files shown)


Eskaan (changes requested):
Is it really necessary to mark all those functions and structs as `pub(crate)`? I know they may not be used anywhere else but does clippy enforce that?

~~Also, I think the `let Some(...) = ... else {...}` wasn't introduced in this but the version before. Can't we set the MSRV to this version?~~ My fault, this is the correct version.

mautamu:
> ~Also, I think the `let Some(...) = ... else {...}` wasn't introduced in this but the version before. Can't we set the MSRV to this version?~ My fault, this is the correct version.

Haha, I thought that too, hence why it took five commits to get that right.

> Is it really necessary to mark all those functions and structs as pub(crate)? I know they may not be used anywhere else but does clippy enforce that?

I'm not entirely sure, the GitHub Action Clippy and my local copy appear to disagree on whether this is necessary, but I didn't dig much further. Fwiw I think that these should be pub not pub(crate); do we want to bypass the lint on those files?

Eskaan:
If your local clippy disagrees, I assume you have stable rust instead of v1.65.0?

I will take a look myself later today at what exactly may cause clippy to recommend `pub(crate)`.

mautamu:
Yep! 
``` installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default)
1.65.0-x86_64-unknown-linux-gnu

active toolchain
----------------

nightly-x86_64-unknown-linux-gnu (default)
rustc 1.69.0-nightly (44cfafe2f 2023-03-03)
```

Curiously it missed a few `let _ = function()`s, but I went in and reversed all of the `pub(crate)` annotations.

VuiMuich (approved):

hertg:
lol, I just yesterday did basically the same work you did here because I missed this PR.
I'm gonna merge this soon if that's ok, and then after that do a PR I prepared that makes use of reusable workflows for the CI.

PR #1027, opened 2023-01-29
Fix leftwm-state not being de-/serialized

Eskaan:
# Description

The JSON Syntax, and thus serde_json only support Maps of `HashMap<String, _>`. My change in #848 changed this map to `HashMap<(String, usize), _>` and with this broke the serialization of the state struct.

This pr fixes the serialization by changing the implementation from a `HashMap` to a `Vec` that gets iterated over to get an element.

Fixes #1001 

## Type of change

- [ ] Development change (no change visible to user)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation only update (no change to the factual codebase)
- [ ] This change requires a documentation update

# Checklist:

- [x] Ran `make test-full` locally with no errors or warnings reported
  Note: To fully reproduce CI checks, you will need to run `make test-full-nix`. Usually, this is not neccesary.

PR #1043, opened 2023-03-07
Make Workspace IDs unique and predictable

hertg:
# Description

Follow up on #848, which made workspace IDs only unique per output, effectively requiring a composite key of `output + id` to identify a workspace. This PR introduces a simple and consistent way of assigning unique IDs to every workspace, as discussed in the comments of #848, particularly in [this comment](https://github.com/leftwm/leftwm/pull/848#issuecomment-1456900361).

**Workspace IDs**
The IDs of workspaces are assigned automatically by LeftWM. They are numeric and counted upwards from $[1..n]$. The workspaces specified in the [config](https://github.com/leftwm/leftwm/wiki/Config#workspaces) receive an ID first. Every configured workspace gets an ID, even if it is on an output that is currently not connected.

```rust
workspaces: [
    ( output: "HDMI-1", y: 0, x: 0, height: 1440, width: 1720 ), // ID: 1
    ( output: "HDMI-1", y: 0, x: 1720, height: 1440, width: 1720 ), // ID: 2
]
```

The count then continues for non-configured / auto-derived workspaces, if there are any. IDs for auto-derived workspaces are assigned in the order that LeftWM receives screen information from the system (xrandr, xinerama, ...).

---

Mergin this PR makes #1027 obsolete

## Type of change

*TODO* Logic behind automatic workspace ID assignment should probably be documented somewhere

# Checklist:


leftwm-core/src/handlers/screen_create_handler.rs
```diff
@@ -7,15 +7,10 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
     ///
     /// Returns `true` if changes need to be rendered.
     pub fn screen_create_handler(&mut self, screen: Screen) -> bool {
+        tracing::debug!("screen create: {:?}", screen);
```
hertg, line 10: to be removed

hertg:
Hey @0323pin
Most builds are currently failing because we run everything against `stable` rust channel for linux (which seems to be `1.67.1`). In order to satisfy clippy on this version, a few adjustments have to be made, like using the `let...else` feature. This however makes it so that the msrv action fails, since it is still on `1.60.0`.

So my question is:
1. Can we update the toolchain version for the msrv action, and if so, to what version? Is there a way for us to figure out what the currently latest available rust version is for the OS you are on? (I believe it's OpenBSD?)
2. How can we keep this action up-to-date so we don't run into this issue too much?

hertg:
@0323pin Some additional info

A test by @VuiMuich showed that we'd need to bump the minimum supported rust version to `1.65`. This would allow us to satisfy pedantic clippy on Rust's current `stable` channel. 

If that is not possible we'd need to add some `#[allow(clippy::needless_lifetimes)]` and `#[allow(clippy::manual_let_else)]` to the code.

What is our current philosophy on the minimum supported rust version?

0323pin:
@hertg thanks for asking.

> Can we update the toolchain version for the msrv action, and if so, to what version?

1.66.1

> Is there a way for us to figure out what the currently latest available rust version is for the OS you are on?

I use NetBSD 😀
But yes, you can find the current available version [here](https://pkgsrc.se/lang/rust) 

> How can we keep this action up-to-date so we don't run into this issue too much?

I'm happy to run test builds, I'm usually running on the latest commit on the main branch but, I'm usually using the latest stable Rust.
So, I'm currenty on 1.67.1 but, the latest version on our repositories is 1.66.1

Generally, NetBSD would be having the latest stable version minus one. Sometimes, it could be two versions behind due to branching. _E.g._ at the end of this month the Q1 stable branch of packages will be released, Rust won't be updated until after the branch.

Hope this helps. Feel free to ask for test builds whenever.

hertg:
@0323pin I just figured that we can have an overview of the latest available rust versions for different OS and distros [on repology](https://repology.org/project/rust/badges).

<details>
<summary>Rust Version Overview</summary>

[Packaging status](https://repology.org/project/rust/versions)
</details>

0323pin:
Yes, of course. That will do, although as you might have noticed, NetBSD is not listed.

Well, it is but for some reason it's listed as **pkgsrc current**

hertg:
@Eskaan Can you give this a testrun?

leftwm-core/src/handlers/screen_create_handler.rs
```diff
@@ -7,21 +7,17 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
     ///
     /// Returns `true` if changes need to be rendered.
     pub fn screen_create_handler(&mut self, screen: Screen) -> bool {
+        tracing::trace!("screen create: {:?}", screen);
```
Eskaan, line 10:
```suggestion
        tracing::trace!("Screen create: {:?}", screen);
```

Eskaan (changes requested):
Along my testing I did not encounter any issues (I did not test as extensively).

Whilst this pr as of itself works, I wanted to leave two suggestions:
1. This pr is breaking for everyone using the current workspace id's (they will change now)
2. There is now a lot of unnecessary checks done that I introduced [(example)](https://github.com/leftwm/leftwm/pull/848/files#diff-a6caf002e502c68d8159046acf3ca4762b238b1b3d3c694a229b914a5ac1eec8L231-R235) where the output gets also matched. These can now be removed as id's are unique.
  I would volunteer to remove them and create a pr on your fork.

hertg:
@Eskaan I fixed the capitalizations across all the logs I found

> I would volunteer to remove them and create a pr on your fork.

Yeah you are very welcome to do that, I think the `output` can also be removed in some other places, like in gutters?

> This pr is breaking [..]

Yes it is, I'll have to also update the CHANGELOG before merging

hertg:
I'm doing this in preparation for #1000 which will likely also be breaking, so those two should be combined within the same release

Eskaan:
https://github.com/hertg/leftwm/pull/3

hertg:
I got the commits from @Eskaan in.
This one is ready for merge IMO.
@VuiMuich any objections?

VuiMuich (approved):
LGTM

I thought I saw a regression with the config option `single_window_border: false` and `Monocle` Layout, but I guess that is also in the main branch.

stevensonmt:
Was this included in a release yet? I'd really like to get the soft reload bug that puts all open windows on a single tag fixed.

Eskaan:
I don‘t think so. To my knowledge we are waiting for the Layout pr (#1000) to not make multiple releases with breaking changes.

stevensonmt:
When I tried the git version a couple days ago I had the same issue. I'll wait for the update since I don't like having to manually track updates. Especially since I rely on leftwm every day.

stevensonmt:
I went back to the regular version, but I'll look when I get back to that machine and see what was installed.

PR #1050, opened 2023-03-23
Refactor file handling

VuiMuich:
# Description

So far this is just a refactor to get started with cleaning up file handling.

I am still considering if pulling all the `path` handling into a similar module in `leftwm-core`, or if this would be kind of redundant, as many `Path` and `PathBuf` references are tied directly to the respective lib anyways.

## Type of change

# Checklist:

### DIFF
--- a/leftwm/src/bin/leftwm-check.rs
+++ b/leftwm/src/bin/leftwm-check.rs
@@ -1,18 +1,12 @@
 use anyhow::{bail, Result};
 use clap::{arg, command};
-use leftwm::{Config, ThemeSetting};
-use ron::{
-    extensions::Extensions,
-    ser::{to_string_pretty, PrettyConfig},
-    Options,
+use leftwm::utils::file_handler::{load_from_file, write_to_file};
+use leftwm::ThemeSetting;
+use std::{
+    env, fs,
+    os::unix::fs::PermissionsExt,
+    path::{Path, PathBuf},
 };
-use std::env;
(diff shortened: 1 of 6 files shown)


VuiMuich:
Getting the return type of `fn read_ron_config` is bugging me out.
That part of Rust has always been a weak point for me and now my Rust is even getting a bit rusted (pun almost intended).
But I am positive I can figure this out eventually...

hertg (approved):
Didn't test, just glanced over the changes and they look good to me

VuiMuich:
Just noticed I broke `leftwm check -m`, but I am not sure, if we really can deprecate the migration already...

VuiMuich:
Going back to draft, as I noticed now that I fixed the migration, fallback to toml works no more.

I mean, we could go this route and don't load `toml` any more but still keep the migration routine around for anyone migrating from an old version...
This would simplify a lot.

hertg:
> Going back to draft, as I noticed now that I fixed the migration, fallback to toml works no more.
> 
> I mean, we could go this route and don't load `toml` any more but still keep the migration routine around for anyone migrating from an old version... This would simplify a lot.

Depends on how long ago the switch to ron was released? If it's fair to assume that most people have migrated, I think going that route is fine as long as we communicate it clearly in the release notes and our announcement channel. We don't have to keep technical debt around for too long, given that we are still in a kind of "beta" state with the project anyway.

PR #1052, opened 2023-03-31
on entering fullscreen, save order. On exit, restore.

### LINKED ISSUE #927: Exiting ToggleFullScreen sometimes does not return to previous layout
### Describe the bug

When running ToggleFullScreen twice (e.g., via a keyboard shortcut) if the current window was not previously the "Main" window, then when exiting full screen mode, the windows layout is switched so that the current window now occupies the "main" window slot.

The behavior of other windows managers that provide a full screen toggle is that pressing the toggle twice should result in exactly the same layout.

1. Select a layout with at least two windows.
2. Focus the window that is not the "main" window of the layout
3. Run ToggleFullScreen twice
4. Note that the current window has moved to the "main" window slot

_No response_

### OS / Distro

Linux 5.15.0-52-generic #58-Ubuntu SMP Thu Oct 13 08:03:55 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

```text
$ leftwm --version
leftwm 0.4.0, Git-Hash: 1cc3a4d
```

### LeftWM Check

```text
:: LeftWM version: 0.4.0
:: LeftWM git hash: 1cc3a4d
    -> Configuration loaded OK 
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK 
:: Checking theme . . .
```

jean-santos:
# Description

Hello! First of all, leftwm is awesome! 
I've stumbled on this issue, investigate for a bit and made a fix.
This is my first PR on leftwm, so I'm not sure if I should've done differently.

Fixes #927

## Type of change

# Checklist:


leftwm-core/src/handlers/command_handler.rs
```diff
@@ -143,6 +143,38 @@ fn toggle_state(state: &mut State, window_state: WindowState) -> Option<bool> {
...
+    let for_active_workspace = |x: &Window| -> bool { x.tag == Some(tag_id) && x.is_managed() };
+
+    if window_state == WindowState::Fullscreen {
+        //Going fullscreen, so we save the window order
+        //or else, we restore it!
+        if toggle_to {
+            let mut to_reorder = helpers::vec_extract(&mut state.windows, for_active_workspace);
+            let handles = to_reorder
+                .iter()
+                .map(|each_handle| each_handle.handle)
```
blackdragon2447, line 157: Would it be better to call the input to the closure wither `w` or `window` since `to_reorder` contains windows and not handles when `map` is called?

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -143,6 +143,38 @@ fn toggle_state(state: &mut State, window_state: WindowState) -> Option<bool> {
...
+                .collect::<Vec<_>>();
+
+            state.window_order = handles;
+            state.windows.append(&mut to_reorder);
+        } else {
+            let to_reorder = &mut state.windows;
+            let mut window_restored = vec![];
```
blackdragon2447, line 164: Would it be better to make this plural (`windows_restores`) nice it is a `Vec`?

blackdragon2447 (commented):
I don't want to be too nitpicky, but that should make the code a bit more readable.

leftwm-core/src/handlers/command_handler.rs
VuiMuich: The `focus_manager` uses a field `pub window_history: VecDeque<MaybeWindowHandle>`, of course in this case "history" is not precisely accurate.

(On a side note, there is a focusing issue that is kind of similar to this one: when a pop-up floating window is opened end "destroyed" e.g. some password dialog, the focus returns to the `main window` an not to the last focused one. If you were interested to extend the PR with this, that would be welcome for sure :wink:)

leftwm-core/src/handlers/command_handler.rs
VuiMuich: I assume storing only the handlers was intended to be the cheaper operation, no?

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -143,6 +143,38 @@ fn toggle_state(state: &mut State, window_state: WindowState) -> Option<bool> {
...
+            for window_handle in &state.window_order {
+                let pos = to_reorder
+                    .iter_mut()
+                    .position(|c| c.handle == *window_handle);
+                if let Some(pos) = pos {
+                    let c = to_reorder.remove(pos);
+                    window_restored.push(c);
```
VuiMuich: May I ask, why you chose `c` here? Maybe `i` like `index` might be more obvious?

jean-santos: > I assume storing only the handlers was intended to be the cheaper operation, no?

Yes, I was going for the handlers because was the closest to a 'window id' I could find. You're saying that I could use `focus_manager` to get those handlers? 

I'm up to check the issue you related as a side note. I checked it a bit but so far I couldn't reproduce.
If you have a way to reproduce it, let me know!

Also I'm changing this PR to a draft because I just found out a bug while having 2 full screen window on different workspace. It seems that we should have a 'backup' for each workspace for the functionality to work correctly

VuiMuich: > Yes, I was going for the handlers because was the closest to a 'window id' I could find. You're saying that I could use `focus_manager` to get those handlers?

No, I meant, maybe to change `window_restored ` to `window_order_history` in a similar way as the focus manager does. Also using `VecDeque` had some benefits but I don't recall the specifics at the moment.

> I'm up to check the issue you related as a side note. I checked it a bit but so far I couldn't reproduce. If you have a way to reproduce it, let me know!

The focus restore issue occurs at most to me, when the `pinentry` window (having a window rule to make this window floating) opens to enter the passphrase for my signing key when I `git commit`.
 
> Also I'm changing this PR to a draft because I just found out a bug while having 2 full screen window on different workspace. It seems that we should have a 'backup' for each workspace for the functionality to work correctly

I guess storing the window order history would make most sense to keep for each tag, no?

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -143,6 +143,38 @@ fn toggle_state(state: &mut State, window_state: WindowState) -> Option<bool> {
...
+        //or else, we restore it!
+        if toggle_to {
+            let mut to_reorder = helpers::vec_extract(&mut state.windows, for_active_workspace);
+            let handles = to_reorder
+                .iter()
+                .map(|w| w.handle)
+                .collect::<Vec<_>>();
+
+            state.window_order = handles;
+            state.windows.append(&mut to_reorder);
```
Eskaan, line 161:
```suggestion
            state.window_order = state
                .windows
                .iter()
                .filter(|window| window.tag == Some(tag_id) && window.is_managed())
                .map(|w| w.handle)
                .collect::<Vec<_>>();
```
Just a suggested change that does way less moves out and into the Vec. Also includes the variable set above. 

This is stuff that happens when I am bored in the evening.

Eskaan (changes requested):
I got around to test your implementation and found a medium issue with it: (note: I just saw you got this issue too already, just noting here)

It only saves the position of the last window put in full screen, if you put a window in full screen, go to another tag and put another one in full screen mode, the first one won't be remembered. I guess this is the same behavior you experienced with multiple workspaces.

I suggest you might consider a solution mapping a window going into fullscreen to an index inside the windows array or similar (I did not find out how exactly this is working yet). 

Otherwise I experienced your solution as surprisingly stable. It has no problem with a window switched to floating mode whilst full-screen nor being moved to another tag.

jean-santos:
Now the window order is saved based on the tag. Also I've included some tests.

VuiMuich:
Can confirm, that the main goal is achieved now.
Though I noticed some regression, that the state file cannot be written anymore, I have not yet looked into what exactly is causing the issue though. Most probable its the additions to `leftwm-core/src/state.rs`.

On the side quest: it still occurs, when using `focus_mode: Clickto //or Driven`.

Eskaan:
I've made some changes on this pr I would propose to merge: [Eskaan#fix-fullscreen-add-performance](https://github.com/Eskaan/leftwm/tree/fix-fullscreen-add-performance)
1. First commit (misc fixes and formatting)
   - Fixed saving the state my moving to ron (and removing the `#[serde(flatten)]´)
   - Changed HashMap behavior from find-and-append to just replace. Even if it was realistically possible to have multiple windows in full-screen on the same tag, this way it wouldn't work. 
   - Some `iter_mut()` that only needed to be an `iter()`
   - Other simplifications like removing unnecessary variables that just cause moves and copies.
2. Second commit (not strictly necessary, but I think this sacrifices an always occupied spot in the heap against multiple memory moves on the heap).
   - Move from `VecDeque` to `Vec`

VuiMuich:
Sorry for the push and revert, was testing eskaans branch and messed up the merge AND pushed the mess :facepalm:

VuiMuich (approved):
I think this is good now, right?

Eskaan (approved):
Tested again, still LGTM

mautamu (approved):
This is great! Thank you for taking this on :). 

One (related) question: do we want this kind of behavior to also apply in the case where a Window requests full screen also? Currently if you have two windows, e.g. Firefox and Alacritty, and Firefox requests fullscreen, the Firefox window will pop to the top. (Doesn't need addressed in this PR, just not sure if we want to track that somewhere too).

VuiMuich:
Thanks @jean-santos for this and sorry it took us so long to merge.

PR #1055, opened 2023-04-16
fix - when window gets detroyed, find the last focused

jean-santos:
# Description

This is a bug mentioned by @VuiMuich on the #1052 which happens when a dialog (i.e. asking for a password) gets destroyed and the focus goes to Main and not to the previous focused window. 
With this PR, the focus goes to the previous focused window (if there was one) or else, it use the default behavior, which is looking for a next/previous window on the same workspace.

## Type of change

# Checklist:

### DIFF
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -69,8 +69,14 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
     /// Process a collection of events, and apply them changes to a manager.
     /// Returns true if changes need to be rendered.
     pub fn window_destroyed_handler(&mut self, handle: &WindowHandle) -> bool {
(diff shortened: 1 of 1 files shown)


jean-santos:
It's a draft because I'm not sure if there is a problem taking a window handle from the focus history directly or if I should remove it from the history beforehand

grndplane:
Thank you, this has needed attention for a while. :-)

VuiMuich:
Well the whole of the `window_destroyed_handler` is kind of a mess, that should get some refactor eventually, I guess.
So fer it works for me and I am not having any side effects so far.

Eskaan (approved):
LGTM. Altough I don't know that much about our focus behavior code, it seems to me that it works when in Sloppy mode.

VuiMuich:
Thanks for fixing this as well @jean-santos!

PR #1058, opened 2023-04-23
Update workspaces on screen change

Eskaan:
# Description

Update workspace configuration on `XRRCrtcChangeNotifyEvent`s from Xrandr (When anything of the monitor configuration gets changed).

# State of draft
- [x] Read `XRRCrtcChangeNotifyEvent`s from xlib Events
- [ ] Update workspace(s) corresponding to the Crtc

## Details about `XRRCrtcChangeNotifyEvent`
From my research, a **crtc** is a virtual output that gets rendered by X11, whilst an **output** is a reference to the actual monitor (or similar, I am not really sure about that). The **screen** is the entire rendered area, i.e. everything covered by the DISPLAY env variable.

An `XRRCrtcChangeNotifyEvent` gets emitted, when one of said crtc's is changed. So for example, when a output is disabled, is moved, or the resolution changes. It also gets emitted for crtc that automatically move in place for the disabled crtc/output (beacause one crtc must always be at `x:0, y:0`). 

## Type of change

# Checklist:

### DIFF
--- a/display-servers/xlib-display-server/src/event_translate.rs
+++ b/display-servers/xlib-display-server/src/event_translate.rs
@@ -4,7 +4,7 @@ use super::{
 };
 use leftwm_core::models::{Mode, WindowChange, WindowType, XyhwChange};
 use std::os::raw::c_ulong;
-use x11_dl::xlib;
+use x11_dl::{xlib, xrandr};
 
 pub struct XEvent<'a>(pub &'a mut XWrap, pub xlib::XEvent);
 
@@ -37,7 +37,23 @@ impl<'a> From<XEvent<'a>> for Option<DisplayEvent> {
             xlib::ButtonPress => Some(from_button_press(raw_event)),
             // Mouse button released.
             xlib::ButtonRelease if !normal_mode => Some(from_button_release(x_event)),
-            _other => None,
+            other => {
+                // Also match Xrandr events
+                if let Some(base) = x_event.0.xrandr_event_base {
+                    match other - base {
+                        xrandr::RRNotify => {
+                            // Xrandr XRRNotifyEvent has multiple subtypes, match them too.
+                            match xrandr::XRRNotifyEvent::from(raw_event).subtype {
+                                xrandr::RRNotify_CrtcChange => from_xrandr_crtc_change(x_event),
+                                _other => None,
+                            }
+                        }
+                        _other => None,
+                    }
+                } else {
+                    None
+                }
+            }
         }
     }
 }
@@ -215,4 +231,10 @@ fn from_button_release(x_event: XEvent) -> DisplayEvent {
     let xw = x_event.0;
     xw.set_mode(Mode::Normal);
     DisplayEvent::ChangeToNormalMode
+}
+
+fn from_xrandr_crtc_change(x_event: XEvent) -> Option<DisplayEvent> {
+    let event = x11_dl::xrandr::XRRCrtcChangeNotifyEvent::from(x_event.1);
+    tracing::warn!("Crtc change for {} - {:?}", event.crtc, event);
+    None
 }
--- a/display-servers/xlib-display-server/src/xwrap.rs
+++ b/display-servers/xlib-display-server/src/xwrap.rs
@@ -20,7 +20,7 @@ use tokio::sync::{oneshot, Notify};
 use tokio::time::Duration;
 
 use x11_dl::xlib;
-use x11_dl::xrandr::Xrandr;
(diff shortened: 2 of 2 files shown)


Eskaan:
> Is this intended to work without `SoftReload`

Yes.

> Then we would just need a smart way to get the additional bars started/killed.

Something I haven't thought about yet, but yes, this is important.

For some time now, I thought about maybe doing some hooks for after-initial-startup and similar for quite some time now.
I always have to run my xrandr script manually because I need to SoftReload after it and putting that in `up` causes a boot loop whilst in `.xinitrc` it dosen't work (I think X11 isn't fully initialized yet at this point).

VuiMuich:
I just had a bunch of crazy ideas, but I haven't had any looks at the code at all, so no idea, if those are even remotely feasible:
- Make the `SoftReload` able to take a bool `no-up-script` and/or `Option<String>` with a path to a script to run instead
- this might be easier to implement, if `leftwm-command` would have actual subcommands instead of taking a single String argument. But ofc this would be quite a big refactor, albeit a quite nice one as well. Real bummer with this one: we would break theme compatibility again.

display-servers/xlib-display-server/src/event_translate.rs
```diff
@@ -216,3 +233,68 @@ fn from_button_release(x_event: XEvent) -> DisplayEvent {
...
+    unsafe {
+        let screen_resources = (xrandr.XRRGetScreenResources)(x_event.0.display, x_event.0.root);
+        let crtc_info = (xrandr.XRRGetCrtcInfo)(x_event.0.display, screen_resources, event.crtc);
+        let outputs = slice::from_raw_parts((*crtc_info).outputs, (*crtc_info).noutput as usize);
+
+        if outputs.len() > 1 {
+            tracing::error!(
+                r#"Leftwm does not support more than one output per crtc (if that is even possible to have). 
+                LeftWM will only apply changes to the first output. 
+                If you are seing this error, please create an issue on our GitHub page and it will be resolved."#
```
Eskaan, line 258: @VuiMuich @hertg what do you think about that?

I have tried some things, but I haven't goten xrandr yet to put both of my monitors (outputs) on one crtc. It is possible to return a `Vec<Screen>` here, but I don't think this kind of exotic configuration even exists (if possible). I've added this error to indicate an issue. This is not fatal at all, you will just have to SoftReload like before to reset the configuration.

hertg: I have to admit that I'm a bit too out of the loop to actually understand what you mean by outputs and crtc. But I like your thinking ahead and very verbose log message. I fully approve of, and appreciate your approach.

Eskaan: image
...I am still trying to understand that myself. I haven't found no real documentation on the internal XRandR API whatsoever. I just try to get together what I can.

I think there are the **output**, the **crtc** and the **screen**. The screen is everything covered by the `DISPLAY` env variable, whilst the output corresponds to a hardware monitor and the crtc is basically the virtual representation of the output.

Eskaan:
I have this feature at a working state now, but of course there is still some stuff to do, as most themes will not work with this. For now this is *working:tm:*, you can check it out if you want and make a few tests yourself.

It's now time to discuss **how to implement the theme update hooks** (breaking or not!).

I'll be waiting 'til after #1000 anyways to not make merge conflicts for that beast.

Btw, rerun the test ci please, that didn't fail because of me.

hertg:
triggered a re-run, it went through now

Eskaan:
I just updated this branch in case anyone wants to test it (I didn't yet and don't exactly expect this to behave after a 3.5k line diff merge).

It's still not finished but should work for now by calling the `up` script on screen change, which should in term reload the theme.

Eskaan:
Update: After short testing I noticed two issues:
- We aren't deduplicating screen updates, so the `up` scripts are called multiple times in a short time (asynchronously!), I don't think that's good for most scripts and, for example, is (presumably) resulting in a core dump from picom for me (the process gets killed before it can boot correctly).
- Some focus issues (?) after creating a new workspace. It's just one tag on two workspaces, but it's fixing itself after calling a different tag to one of them. I'm unsure if this problem is related to the pr.

AnthonyMichaelTDM:
For handling reload hooks, why not just add a new (optional) `reload` script in addition to the current `up` and `down` scripts, if the script is present it will use it to update/reload workspaces or whatever and if it's not present we either disable the feature (do nothing) or default to using the `up` script

PR #1062, opened 2023-05-02
Remove field codes from Exec string when autostarting apps

### LINKED ISSUE #1061: Field codes from Exec string are not removed when autostarting an app
### Describe the bug

Field codes from Exec string are not removed when autostarting an app, which leads to field code passed to the application being started (which is not desirable as applications may try to do something with the argument, for example to open it as a link in case of a browser).
before

_No response_

Add a desktop entry to autostart folder with Exec string that contains '%u' field code or similar, for example `Exec=/usr/lib/librewolf/librewolf %u`.

_No response_

### OS / Distro

EndeavourOS

_No response_

### LeftWM Check

```text
:: LeftWM version: 0.4.1
:: LeftWM git hash: 3723a8e
    -> journald-log OK
    -> lefthk OK
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
```

realGWM:
# Description

Remove field codes from Exec string when autostarting apps.

Fixes #1061

Before:
before
After:
after

## Type of change

# Checklist:

### DIFF
--- a/leftwm-core/src/utils/child_process.rs
+++ b/leftwm-core/src/utils/child_process.rs
@@ -176,9 +176,12 @@ fn boot_desktop_file(path: &Path) -> std::result::Result<Child, EntryBootError> 
         return Err(EntryBootError::Hidden);
     }
 
-    if entry.exec.is_none() {
+    let Some(exec) = entry.exec else {
         return Err(EntryBootError::NoExec);
-    }
+    };
+
+    let exec = remove_field_codes(&exec);
+
     let wd = entry
         .path
         .unwrap_or_else(|| dirs_next::home_dir().unwrap_or_else(|| PathBuf::from(".")));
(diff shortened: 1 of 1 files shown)


realGWM:
My solution is not perfect for 2 reasons:

* It hardcodes the field codes (which can actually be any alphabetic character)
* It ignores escaping (e.g. `%%u` should be replaced to `%u`, but currently is replaced with `%`.

I don't see a good way to solve it without adding regex as a dependency (can I do that?), so I'm open to suggestions

VuiMuich:
I believe we already de Pu end on regex.
The window rules and maybe also named workspaces use it iirc.

realGWM:
Actually after thinking more about it, I don't know how to write a proper regex for it either.. :/

If we do something like
* replace '%%' to '%' in the exec string (to "un-escape" percentage signs)
* remove '%[a-zA-Z]' from the string
then "%%u" will be changed to "", while the correct behavior is "%u".
doing it in reverse will result in "%" which is also incorrect.

I've prepared a somewhat ugly looking solution, but it seems to be working as intended (there are some tests also), with some quirks:
* [low-impact] when encountering percentage sign at the end of the string or followed by non-alphabetic non-% character it leaves it as is, although reading the spec it is likely illegal (the other option is to return some Err in this case, but I'm not sure what is better here).
* [a bit more important] the spec also says "Implementations must undo quoting before expanding field codes and before passing the argument to the executable program.", but I think we do no such thing currently. This means that we anyway have some deviations from the spec, so probably the previous point doesn't matter much, still might be a good task for a follow-up.

Let me know what you think about this (commit coming soon)

VuiMuich:
> Actually after thinking more about it, I don't know how to write a proper regex for it either.. :/
> 
> If we do something like
> 
> * replace '%%' to '%' in the exec string (to "un-escape" percentage signs)
> * remove '%[a-zA-Z]' from the string
> 
> then "%%u" will be changed to "", while the correct behavior is "%u". doing it in reverse will result in "%" which is also incorrect.
> 
> I've prepared a somewhat ugly looking solution, but it seems to be working as intended (there are some tests also), with some quirks:
> 
> * [low-impact] when encountering percentage sign at the end of the string or followed by non-alphabetic non-% character it leaves it as is, although reading the spec it is likely illegal (the other option is to return some Err in this case, but I'm not sure what is better here).
> * [a bit more important] the spec also says "Implementations must undo quoting before expanding field codes and before passing the argument to the executable program.", but I think we do no such thing currently. This means that we anyway have some deviations from the spec, so probably the previous point doesn't matter much, still might be a good task for a follow-up.
> 
> Let me know what you think about this (commit coming soon)

Maybe just add this explanation as doc comment before the function, as `leftwm-core` technically is a library this seems well fitting.

realGWM:
> Maybe just add this explanation as doc comment before the function, as `leftwm-core` technically is a library this seems well fitting.

Done, I've added doc comment + updated comments.
Also small changes to the code - it now uses shadowing not to accidentally use old exec string.

realGWM:
Also squashed the changes into one commit.
From my side it is ready to merge, let me know if you think something needs to be changes

leftwm-core/src/utils/child_process.rs
```diff
@@ -398,4 +440,37 @@ mod tests {
...
+            "/path/to/app %%"
+        );
+
+        let sample_exec_with_field_code_and_escaped_percentage_signs = "/path/to/app %%%%%u";
+        assert_eq!(
+            remove_field_codes(sample_exec_with_field_code_and_escaped_percentage_signs),
+            "/path/to/app %%"
+        );
+
+        let bad_exec1 = "/path/to/app %^";
```
VuiMuich, line 470: Nice idea to spin the test like this 👍

VuiMuich (approved):

VuiMuich:
Lgtm, thanks for this little nicety!

Btw: we usually do a squash merge and edit the commit message to only keep relevant commits in (omitting things like "clippy", "format", "revert" or "merge main" etc.) the notes. So you wouldn't need to do this in a future contribution.

PR #1067, opened 2023-05-23
Improve performance of `set_layout`

vrmiguel:
# Description

(Yet more performance PRs :pleading_face:  please bear with me)

Attempts to improve performance of `set_layout`  by not allocating stuff on the heap, hopefully reducing latency

## Changes:

```rust
           let tags_windows = state
                    .windows
                    .iter()
                    .filter(|w| w.has_tag(&tag_id) && w.is_managed() && !w.floating())
                    .collect::<Vec<&Window>>();
```

Here we were allocating a Vec of Window references when we only need the first two results of this filter, which we can do from Iterators alone

```rust
if let Some(w) = tags_windows.get(1).copied() {
      to_focus = Some(w.clone());
}
```

`to_focus` now gets set to `tdw` directly since `tdw` was the second element of the list anyways. `to_focus` is now just a reference to `Window`, since we'll just need its handle at the end of the function

I also moved up the `focused_window.is_some()` check since we don't have to work to do if it's None

## Type of change

# Checklist:

### DIFF
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -431,35 +431,35 @@ fn set_layout(layout: Layout, state: &mut State) -> Option<bool> {
             None => false,
         };
         if !is_focused_floating {
-            let mut to_focus: Option<Window> = None;
+            let mut to_focus = None;
 
             if layout == Layout::Monocle {
                 to_focus = state
                     .windows
                     .iter()
-                    .find(|w| w.has_tag(&tag_id) && w.is_managed() && !w.floating())
-                    .cloned();
+                    .find(|w| w.has_tag(&tag_id) && w.is_managed() && !w.floating());
             } else if layout == Layout::MainAndDeck {
-                let tags_windows = state
-                    .windows
-                    .iter()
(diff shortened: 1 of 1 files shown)


VuiMuich:
Thanks for looking into performance! It is always great to see such contributions :heart:

While checking this out I alse merged #1000 to see if there would be merge conflicts and basically it boils down to the changed import in `leftwm-core/src/handlers/command_handler.rs:447:33`, so @hertg with your consent I will merge this and leaf this easy conflict resolve to your PR.

hertg:
@VuiMuich Sounds good :)

vrmiguel:
> Thanks for looking into performance! It is always great to see such contributions heart

Thanks a lot! I'm afraid of bothering contributors with PRs like these since they might seem like just a chore

If you guys would like more I'm glad to keep contributing

VuiMuich:
> > Thanks for looking into performance! It is always great to see such contributions heart
> 
> Thanks a lot! I'm afraid of bothering contributors with PRs like these since they might seem like just a chore
> 
> If you guys would like more I'm glad to keep contributing

Some one taking care of the chore is actually very welcome and not bothering at all!
If you are up for more just keep things coming.

I will be possibly a little less active the next two weeks, but every now and then I might still have a peek at whats going on here.

PR #1072, opened 2023-05-30
Make configuration errors more visible

realGWM:
# Description

I had a bad configuration (which I noticed because default config was used instead), so I ran `leftwm-check` to check the problem, and at first I thought there was a bug and it didn't catch the error, but it turned out that I just missed the error because it doesn't bring attention to itself (compared to theme errors).

configuration_error

WDYT about making it more visible just like some other errors?

configuration_error_after

## Type of change

# Checklist:


VuiMuich:
Thanks this should indeed be helpful.
And thanks for taking the time to compose such a detailed PR description, especially for a single line commit!

PR #1075, opened 2023-06-06
bump deps and fix build error

VuiMuich:
# Description

Ran `cargo update` and ran into an issue that `git_version` failed on our way we implemented the fallback.
Fixed this.
Built against rust 1.70.0 (`make test-full`) and 1.65.0 (`cargo b` and `cargo r --bin leftwm-check`).

Edit: also fixes a `yanked` warning for `crossterm-channel` that is thrown by `cargo audit`

## Type of change

# Checklist:

### DIFF
--- a/leftwm/src/bin/leftwm-check.rs
+++ b/leftwm/src/bin/leftwm-check.rs
@@ -36,7 +36,7 @@ async fn main() -> Result<()> {
     );
     println!(
         "\x1b[0;94m::\x1b[0m LeftWM git hash: {}",
-        git_version::git_version!(fallback = option_env!("GIT_HASH").unwrap_or("NONE"))
+        git_version::git_version!(fallback = "unknown")
     );
     if matches.get_flag("migrate") {
         println!("\x1b[0;94m::\x1b[0m Migrating configuration . . .");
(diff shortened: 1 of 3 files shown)


leftwm/src/bin/leftwm-check.rs
```diff
@@ -36,7 +36,7 @@ async fn main() -> Result<()> {
     );
     println!(
         "\x1b[0;94m::\x1b[0m LeftWM git hash: {}",
-        git_version::git_version!(fallback = option_env!("GIT_HASH").unwrap_or("NONE"))
+        git_version::git_version!(fallback = "unknown")
```
mautamu, line 39: Out of curiosity, did this get deprecated?

VuiMuich: Yeah, I believe the upstream repo fixed the depenency to the now deprecated `proc-macro-hack`.

Do you think we should add some tests to verify we also can get the `unknow` output releably? (I guess we should do this in a separate PR then, though.)

VuiMuich: After thinking of this, I don't have an idea how to spoof the `git hash` or the absence of one nicely for a test. So I guess we just keep it as is for now.

PR #1077, opened 2023-06-08
feat(leftwm-check): Add a check for other binaries to leftwm-check

mautamu:
# Description

This commit adds an additional check to `leftwm-check`: `check_binaries`. This is similar in nature to `leftwm-check::check_feature_dependencies`, but also checks for `leftwm`, `leftwm-state`, `leftwm-command`, `leftwm-worker`, and `leftwm-check` to ensure that all leftwm-required binaries exist in the user's current PATH variable.

## Type of change

To update [Troubleshooting](https://github.com/leftwm/leftwm/wiki/Troubleshooting):
## Configuration
| Issue | Description | Solution |
|-|-|:-:|
| No config.ron file exists | LeftWM does not always ship with a `config.ron`. You will need to execute LeftWM at least once for one to be generated. | Try the following: ``` leftwm-worker ``` |
| Config.ron is not being parsed | LeftWM ships with a binary called leftwm-check. It might not be installed by the AUR. | Try the following: ``` leftwm-check ``` |
| Keybinding doesn't work/foreign keyboards | You may need to set the modifier (`modifier = []`) or to specify the name of the key, e.g. `eacute` for é. If it is a keypad number, use a modkey with value `lock`.| Refer to [this](https://docs.rs/x11/latest/x11/keysym/index.html) list of keys |
| Keybinding doesn't work | It's likely you need to specify a value, a modifier, or have a typo. | Refer to the definition above or open an issue |
| Not all required binaries present | It's likely you don't have all required binaries in your PATH variable. | Try the following: Install leftwm using `cargo install --path {path}` where {path} is a directory in your PATH variable; set the PATH variable, usually in .bashrc, .profile, or similar depending on your shell |

# Checklist:

### DIFF
--- a/leftwm/src/bin/leftwm-check.rs
+++ b/leftwm/src/bin/leftwm-check.rs
@@ -51,7 +51,14 @@ async fn main() -> Result<()> {
         return Ok(());
     }
 
-    match check_enabled_features() {
+    match check_enabled_features(verbose) {
+        Ok(_) => {}
+        Err(err) => {
+            println!("\x1b[1;91mERROR:\x1b[0m\x1b[1m {err} \x1b[0m");
+        }
+    }
+
+    match check_binaries(verbose) {
         Ok(_) => {}
         Err(err) => {
             println!("\x1b[1;91mERROR:\x1b[0m\x1b[1m {err} \x1b[0m");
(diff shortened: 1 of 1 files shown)


VuiMuich:
Nice work!
:heart: the coc comments to make clippy happy about the error cases.

---
This PR and the fact that the merges from earlier today lead to an incompatibility in state files I was wondering if we would like to have a check, that warns about that fact. I don't think it should report an `OK message` but just a warning if the current and the `future state file` would be incompatible.
This might be potentially quite a bunch of code for a niche feature, so its debatable if this would be even feasible to pursue..

PR #1080, opened 2023-06-08
fix `template.liquid` for `basic_lemonbar` example theme

VuiMuich:
# Description

as raised on discord, the template was outdated.

## Type of change

The `liquid templating` section in the wiki should be checked, if it mentions the deprected `{{ localtime }}`

# Checklist:


VuiMuich:
Checked the wiki, doesn't look like we need to update it for this. If I missed something, let me know..

Issue #1083, opened 2023-06-09
Ability to jump to next/previous non-empty workspace


simtd:
### Describe the feature

As far as I see there are no issues open describing this functionality.

This is a feature that bspwm has and also patches exist for dwm to add this functionality also. Basically it's the ability to press a hotkey to cycle through workspaces that has windows/clients on them. Both back and fourth. For my personal use case I have bound "super + ." to jump to the next non-empty workspace and "super + ," to go the other way in the window manager I use currently.

### How important is this feature to you?

My workflow is impaired without it

### How important do you think is this feature for the average LeftWM user?

Would probably be nice to have

VuiMuich:
For reference on implementation, a similar optional `bool` value can be found in the `MoveWindowToPreviousTag` and `MoveWindowToNextTag` commands.

Eskaan:
Above pr got merged and released in  [v0.4.2](https://github.com/leftwm/leftwm/releases/tag/0.4.2). Thanks for your contribution!

PR #1086, opened 2023-06-10
Add Optional Behavior Settings for FocusNextTag and FocusPreviousTag

### LINKED ISSUE #782: [Feature request] go to next unused tag
Let's suppose I have a terminal in tag 1 and a browser in tag 2. I would like to assign a keybinding to go straight to tag 3 (which would be the next unused tag... that is, no apps already open on that tag)

Silicasandwhich:
# Description

Add an optional parameter to the FocusNextTag and FocusPreviousTag keybind commands that lets the user ignore used tags or ignore empty tags.

Fixes #782 #1083 

## Type of change

* */wiki/External-Commands*

| Command         | Arguments (if needed)     | Notes |
|--------------|-----------|------------|
| FocusNextTag | ```behavior``` | string (defaults empty), determines command behavior. No value or ```default``` will simply move focus to the next tag, ```goto_used``` and ```ignore_empty``` will go to the next used tag, and ```goto_empty``` and ```ignore_used``` will go to the next empty tag. |
| FocusPreviousTag | ```behavior``` | string (defaults empty), determines command behavior. No value or ```default``` will simply move focus to the previous tag, ```goto_used``` and ```ignore_empty``` will go to the previous used tag, and ```goto_empty``` and ```ignore_used``` will go to the previous empty tag. |

* */wiki/Config#keybind-commands*

## FocusNextTag

Moves the focus from the current tag to the next tag in a given workspace. An optional parameter can be specified to change the behavior of focusing. No value or ```default``` will simply move focus to the next tag, ```goto_used``` and ```ignore_empty``` will go to the next used tag, and ```goto_empty``` and ```ignore_used``` will go to the next empty tag.

Example:

```rust
         (command: FocusNextTag, modifier: ["modkey"], key: "n"), // current default config
        // (command: FocusNextTag, value: "default", modifier: ["modkey"], key: "n"), // same behaviour as no value
        (command: FocusNextTag, value: "goto_used", modifier: ["modkey", "Shift"], key: "n"), //same as value `ignore_empty`
        (command: FocusNextTag, value: "goto_empty", modifier: ["modkey", "Alt"], key: "n"), //same as value `ignore_used`


```
## FocusPreviousTag

Moves the focus from the current tag to the next tag in a given workspace. An optional behavioral parameter can be specified in the same way as with [FocusNextTag](#FocusNextTag).

Example:

```rust
        (command: FocusPreviousTag, modifier: ["modkey"], key: "m"), // current default config
        // (command: FocusPreviousTag, value: "default", modifier: ["modkey"], key: "m"), // same behaviour as no value
        (command: FocusPreviousTag, value: "ignore_empty", modifier: ["modkey", "Shift"], key: "m"), //same as value `goto_used`
        (command: FocusPreviousTag, value: "ignore_used", modifier: ["modkey", "Alt"], key: "m"), //same as value `goto_empty`
        
        
```

the output of ```leftwm command --list``` has been adjusted to reflect the new functionality

# Checklist:


Silicasandwhich:
one thing to note is that currently when there are no empty tags, the command just doesn't do anything. however, there might be benefit to creating a new tag when the command is called in this situation, similar to how calling ```GoToTag``` on a non-existent index will create a new tag. we could also make it an option, similar to the swap option on ```GoToTag```

leftwm-core/src/command.rs
```diff
@@ -32,6 +32,7 @@ pub enum Command {
         tag: TagId,
         swap: bool,
     },
+    GoToNextEmptyTag,
```
mautamu, line 35: Might want to rename this "FocusNextEmptyTag" to keep consistent with "FocusNextTag". Not sure if we'd also want to add a FocusPreviousEmptyTag as well, but something to consider.

One other idea would be to use FocusNextTag and FocusPreviousTag, but make them have an optional boolean:
```rs
FocusNextTag(Option<bool>),
FocusPreviousTag(Option<bool>)
```

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -1397,4 +1420,87 @@ mod tests {
         assert_eq!(manager.state.windows[0].border(), 1);
         assert_eq!(manager.state.windows[1].border(), 1);
     }
+
+    #[test]
```
mautamu, line 1424: Thank you so much for contributing good unit tests!

mautamu (commented):
Hello, fellow adventurer!

Thank you so much for making this pull request! Just had a few stylistics remarks, but overall a solid commit and I'm glad LeftWM will be getting this feature soon!

leftwm-core/src/command.rs
Silicasandwhich: Name change seems like the best option to me, ill work on doing that when im available. My thing with making it an optional parameter is that most of the other optional parameters for commands are for changing the behavior of the command (e.g. the swap parameter for GotoTag), while going to the next empty tag feels different enough from the FocusNextTab command that an average user probably won't intuit that this functionality is contained in it, potentially making it more difficult for a user glancing at the documentation to learn that this feature is even an option. also, to a lesser extent, if the comment on the original issue about making more advanced tag control options in general is part of the plans for the future, i think it would set a bad precedent if this single control feature is tied to the FocusNextTab command, though it's not like the code couldn't be refractored to accomodate that if that becomes the case. either way, this is just my perspective as an outsider looking in (and i probably spent more words writing about this change than it would take to implement it lol), and i would be more than happy to change the code to reflect this change if it would help make the codebase clearer.

as for FocusPreviousEmptyTag, i feel like it wouldn't have that much of a use, as i feel like the main use case for FocusNextEmptyTag is when you just want a clean workspace and don't particularly care where it is. though, with how trivial it would be to implement, it might be a good idea regardless.

VuiMuich:
> one thing to note is that currently when there are no empty tags, the command just doesn't do anything. however, there might be benefit to creating a new tag when the command is called in this situation, similar to how calling ```GoToTag``` on a non-existent index will create a new tag. we could also make it an option, similar to the swap option on ```GoToTag```

A very warm welcome and thank you from my side as well, fellow adventurer!

Tbh, I would actually prefer the Boolean option, as just [commented here](https://github.com/leftwm/leftwm/issues/1083#issuecomment-1584688565)

This would close #1083 then as well.

Silicasandwhich:
just saw your edit as i was pushing the previous commit, ill definitely look into doing that tomorrow. internally, i think the three options would be represented by a ```FocusTagDeltaBehavior``` enum, with the options ```Default```, ```IgnoreEmpty```, and ```IgnoreUsed```.  im not entirely familiar with the code base, so where would it be most appropriate to put this enum? I noticed that the scratchpad command have a similar enum in the ```scratchpad_handler.rs```, but it also has around 1300 lines of other functionality, which makes it seem like overkill to store a 3-option enum in the same place lol. the other option would be in the ```command.rs``` file under leftwm-core, but seeing as it's literally just the command enum, that doesn't seem right either.

VuiMuich:
Sorry to come with more and more ideas, while you are in the middle of working on this.

I think I like to keep it separate funktions in the background, but given the similarities of this options it might be worth thinking of ways to have least repetitive code.

But before it gets overly complicated I'd say adding the feature(s) is a higher priority to code style which could be refactored in a later PR.

VuiMuich:
Awesome! Will give this a full review and live test asap.

Nice catch on the `Makefile`, btw.

Silicasandwhich:
Sorry for the late response, I'm currently on vacation so I don't have a lot of free time. I've added in another option to ```FocusNextTag``` and ```FocusPreviousTag``` for ignoring empty tags. As for the ```ignore_empty``` vs ```goto_used```, i ended up just... doing both? ```ignore_empty``` and ```goto_used``` will both map to the IgnoreEmpty behavior, while ```ignore_used``` and ```goto_empty``` will map to the IgnoreUsed behavior. Though, if that ends up being too confusing, we could just remove one or the other. Either way, I'm going to update the wiki snippets once I post this comment, and likely change the title of the PR to something more appropriate to the expanded functionality. Obviously I'm going to run this build a little bit more on my own machine to catch any edge cases I didn't account for in the tests, but so far I think it's been pretty stable, and ive tried to provide at least a semblance of a platform to expand this functionality if other PRs and commits want to add additional behaviors to the ```FocusNextTag``` and ```FocusPreviousTag``` commands.

Silicasandwhich:
im also realizing that this may imply changes to [leftwm-config](https://github.com/leftwm/leftwm-config), im not exactly sure where it's sourcing its keybind checking from, but it definitely did not like the new commands in my config file when i was testing the first commit lol, and by extension im guessing it won't flag invalid values for the behavior parameter on the new FocusNextTag and FocusPreviousTag

VuiMuich:
Tagging @blackdragon2447 for the `leftwm-config` question, as you might have the best knowledge of that code. If you don't have time for this rn, I can have a look at it when reviewing.

@Silicasandwhich No worries about late replies, enjoy your vacation and thanks for this great contribution!

VuiMuich:
After a brief look at `leftwm-config`'s code, I'd say that this is all manually coded.

leftwm-core/src/utils/command_pipe.rs
```diff
@@ -441,4 +470,24 @@ mod test {
...
+    }
+
+    #[test]
+    fn build_focus_previous_tag_without_parameter() {
+        assert_eq!(
+            build_focus_previous_tag(""),
+            Command::FocusPreviousTag {
+                behavior: command::FocusDeltaBehavior::Default
+            }
+        )
```
VuiMuich, line 491: To make github clippy happy:
```suggestion
        );
```

leftwm-core/src/utils/command_pipe.rs
```diff
@@ -441,4 +470,24 @@ mod test {
...
     }
+
+    #[test]
+    fn build_focus_next_tag_without_parameter() {
+        assert_eq!(
+            build_focus_next_tag(""),
+            Command::FocusNextTag {
+                behavior: command::FocusDeltaBehavior::Default
+            }
+        )
```
VuiMuich, line 481: To make github clippy happy:
```suggestion
        );
```

VuiMuich (changes requested):

leftwm-core/src/utils/command_pipe.rs
```diff
@@ -301,6 +302,34 @@ fn build_decrease_main_width(raw: &str) -> Result<Command, Box<dyn std::error::E
...
 
+fn build_focus_next_tag(raw: &str) -> Command {
+    match raw {
+        "ignore_empty" | "goto_used" => Command::FocusNextTag {
+            behavior: command::FocusDeltaBehavior::IgnoreEmpty,
+        },
+        "ignore_used" | "goto_empty" => Command::FocusNextTag {
+            behavior: command::FocusDeltaBehavior::IgnoreUsed,
+        },
+        _ => Command::FocusNextTag {
```
VuiMuich, line 313: Not entirely sure, if we shouldn't better match on actual `none | "default"` and error on `_`?
Right now `leftwm-command "FocusNextTag gurke` is behaving like `default`.

leftwm-core/src/utils/command_pipe.rs
```diff
@@ -301,6 +302,34 @@ fn build_decrease_main_width(raw: &str) -> Result<Command, Box<dyn std::error::E
...
+
+fn build_focus_previous_tag(raw: &str) -> Command {
+    match raw {
+        "ignore_empty" | "goto_used" => Command::FocusPreviousTag {
+            behavior: command::FocusDeltaBehavior::IgnoreEmpty,
+        },
+        "ignore_used" | "goto_empty" => Command::FocusPreviousTag {
+            behavior: command::FocusDeltaBehavior::IgnoreUsed,
+        },
+        _ => Command::FocusPreviousTag {
```
VuiMuich, line 327: see above.

VuiMuich (commented):
Besides the things below, an update to the printout of `lefwm-command --list` (leftwm/src/bin/leftwm-command.rs:44 and following) would be needed.

VuiMuich:
I tried to patch things up, to get on with the testing, its not pretty though and the last two tests I didn't get to work so I commented them out for now. If you want to check it out its on my fork: https://github.com/VuiMuich/leftwm/tree/Silicasandwhich/main

leftwm-core/src/utils/command_pipe.rs
VuiMuich: I missed that this is about command pipe, so obviously `none` is in this case `""`..

Silicasandwhich:
alright, yeah, the changes you made look good to me. i can work on backporting them to this PR when i get home.

VuiMuich:
> alright, yeah, the changes you made look good to me. i can work on backporting them to this PR when i get home.

I am not very happy about all that lifetimes annotations needed, but my rust-fu is also to weak to "do better then the compiler suggests" in this area...

Silicasandwhich:
yeah, i feel you... i think that since the error references the string slice containing the value, you kinda need lifetime annotations to ensure the string slice sticks around as long as the error does. the one thing i can think of in this situation is storing a String instead of a &str and calling ```to_owned``` on ```raw``` in the builder functions, which would involve a clone, but like.... it's one clone in an error state, i can't imagine the performance implications are that dire lol. also, it's been a bit since i read the rust book, so i could be completely wrong here.

Silicasandwhich:
ok, so... i don't know if this is just a problem on my end, but it seems like not only has the external command stopped working with the newest few commits, but so has the keybinding. i can't tell if this is just a problem on my end, though, as even when reverting the past few commits i still have this problems

leftwm-core/src/utils/command_pipe.rs
Silicasandwhich: implemented in newer commits

leftwm-core/src/utils/command_pipe.rs
Silicasandwhich: see above

Silicasandwhich:
nevermind, i found the problem, i had installed leftwm from git using cargo in the process of troubleshooting the external commands not working (the problem was that I forgot quotations marks), and my system seemed to be using those binaries instead of my installed ones.

mautamu (approved):
Looks great! Thank you so much for all of your diligence on this @Silicasandwhich and @VuiMuich!

VuiMuich (approved):

VuiMuich:
There is just one thing, that I don't consider a blocker, but wanted to ask for opinions while we keep it in live testing for one more day before we merge:
The terms `used_tags` and `unused_tags` in the `command_handler` functions (I.e. line 161) might be confusing. BUT I have been thinking of introducing/changing tag related terminology in general a bit, which in detail should be discussed in its own enhancement issue.
Just briefly it goes like "windows get (un-)tagged" and "workspaces (un-)subscribe to tags". The hope is to further make the difference to dwm/i3 style workspaces clearer.

Silicasandwhich:
yeah, i agree that used and unused are kind of vague. the other option i was thinking about was occupied/unoccupied, but if the new terminology is tags as a property of windows rather than containers of windows, that might not be ideal.

VuiMuich:
Yeah, I thought of occupied as well and often used it in support requests actually, but somehow it feels not right.. but as said, thats ok to be taken care of at another time.

VuiMuich:
Thanks @Silicasandwhich for this high quality contribution. :star_struck:

If you are hungry for more, you know where you find a bunch of open issues or just ask on the discord :wink:

Silicasandwhich:
yeah, no problem. leftwm's the only window manager that really clicked for me, and im definitely interested in helping to make it better

VuiMuich:
That is awesome to read!
Thanks again 🙏🏻

VuiMuich:
Updated the wiki.

PR #1098, opened 2023-06-21
Improved Errors and Documentation for External Commands

### LINKED ISSUE #1096: Better error handling for external commands
### Describe the enhancement

There are a few things about `leftwm-command` that make it difficult to get a grasp of as a new user, and difficult to work with as an experienced one:

1. when you run a command with `leftwm-command`, errors are not logged to stdout or stderr unless it literally had an error writing to the pipe file. for example, if i run `leftwm-command StartMeowing`, there is no indication that what i just sent was an invalid command. there is no output whatsoever. perhaps more frustrating is the case of malformed command: executing `leftwm-command GoToTag 2` returns zero output and doesn't have any effect on the desktop, which can (and has) led users to believe that the command is broken, the command only works as a keybinding, the options are broken (in the case of something like `FocusNextTag` which has optional parameters), etc.
2. thankfully, errors are actually logged somewhere, but that "somewhere" isn't very consistent. On my laptop, despite building with the makefile and never changing any of the default options, errors are logged to the syslog (/var/log/syslog). by default, it logs to journald, and i think it did when i was trying to run it remotely off of my server? the other option is to a file, and i finally figured out which file it was when i was looking through the "building from source" section of the readme. the issue here is that this behavior is barely documented. the only time that any mention of logging is made is in the building from source section of the readme, and it basically just says that logging happens to one of the aforementioned places. it doesn't mention what gets logged, or what the walls upon walls of `AVOID_STRUT` mean. And furthermore, on the External Commands section of the wiki, there is no mention of the fact that any command errors will only appear in the log file. now, once you actually know about this behavior, it's not too cumbersome to deal with. it's certainly not ideal, but having ```journalctl -f /usr/local/bin/leftwm-worker``` or ```tail -f /var/log/syslog | grep leftwm``` open in another terminal gives you more or less what you need. however...
3. a lot of the errors that leftwm gives aren't very helpful. I mean, invalid command like `leftwm-command StartMeowing` give no error whatsoever, and while nobody's going to think that StartMeowing is actually a valid command, there's going to be people who make a typo and run `GoOtTag`, or think that the commands aren't case sensitive and run `gototag` (honestly they really should be, but that's another thing entirely), or are expecting a keybind command like `MoveToLastWorkspace` to also be an external command, and none of those people are getting any feedback whatsoever, which leads to a lot of confusion. Even when it does actually show an error, it can often be unhelpful. a common pitfall of new users is to think that leftwm-command wraps all of its input into a single command, so they run something like `leftwm-command GoToTag 2` and aren't sure why nothing is happening. If they find the log, the error they see will be something along the lines of `An error occurred while parsing the command: cannot parse integer from empty string`, something that only really makes sense in this context if you have looked at the source code for the command parser and know how it works, and if that was the case, they wouldn't have made this mistake in the first place (well, theoretically, at least. it certainly didn't stop me). 

In conclusion, the way that errors are currently being handled for external commands sets up new users for confusion, and sets up veteran users for inconvenience. In the short term, I think the wiki needs to reflect where error logging happens on the external commands page, and some mention of it should be made in the leftwm-command itself (at minimum in the help page, ideally there should be some sort of mention when a command is sent that you can check for errors at {wherever}). In the long term, error messages *need* to be improved, and in the very long term, errors should really be written to the stderr of whatever called the command. Also, the fact that leftwm-command treats every new argument as a new command, therefore requiring you to put quotes around everything, is quite unintuitive and clunky, but changing that would break the configs of everyone who's already gotten used to it. This is a little pie-in-the-sky, but it would be nice if we could intelligently determine if the user forgot to put quotes around their command and tell them what they need to do to fix it in the error.

Would be nice to have

Silicasandwhich:
# Description

See #1096 for the exigence behind this pull request

TASK LIST:
- [x] leftwm-log command
- [x] improved error messages for leftwm-command
- [x] log when invalid command is attempted
- [x] documentation for where errors are logged
- [x] reference leftwm-log in leftwm-command help and/or after command is sent
- [x] reliably send command errors to stderr instead of exclusively logging

if that last one can be accomplished, then task 5 (or even 4) can probably be ignored.

Fixes #1096 

## Type of change

* *[/wiki/Troubleshooting:](https://github.com/leftwm/leftwm/wiki/Troubleshooting)*

# Troubleshooting
## Table of contents

- [Logging](#logging)
- [Configuration](#configuration)
- [Environment](#environment)

## Logging
LeftWM keeps track of errors and information by logging it, including external command errors, errors when loading themes, and startup information. If something doesn't seem to be working how it should, such as external commands having seemingly no effect, the log is a good place to start looking. Additionally, including relevant logging information in issues and support conversations can help others determine what is wrong. 

By default, leftwm will log to the systemd journal, which can be accessed by `journalctl`. to only get leftwm journal entries, try `journalctl ${which leftwm-worker}` or `journalctl ${which lefthk-worker}`. those should display most if not all errors logged by leftwm. Furthermore, if you specified the `sys-log` or `file-log` build options, you can find a log at `/var/log/syslog` or `/tmp/leftwm/<log-file-by-datetime-of-launch>`.

Additionally, you can run `leftwm-log` to interface with any of these. it uses a regex to grab all journalctl entries for leftwm processes, provides instant access to the session's current log file in /tmp/leftwm/, allows you to follow the file as it grows similar to `tail -f`, and more. see `leftwm-log --help` for more information.

* *[/wiki//External-Commands:](https://github.com/leftwm/leftwm/wiki/External-Commands)*

This is a brief overview of the available external commands and their possible arguments.

You can pass the string of the external command directly to the pipe file. The file will be named "command-{display number}.pipe", where the display number can be found using ```echo $DISPLAY```.
For example from a shell you could use:
```shell
echo "SetLayout CenterMain" > $XDG_RUNTIME_DIR/leftwm/command-0.pipe
```
when on display 0.
If you are on the leftwm 0.2.8 or above, external commands can be passed in using leftwm-command. Commands that contain arguments require quotes.
For example:
```shell
leftwm-command "SetLayout CenterMain"
```
errors will be logged to both stderr and to the log. see `leftwm-log --help` or the [troubleshooting page](https://github.com/leftwm/leftwm/wiki/Troubleshooting#logging) for more information

Commands can also be chained together, for example:
```shell
leftwm-command "SendWindowToTag 2" "SendWorkspaceToTag 0 2"
```
**NOTE:** each argument to leftwm-command istreated as a separate command, so commands and their parameters must be group together, as in the above example.

# Checklist:


Silicasandwhich:
so im at an impasse in terms of how to handle making invalid commands log an error, as the problem appears to be the command handler in leftwm-core passing every command that isn't explicitly defined to a separate command handler defined in the leftwm config. The thing is, defining custom commands doesn't seem to be a documented feature anywhere, and I can't figure out how to tap into it myself, so im wondering if this is an actual thing or not? because if it isn't it would be much easier to just move the few commands that are controlled by config.rs (LoadTheme and UnloadTheme iirc) to the command handler and do error handling there.

Silicasandwhich:
tagging @VuiMuich as i presume you know more about the codebase than i do

Silicasandwhich:
so i made it log errors to stdout, but there are two problems that need to be solved:

1) im occasionally getting "broken pipe" errors from the command pipe when running multiple external commands in short succession. im not sure what's causing this, or if it was already an issue on main and im just now discovering it because sending ten "GoToTag"s in 40 milliseconds isn't common behavior, but it should be investigated
2) i was also getting a lot of  errors that seemed to suggest that the return pipe was getting double dropped somehow, but they seem to have disappeared after merging, so im not sure what's going on there. something to watch out for though.
3) errors and successes are both sent to stdin, so we should change it so that errors are sent to stdout (so basically figuring out how to send a Result<String, String> over the pipe instead of just a string).

these three issues in addition to making invalid commands log as errors should be fixed before this is merged

VuiMuich:
> so im at an impasse in terms of how to handle making invalid commands log an error, as the problem appears to be the command handler in leftwm-core passing every command that isn't explicitly defined to a separate command handler defined in the leftwm config. The thing is, defining custom commands doesn't seem to be a documented feature anywhere, and I can't figure out how to tap into it myself, so im wondering if this is an actual thing or not? because if it isn't it would be much easier to just move the few commands that are controlled by config.rs (LoadTheme and UnloadTheme iirc) to the command handler and do error handling there.

I'll have to dig in a bit myself to recall the specifics, but in a brief answer: iirc this in parts was introduced with the librification of `leftwm` to `leftwm-core`, so removing the feature might break leftwm for users who build their own WM on the basis of `leftwm-core`.
Lack of documentation is really frustrating sometimes, but at the same time it's hard to document code you haven't written yourself...

VuiMuich:
At some point there were plans to replace the command-pipe entirely with dbus, but that contributor had to withdraw their activity for personal reasons.

I also have a vague memory that those plans were the reason, why we left some annoying things untouched, as we thought we would replace it anyways.

Silicasandwhich:
yeah, switching to dbus would certainly make things more robust. my main concern is that it would break compatibility with themes and other scripts that use echoing to the pipe to run commands (if it's any sign, more than a third of the themes in the community themes repo would break). like, it would simultaneously be a change that almost exclusively benefits contributors, yet would also break enough configs of users to necessitate a version bump. but, idk, might still be something to play around with.

Silicasandwhich:
ok, so ive mostly made invalid commands log errors, but there are a few new things that need fixing before merge (besides refactoring) besides what was said in the previous comment:

1. leftwm command still has infinite blocks occasionally and im not sure why, it might just be that i implemented the tokio concurrency thing wrong
2. the way that I made invalid commands log errors is kind of janky, and while it would be best to expose some way for the custom command handler to return errors, but trying to implement that not only requires a metric ton of refractoring, but is also kind of difficult since the command_pipe is what's logging the errors right now, and the config is only referenced in the command handler step (which comes later), so it would require some sort of change to the order that things are processed (though, im now realizing that the command_pipe could just pass its parse errors to the command handler, and the logging could happen there instead).
3. currently, custom commands are only being executed if arguments are passed along side them (which is particularly worrying for UnloadTheme), and a fix for that is pending in anticipation of a fix for point 2, and because of point 2, errors from UnloadTheme and LoadTheme are not being logged to stdout

Silicasandwhich:
assuming this last test  passes, this PR has more or less accomplished its initial goals. There are two main issues that should be solved (as well as a review of the proposed documentation changes, as well as a general review, ive gotten so tunnel-visioned over the course of making this PR that ive doubtlessly missed several glaring issues) before the PR is merged:

1. infinite blocking is still happening inconsistently, i presume it's some sort of race condition since it's so inconsistent, but im not sure where it could be happening. the code should be awaiting exactly one response per command sent before exiting out, so im not sure what's halting it.
2. all diagnostics are still logging to stdout instead of errors going to stderr and oks going to stdout. i have an idea for how to do this (essentially send either "OK:" or "ERROR:" before the diagnostic message and parse that to generate an exit code / decide between println! and eprintln!), but it's currently 3:25 AM where I am, so that's gonna be a job for tomorrow.

regardless, issue number 1 is going to require someone with more experience with tokio and concurrent programming  than i do (so... literally anyone whose done either before), so im going to open it up for review. it would also be helpful in getting another set of eyes to find anything that got messed up during all the reversions and working tree nukes i did trying to figure out how to properly do that second to last command.

Silicasandwhich:
alright, so on the new commit, i can no longer reproduce the infinite blocking. after I update the documentation changes to reflect the new functionality, i will request a review

VuiMuich:
Awesome work 😎 Thank you so much!
I plan to review this tomorrow morning.

Silicasandwhich:
i merged with the main branch, it seems to be playing well with the new layouts stuff and the command changes. one thing that could be improved is a better command for leftwm log with journalctl logging, since right now if you don't use -f, it takes quite a while for grep to search through your entire log file for leftwm and lefthk logs. if there's a better way to do this, like somehow being able to give multiple executables for journalctl to output for, that would be far preferable

Silicasandwhich:
im stupid, just realized you can literally do just that... leftwm-log, lefthk-worker, and leftwm-command seem to be the only ones with tracing utilized, so im just gonna have it pull from those. newest commit should have everything ready

leftwm/src/bin/leftwm-log.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+fn get_command() -> clap::Command {
+    command!("LeftWM Log")
+        .about("retrieves information logged by leftwm-worker")
+        .help_template(leftwm::utils::get_help_template())
+        .args(&[
+            arg!(-J --journald "use journald log (default)"),
+            arg!(-S --syslog "use syslog (default if built with no journald support"),
```
VuiMuich, line 51:
```suggestion
            arg!(-S --syslog "use syslog (default if built with no journald support)"),
```

leftwm/src/bin/leftwm-log.rs
```diff
@@ -0,0 +1,127 @@
...
+            arg!(-F --file "use file (default if built with no syslog support"),
```
VuiMuich, line 52:
```suggestion
            arg!(-F --file "use file (default if built with no syslog support)"),
```

VuiMuich: Should we hint on the path of the log file here?

leftwm/src/bin/leftwm-log.rs
```diff
@@ -0,0 +1,127 @@
...
+            arg!(-i --"ignore-build-opts" "attempt logging regardless of build options"),
```
VuiMuich, line 53: Sorry for my ignorance, but what be the usecase for this flag?
I don't mind adding it, I just don't get it, what situation it would be beneficial in.

VuiMuich (commented):
Looking quite good so far, just a bunch of minor nitpicks in the `leftwm-log` flags.

leftwm/src/bin/leftwm-log.rs
Silicasandwhich: so i found it useful because for whatever reason my journald log was cloning to my syslog, the previous implementation of the systemd log was super slow, and i somehow found it more preferable to program in an entire  flag rather than change the build options? either way, we can probably remove it without very many people being upset

leftwm/src/bin/leftwm-log.rs
Silicasandwhich: added in latest commit

leftwm/src/bin/leftwm-log.rs
Silicasandwhich: removed in latest commit

leftwm/src/bin/leftwm-log.rs
Silicasandwhich: when called with -F, leftwm-log now prints out the path to the log file first. there could be better ways to expose this, so ill leave this unresolved for now.

Eskaan (changes requested):
Great pr you did there! I haven't actively tested this yet, but this is my first look on the code with a few formatting requests.

leftwm/src/bin/leftwm-command.rs
```diff
@@ -17,23 +20,42 @@ async fn main() -> Result<()> {
...
+                        match result{
+                            "OK:" => println!("{command}: {msg}"),
+                            "ERROR:" => {eprintln!("{command}: {msg}");exit_code = 1;},
+                            _ => println!("{command}: {res}"),
+                        }
+                    }else{
+                        println!("{command}: {res}");
+                    }
+            }
+                _ = timeout(5000) => {eprintln!(" WARN: timeout connecting to return pipe. Command may have executed, but errors will not be displayed."); exit_code = 1;},
```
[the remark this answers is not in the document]
Silicasandwhich: fixed in latest commit

leftwm/src/bin/leftwm-command.rs
```diff
@@ -103,3 +125,35 @@ fn print_commandlist() {
...
+
+    #[error("Couldn't connect to file: '{0}'")]
+    ConnectToFile(PathBuf),
+}
+
+async fn get_return_pipe() -> Result<ReturnPipe, Error> {
+    let file_name = ReturnPipe::pipe_name();
+
+    let pipe_file =
+        place_runtime_file(&file_name).map_err(|_| Error::CreateFile(file_name.clone()))?;
```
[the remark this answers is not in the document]
Silicasandwhich: fixed in latest commit

leftwm/src/config.rs
```diff
@@ -447,27 +447,49 @@ impl leftwm_core::Config for Config {
...
-                    return false;
+                    write_to_pipe(&mut return_pipe, "ERROR: Command not recognized");
+                    false
+                }
+            }
+        } else {
+            match command {
+                "LoadTheme" => {
+                    tracing::warn!("missing parameter theme_path");
+                    write_to_pipe(&mut return_pipe, "ERROR: missing parameter theme_path");
```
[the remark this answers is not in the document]
Silicasandwhich: fixed in latest commit

Silicasandwhich:
i don't know why the nix flake failed to build on the latest commit, the logs suggest an internal error on github's part? not sure

VuiMuich:
Thank you @Silicasandwhich, as said befor awesome contribution!! :tada: :rocket:

PR #1101, opened 2023-06-24
Fix and improve basic_lemonbar theme

simtd:
# Description

I fixed and improved the basic_lemonbar theme. The previous one didn't work for me because it was coded to only work for a multiple workspace setup. Now it should work for both one or multiple workspaces. 

New features or fixes:
- now works if user only has one workspace
- clock is independent of leftwm-state
- added clickable tags
- the status text not provided by leftwm-state only shows up on one instance of lemonbar
- added uname info
- changed lemonbar size slightly
- got rid of compositors
- better naming/syntax
- different theming

Fixes #(issue)

## Type of change


simtd:
I am working on adding a README and some other small minor changes. Hold on.

VuiMuich:
Looking very nice, thanks for all the effort.
The up script isn't actually not very basic, but I don't really mind 😃
It will probably take till tomorrow for me to do a final review BBC and merge.

simtd:
It was as basic as I could make it while still showing how to use lemonbar properly. Also the fact that there can be multiple workspaces adds some complexity.

VuiMuich:
This is excelent! Thank you so much! :pray:

PR #1116, opened 2023-08-10
Add back output field to leftwm-state

### LINKED ISSUE #1111: Add (back) output field in leftwm-state
### Describe the feature

Add back the information on which output/monitor a workspace is.
Before [this commit](https://github.com/leftwm/leftwm/commit/7ed4109ad5a21de35bfda04c6f4f791bae15057e) there was the `output` field which contained the physical monitor on which a workspace is display.

I work on a laptop, with the built-in screen and, from time to time, an external 4k Monitor, which I split up into 4x ~1080p workspaces.
As (I assume, many others) I use it in combination with eww. I use the `up` script to generate the bars for each workspace and attach the position and monitor to the bar accordingly. Before the mentioned change it was easily possible for me to achieve this. Now I have to riddle around to find out which workspace is placed on which monitor. It makes the whole workflow needlessly complicated.

I have to check

1. In which order are the monitors aligned?
2. Does one (or multiple) of the active outputs have multiple workspaces?
    - I have to parse this from the config
3. Which workspace index belongs to which output?

I don't know, may be a niche edge-case

Eskaan:
# Description

Fix an regression in #1043 that removed the output field from `leftwm-state`.

Fixes #1111

## Type of change

# Checklist:

  Note: Independent clippy warnings were ignored.

### DIFF
--- a/leftwm-core/src/models/dto.rs
+++ b/leftwm-core/src/models/dto.rs
@@ -4,6 +4,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Viewport {
     pub id: usize,
+    pub output: String,
     pub tag: String,
     pub h: u32,
     pub w: u32,
@@ -36,6 +37,7 @@ pub struct TagsForWorkspace {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct DisplayWorkspace {
     pub id: usize,
+    pub output: String,
     pub h: u32,
     pub w: u32,
     pub x: i32,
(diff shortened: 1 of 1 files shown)


VuiMuich:
I'd love to test this, but things are not smooth yet on my new NixOS installation.

@YannikSc, would you have a chance to give this PR a spin and confirm the fix?

---
Details for the curious mind about my NixOS issues: So with the workflow as discussed on discord yesterday, whenever I reload after binaries are changed leftwm-state breaks and doesn't output anything at all.
I had this when switching from 0.4.1 -> 0.4.2 and now from `HEAD`->`PR1116`.

YannikSc:
Hey, I just tested it with two external monitors attached and everything worked as expected.
The workspaces have their output field and it matches with the tag, that's displayed on it!

Thank you very much for implementing this!

PR #1120, opened 2023-08-21
Space comments consistently

Eskaan:
# Description

Spaced the doc comments consistently
Also silently fixing a self-made typo

## Type of change

# Checklist:


VuiMuich:
:zany_face: "neccesary" is a classic for me

It feels like those spaces silently sneak away every now and then, in my memory I believe to have fixed this a couple of times already, but they keep disappearing. Thanks for this consistency chore :grinning:

PR #1121, opened 2023-08-21
add `ToggleMaximized` command

### LINKED ISSUE #973: Maximize window
### Describe the feature

Coming from awesomeWM there is the concept of fullscreen and maximized. Fullscreen covers the entire screen, maximized covers everything except the bar (gets rid of the useless gaps and other windows)

Not sure how this would be implemented in LeftWM but would be useful when doing focused work but still wanting to e.g. see the clock in polybar.

Would be nice to have

BKSalman:
# Description

Add a `ToggleMaximized` command that covers everything except the bar, to make it easy to focus on a window without getting rid of the bar

Fixes #973

## Type of change

[External command wiki page](https://github.com/leftwm/leftwm/wiki/External-Commands) should be updated

# Checklist:

- [x] add `ToggleMaximized` command
- [ ] ~~Maximize next/previous window when switched to (like when using monocle layout)~~
- [x] Add tests (like the fullscreen ones)
- [x] Keep floating windows on top
- [x] fix for floating windows not getting maximized

### DIFF
--- a/display-servers/xlib-display-server/src/xwrap/getters.rs
+++ b/display-servers/xlib-display-server/src/xwrap/getters.rs
@@ -446,13 +446,25 @@ impl XWrap {
     /// Returns the states of a window.
     #[must_use]
     pub fn get_window_states(&self, window: xlib::Window) -> Vec<WindowState> {
-        self.get_window_states_atoms(window)
+        let window_states_atoms = self.get_window_states_atoms(window);
+
+        // if window is maximized both horizontally and vertically
+        // `WindowState::Maximized` is used
+        // instead of `WindowState::MaximizedVert` and `WindowState::MaximizedHorz`
+        let maximized = window_states_atoms.contains(&self.atoms.NetWMStateMaximizedVert)
+            && window_states_atoms.contains(&self.atoms.NetWMStateMaximizedHorz);
+
+        let mut window_states: Vec<WindowState> = window_states_atoms
             .iter()
             .map(|a| match a {
                 x if x == &self.atoms.NetWMStateModal => WindowState::Modal,
                 x if x == &self.atoms.NetWMStateSticky => WindowState::Sticky,
-                x if x == &self.atoms.NetWMStateMaximizedVert => WindowState::MaximizedVert,
-                x if x == &self.atoms.NetWMStateMaximizedHorz => WindowState::MaximizedHorz,
+                x if x == &self.atoms.NetWMStateMaximizedVert && !maximized => {
+                    WindowState::MaximizedVert
+                }
+                x if x == &self.atoms.NetWMStateMaximizedHorz && !maximized => {
+                    WindowState::MaximizedHorz
+                }
                 x if x == &self.atoms.NetWMStateShaded => WindowState::Shaded,
                 x if x == &self.atoms.NetWMStateSkipTaskbar => WindowState::SkipTaskbar,
                 x if x == &self.atoms.NetWMStateSkipPager => WindowState::SkipPager,
@@ -462,7 +474,13 @@ impl XWrap {
                 x if x == &self.atoms.NetWMStateBelow => WindowState::Below,
                 _ => WindowState::Modal,
             })
-            .collect()
+            .collect();
+
+        if maximized {
+            window_states.push(WindowState::Maximized);
+        }
+
+        window_states
     }
 
     /// Returns the atom states of a window.
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -951,13 +951,13 @@ mod tests {
                     .iter_mut()
                     .find(|w| w.handle == window_handle)
                 {
-                    if window_state == WindowState::Fullscreen {
+                    if window_state == WindowState::Fullscreen
+                        || window_state == WindowState::Maximized
+                    {
                         if toggle_to {
-                            window.set_states(vec![WindowState::Fullscreen]);
-                        } else if let Some(state_pos) = window
-                            .states()
-                            .iter()
-                            .position(|s| *s == WindowState::Fullscreen)
+                            window.set_states(vec![window_state]);
+                        } else if let Some(state_pos) =
+                            window.states().iter().position(|s| *s == window_state)
                         {
                             let mut states = window.states();
                             states.remove(state_pos);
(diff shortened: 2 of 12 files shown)


leftwm-core/src/handlers/command_handler.rs
```diff
@@ -238,8 +239,8 @@ fn toggle_state(state: &mut State, window_state: WindowState) -> Option<bool> {
     let toggle_to = !window.has_state(&window_state);
     let tag_id = state.focus_manager.tag(0)?;
 
-    if window_state == WindowState::Fullscreen {
-        //Going to fullscreen, so we save the window order
+    if window_state == WindowState::Fullscreen || window_state == WindowState::Maximized {
+        //Going to fullscreen/maximized, so we save the window order
```
VuiMuich, line 243:
```suggestion
        // Going to fullscreen or maximized, so we save the window order
```
The spacing in the next line is fixed in `main`, but this would conflict.

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -269,6 +270,7 @@ fn toggle_state(state: &mut State, window_state: WindowState) -> Option<bool> {
     state.handle_window_focus(&handle);
     match window_state {
         WindowState::Fullscreen => Some(true),
+        WindowState::Maximized => Some(true),
         _ => Some(false),
```
VuiMuich, line 274:
```suggestion
        WindowState::Fullscreen | WindowState::Maximized => Some(true),
        _ => Some(false),
```
to make clippy happy

display-servers/xlib-display-server/src/xwrap/getters.rs
```diff
@@ -462,7 +474,13 @@ impl XWrap {
                 x if x == &self.atoms.NetWMStateBelow => WindowState::Below,
                 _ => WindowState::Modal,
             })
-            .collect()
+            .collect();
+
+        if maximized {
+            window_states.push(WindowState::Maximized);
+        }
```
VuiMuich, line 481: Can't we condense this in an `if-let` in line 454?
Maybe this would help to make the if in the match arms nicer as well.

I mean, that's not a real blocker, more like aesthetic.

leftwm-core/src/handlers/command_handler.rs
```diff
@@ -238,8 +239,8 @@ fn toggle_state(state: &mut State, window_state: WindowState) -> Option<bool> {
     let toggle_to = !window.has_state(&window_state);
     let tag_id = state.focus_manager.tag(0)?;
 
-    if window_state == WindowState::Fullscreen {
-        //Going to fullscreen, so we save the window order
+    if window_state == WindowState::Fullscreen || window_state == WindowState::Maximized {
+        //Going to fullscreen/maximized, so we save the window order
         //or else, we restore it!
```
VuiMuich, line 244:
```suggestion
        // Going to fullscreen or maximized, so we save the window order
        // or else, we restore it!

File: display-servers/xlib-display-server/src/xwrap/getters.rs (hunk 3 above)
Comment (BKSalman) [reply display-servers/xlib-display-server/src/xwrap/getters.rs:481]: I can't think of a way to do it in `if-let` syntax, could you provide an example?

File: leftwm-core/src/models/tag.rs [hunk 5]
```diff
@@ -242,6 +242,23 @@ impl Tag {
...
+            windows
+                .iter_mut()
+                .filter(|w| {
+                    w.has_tag(&self.id)
+                        && w.transient.unwrap_or_else(|| 0.into()) == handle
+                        && w.is_managed()
+                })
+                .for_each(|w| {
+                    w.set_visible(true);
+                });
```
Eskaan, line 261: This code is mainly repeating the above one, so how about 
```suggestion
        } else if let Some(window) = windows
            .iter_mut()
            .find(|w| w.has_tag(&self.id) && (w.is_maximized() || w.is_fullscreen()))
        {
            window.set_visible(true);
            window.normal = if w.is_fullscreen() { 
                workspace.xywh 
            } else { 
                Xyhw::from(workspace.rect())
            };
            let handle = window.handle;
            windows
                .iter_mut()
                .filter(|w| {
                    w.has_tag(&self.id)
                        && w.transient.unwrap_or_else(|| 0.into()) == handle
                        && w.is_managed()
                })
                .for_each(|w| {
                    w.set_visible(true);
                });
```
Note that this only works because you can only have one window either maximized or in fullscreen.

BKSalman: I still want to add different behaviour to maximized windows, for example the ability to cycle between windows without reordering them (still figuring it out), so I'm currently working on how that works, after I finish it, I will try to remove duplicated code and stuff

BKSalman:
ok I think this is ready for review + merge now

BKSalman:
resolved conflicts with main branch (hopefully I didn't mess anything up)

display-servers/xlib-display-server/src/xwrap/getters.rs
VuiMuich: My bad, I overlooked, that `window_state` is just declared in line 452...

VuiMuich (approved):
LGTM

Thanks Salman

BKSalman:
There is a small bug in my code, but will fix it in another PR, thanks for merging!! :)

VuiMuich:
Oops, didn't read it, sorry.
Thanks for your efforts, I like your spirit, just keep the PRs coming :grinning:

BKSalman:
haha, npnp, will get on it when I can

Eskaan:
I want to note a side-effect of the pr: By using the window Atoms, Applications can now call for themselves to me maximized. For example Factorio (a game) does this.

I really like this btw, it adds to the feature parity with the spec.

BKSalman:
> I want to note a side-effect of the pr: By using the window Atoms, Applications can now call for themselves to me maximized. For example Factorio (a game) does this.
> 
> 
> 
> I really like this btw, it adds to the feature parity with the spec.

Niiice, totally intended 💯🤓

VuiMuich:
> I want to note a side-effect of the pr: By using the window Atoms, Applications can now call for themselves to me maximized. For example Factorio (a game) does this.
> 
> I really like this btw, it adds to the feature parity with the spec.

Why am I not surprised you are also a fellow factorio player? :joy:

PR #1122, opened 2023-08-21
nixos-vm and flake changes

vesdev:
# Description
rewrote the flake to use flake-parts, crane and the oxalica overlay
which makes it a bit cleaner, getting rid of flake-utils and weird hacky merges.

also changed to leftwm.overlays.default from the deprecated leftwm.overlay
this will break for existing users of the overlay
will require a small rename to fix

but mainly adding a way to build a development environment on nixos conveniently
you just run `nix develop` then `left-vm` to run a vm with lefwm built from source

## Type of change

# Checklist:

`make test-full-nix` will produce a error since it expects a filesystem to be defined
but nixos-rebuild build-vm creates a ext3 root filesystem automatically
so this can be ignored


nixos-vm/configuration.nix
```diff
@@ -0,0 +1,38 @@
+{ config, pkgs, ... }:
+{
+  
+  system.stateVersion = "23.05";
+  nix.settings.experimental-features = [ "flakes" ];
+
+  boot.kernelPackages = pkgs.linuxPackages_latest;
+  i18n.defaultLocale = "en_US.UTF-8";
```
VuiMuich, line 8: Could we somehow inherit this from the host-system?

VuiMuich:
Just had a minute to give it a shot, but I get:
```
>>> ./nixos-vm/run.sh
building a leftwm virtual machine...
building the system configuration...
error: flake 'git+file:///home/vuimuich/Git/leftwm' does not provide attribute 'packages.x86_64-linux.nixosConfigurations."nixbook".config.system.build.vm', 'legacyPackages.x86_64-linux.nixosConfigurations."nixbook".config.system.build.vm' or 'nixosConfigurations."nixbook".config.system.build.vm'
./nixos-vm/run.sh: line 5: ./result/bin/run-leftwm-vm: No such file or directory
```
:neutral_face:

BKSalman:
tested it after latest commit, and it works now :+1:

vesdev:
ok so now its deriving the version from leftwm/Cargo.toml
doesnt need to be bumped by hand anymore.

and specified the filesystem so it silences the CI error

also figured out that error was just the hostname defaulting to the system one, should work now.

VuiMuich:
Ok, now it works! Woohooo :tada: 

I guess I would merge this now and if it is possible qol improvments like using same lokales and keyboard layout like the host system can be added in later PR.

VuiMuich:
Hmm, the last commit, where I thought I fixed that we miss to get the git-hash during compilation doesn't work in the VM.
The rust code for this is in the leftwm-check.rs and requires git as a build dependency, so I thought adding git to the deps should have worked.

nixos-vm/vm.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+left-vm () {
+  export SCRIPTPATH=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+  echo building a leftwm virtual machine...
+  echo $SCRIPTPATH
+  cd "$SCRIPTPATH"
+  nixos-rebuild build-vm --flake ../#leftwm
+  ./result/bin/run-leftwm-vm
```
VuiMuich, line 10: Is there a reason to stay in nixos-vm? Or could we `cd -` at the end of the script?

BKSalman: the script changes to the nixos-vm directory to have the `result` folder inside `nixos-vm`, and can be `.gitignore`d, but you can either `.gitignore` it both in the root of the project, and inside `nixos-vm` to make sure it's not included

a long way to say "yes, you can", but just giving more context

vesdev: it was a mistake on my part and had the fix locally already, cd - is the correct behaviour. And yeah the reason for it is just to have it dump the result and disk image in the folder regardless of what the current directory is, cleaner this way imo. so you can run the command safely in a folder like leftwm/src and it will get gitignored properly. Could use the root folder aswell but i personally think littering the root with 2 extra build files is a bit ugly.

VuiMuich:
is there any reason, why not `.nixos-vm` hide it, to even less clutter the project root?

vesdev:
> is there any reason, why not `.nixos-vm` hide it, to even less clutter the project root

thats a good idea

vesdev:
I dont see any other changes this would need.

there is a minor issue where remote-viewer auto resize does not work
even though spice vdagent is running and working. Not important enough to block merge

and getting the git hash during compilation issue, is also present on the upstream.
Could be a issue with both flakes? or in the source? if it works correctly without nix then the current flake is also broken.

vesdev:
after a digging a little bit i think i found why its failing to get the git hash.
git_version calls git describe during compilation which fails

relevant issue:
https://github.com/ipetkov/crane/issues/160

VuiMuich:
I am basically ok with this.
Just wanted to mention https://github.com/leftwm/leftwm/pull/1075, where this code was last time touched.
The original before #1075 surely was more elegant, but broken by some upstream changes idk if there possibly was some upstream fixes so we could go back to the original. If not I am fine to get this merged.

vesdev:
made it a bit more elegant but i think its more sensical to have GIT_HASH take priority when its defined so it short-circuits the git_version rather than be used as a fallback

leftwm/src/bin/leftwm-check.rs
```diff
@@ -36,11 +36,7 @@ async fn main() -> Result<()> {
     );
     println!(
         "\x1b[0;94m::\x1b[0m LeftWM git hash: {}",
-        if let Some(git_hash) = option_env!("GIT_HASH") {
-            git_hash
-        } else {
-            git_version::git_version!(fallback = "unknown")
-        }
+        option_env!("GIT_HASH").unwrap_or(git_version::git_version!(fallback = "unknown"))
```
VuiMuich, line 39: Thats perfect :star_struck:

VuiMuich:
@vesdev thanks for the great work! :tada:

PR #1136, opened 2023-08-27
Fix fullscreen resize and respect margins

Eskaan:
# Description

This pr supersedes #1134, I used a simpler solution to respect the margins.

I also fixed the issue where the window wouldn't drop out of fullscreen/maximized when resized/moved.

There is also a slightly changed resize behavior now: When resizing a tiled window, the back ones instantly take it's space instead of after letting the window go. Tell me if you like it, it's just a question of using the third forced update statement or not.

This pr closes #1134

## Type of change

# Checklist:

- [x] Clippy and fmt currently fail due to the new rust version, but nothing is related to my code
- [x] Tests run fine

### DIFF
--- a/leftwm-core/src/handlers/display_event_handler.rs
+++ b/leftwm-core/src/handlers/display_event_handler.rs
@@ -135,14 +135,22 @@ fn prepare_window(state: &mut State, handle: WindowHandle) {
         // Un-pin window if maximized or in fullscreen
         if w.is_fullscreen() {
             w.reset_float_offset();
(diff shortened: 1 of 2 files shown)


BKSalman:
ah, thanks for making it better!

my brain was fried when I did my PR lol

Eskaan:
Thanks @VuiMuich, guess that one slipped my eye

VuiMuich:
actually I got it fixed up, when I was checking why CI fmt failed on all recent PRs :joy: 

BTW those failures all where from `let-env` fmt stabilization.

VuiMuich:
Thanks @Eskaan, great work as always!

PR #1138, opened 2023-08-28
make floating windows stay on top of maximized windows

### LINKED ISSUE #943: Fullschreen mode stacking order setting
### Describe the enhancement

My main applications spawn allot of windows where i need to configure things.

Currently the fullshreen window always tries to stay on top which is incredibly annoying.
It whould be great if there where options to change the stacking order for fullschreen windows, preferably in the window_rules.

Preferably something like
```rust
    window_rules: [
        ( window_class: "Houdini FX", spawn_floating:  true , fs_stackorder : back),
    ],
```

BKSalman:
# Description

in my #1121 PR I meant to make floating windows go on top of maximized windows at all times, had some regression in this regard in #1126 

Fixes #943

## Type of change

nuffin

# Checklist:

### DIFF
--- a/leftwm-core/src/models/tag.rs
+++ b/leftwm-core/src/models/tag.rs
@@ -246,6 +246,7 @@ impl Tag {
             .iter_mut()
             .find(|w| w.has_tag(&self.id) && w.is_maximized())
         {
+            // Update maximized window
             window.set_visible(true);
             window.normal = Xyhw::from(workspace.rect());
             let handle = window.handle;
@@ -260,10 +261,12 @@ impl Tag {
                     w.set_visible(true);
                 });
 
-            // Update the location and visibility of non-normal + non-maximized windows.
+            // Update all windows except normal non-floating windoows and maximized window
             windows
                 .iter_mut()
-                .filter(|w| w.has_tag(&self.id) && !w.is_normal() && !w.is_maximized())
+                .filter(|w| {
+                    w.has_tag(&self.id) && (!w.is_normal() || w.floating()) && !w.is_maximized()
+                })
                 .for_each(|w| {
                     w.set_visible(true);
                     // Don't change docks and desktop xyhw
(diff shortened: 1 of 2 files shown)


VuiMuich:
@BKSalman Have you checked how much freedom there is in the xorg specs to define additional layers? I have a vague memory, that six was matching the specs exactly.
I would prefer to work out of specs or explicitly specified allowance to to extend the default specs.

BKSalman:
> Have you checked how much freedom there is in the xorg specs to define additional layers? 

honestly I haven't checked, it just made sense to me to do this

I will check if this is allowed or not

BKSalman:
> Layered stacking order

> Some window managers keep the toplevel windows not in a single linear stack, but subdivide the stack into several layers. There is a lot of variation among the features of layered stacking order implementations. The number of layers may or may not be fixed. The layer of a toplevel window may be explicit and directly modifiable or derived from other properties of the window, e.g. the type of the window. The stacking order may or may not be strict, i.e. not allow the user to raise or lower windows beyond their layer. 

reference: https://specifications.freedesktop.org/wm-spec/latest/ar01s02.html#idm46635829501088

does this mean it can be any number of layers?

VuiMuich:
👍 I'd say yes. And if I understand correctly it might even allow for custom order, like having a config option setting up stacking preferences/rules, which is very interesting imo.

VuiMuich:
Thanks and BTW congrats on the iconic PR number :grinning:

BKSalman:
Thank you for merging ( and for tolerating me :) )

> on the iconic PR number

umm, I'm not cultured enough to know this lol...

VuiMuich:
Iirc 1138 was the first (or one of the first) films by Geoerge Lucas and therefore got multiple times referenced in Star Wars and Indiana Jones movies.

BKSalman:
> Iirc 1138 was the first (or one of the first) films by Geoerge Lucas and therefore got multiple times referenced in Star Wars and Indiana Jones movies.

aah I see!

> Btw: We are not tolerating you, we appreciate your contributions a lot. In times like these, where the core team is occupied by other projects, its contributors like you who keep projects going and make it feel alive.

I appreciate your efforts too! thank you for this awesome window manager (pun not intended)

PR #1140, opened 2023-09-03
Flake fix + flake mold addition

BKSalman:
# Description

I was checking the output of `leftwm state` on my machine and it was showing workspaces with empty `output` field, so I read the code, and saw that only xrandr supports screen names, so I added xrandr to the flake.

another addition I did to the flake is adding [mold](https://github.com/rui314/mold), which is a faster linker, the way it's added is quite hacky, but it works.

@VuiMuich @vesdev could you please test it on your machine to make sure it actually works

## Type of change

nuffin

# Checklist:


VuiMuich:
I will probably have time to get a look earliest mid week.

BKSalman:
ok removed the hacky mold part, and actually checked the binary outputted to see if it actually used mold using

```bash
readelf -p .comment <executable-file>
```

flake.nix
```diff
@@ -65,7 +70,14 @@
           # `nix develop`
           devShells.default = pkgs.mkShell
             {
-              buildInputs = commonArgs.buildInputs ++ [ pkgs.pkg-config pkgs.systemd ];
+              NIX_CFLAGS_LINK = "-fuse-ld=mold";
+              
+              buildInputs = with pkgs;[
+                mold
+                clang
```
VuiMuich, line 77: tbh I am not sure what to think about mold.
In general it looks promising and I know it is in stable release now, but I don't have a clue about the implications for the binaries produced.

BKSalman: I only use it because it's a faster linker, nothing else

flake.nix
```diff
@@ -21,7 +21,7 @@
           "aarch64-linux"
         ];
 
-        perSystem = { pkgs, system, ... }: 
+        perSystem = {pkgs, system, ... }:
```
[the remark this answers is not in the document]
BKSalman: I could add a formatter to the flake so it stays consistent

flake.nix
VuiMuich: Maybe this is just my ignorance of how linking actually works. 
As I understood this so far the linker is building the final binary, right? So isn't there a "risk" that different linkers result in different binaries and in the worst case - caused by a linker bug - result in a bug of the binary itself?

I just want to be sure we don't place a potential foot gun here, that might send us on a potentially nasty bug hunt.
Also (not that I have done that yet, but eventually might look into it, or someone else wants to use it) how well does mold integrate into IDE debuggers? AFAIK most of the IDEs rely on llvm for debugging rust. 

So far I only found comparisons of speed for lld vs mold, but if you could educate me on the topic and clear out my worries, I would be happy to go for mold.

PR #1143, opened 2023-09-07
Pinned tags

### LINKED ISSUE #978: Certain tags should only open in certain monitors
### Describe the feature

Be able to assign a tag to a specific monitor, so that they don't open anywhere without it. This is how we set this up in xmonad:
image
Workspace 3 will only open on monitor 1 with this keybind.
i3 handles it [like so](https://i3wm.org/docs/userguide.html#workspace_screen).

LeftWM is unusable without it

BKSalman:
# Description

this PR is for adding the `pinned_tags` feature to the config, I'm submitting it even tho it's still WIP, but this could allow for reviews and critiques early on so it doesn't go too far without reviews and stuff

for now this supports pinning tags to a workspace, but lacks handling some edge cases

(also git history sucks but will fix it later)

the current breaking change is having to use `{ output: ... }` for the `workspaces` list, instead of `( output: ... )`

Fixes #978

## Type of change

add to the `pinned_tags` field of the config to the wiki page 

# Checklist:

# TODOs
- [x] add pinned_tags field to config (maybe could use tag index instead of tag name)
- [x] pin tags to workspaces
- [x] don't swap pinned tags out of the workspaces it's pinned at
- [x] fallback to a tag when moving a non-pinned tag to a workspace with pinned tags (swapping with pinned tags is prevented)
- [ ] move mouse to pinned tags when switching to it
- [ ] fix multiple workspaces with pinned tags
- [ ] add tests

### DIFF
--- a/display-servers/xlib-display-server/src/lib.rs
+++ b/display-servers/xlib-display-server/src/lib.rs
@@ -144,10 +144,11 @@ impl XlibDisplayServer {
     /// Return a vec of events for setting up state of WM.
     fn initial_events(&self, config: &impl Config) -> Vec<DisplayEvent> {
         let mut events = vec![];
-        if let Some(workspaces) = config.workspaces() {
+        if let Some(mut workspaces) = config.workspaces() {
             let screens = self.xw.get_screens();
-            for (i, wsc) in workspaces.iter().enumerate() {
-                let mut screen = Screen::from(wsc);
+            for (i, wsc) in workspaces.iter_mut().enumerate() {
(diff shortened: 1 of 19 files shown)


BKSalman:
(will remove all irrelevant random debugging logging)

BKSalman:
Currently this has bugs, but could benefit from an early review

BKSalman:
Yo, I'm sorry I kinda ditched this PR, but a lot has happened since then, and I don't have a lot of time to work on this unfortunately (also the implementation kinda sucks)

so I think I will close this PR if that's ok

PR #1156, opened 2023-10-19
Refactor window sorting

Eskaan:
# Description

- Refactor to the window sorting to a cleaner solution.
- Refactor `Window.states` to `pub` (set and get methods were there, so why not make it it pub)
- Move from manually getting transient windows whilst sorting windows to adding the `Above` WM state to them.
- Side-effect: We now sort windows with the `Above` WM state on top of everything. Windows set this property on their own.

## Type of change

  - As the window sorting now understands the `Above` WM hint, windows can set themselves to be on top.
  - This pr places unmanaged windows above **all** windows instead of all but fullscreen. A visible unmanaged window is yet to be found, so it's not marked as breaking.

# Checklist:

### DIFF
--- a/display-servers/xlib-display-server/src/lib.rs
+++ b/display-servers/xlib-display-server/src/lib.rs
@@ -100,7 +100,7 @@ impl DisplayServer for XlibDisplayServer {
             DisplayAction::Unfocus(h, f) => from_unfocus(xw, h, f),
             DisplayAction::ReplayClick(h, b) => from_replay_click(xw, h, b),
             DisplayAction::SetState(h, t, s) => from_set_state(xw, h, t, s),
-            DisplayAction::SetWindowOrder(fs, ws) => from_set_window_order(xw, fs, ws),
+            DisplayAction::SetWindowOrder(ws) => from_set_window_order(xw, ws),
             DisplayAction::MoveToTop(h) => from_move_to_top(xw, h),
             DisplayAction::ReadyToMoveWindow(h) => from_ready_to_move_window(xw, h),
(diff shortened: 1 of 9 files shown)


Eskaan:
@VuiMuich I am currently questioning myself above what windows such a transient window should be placed. Should we stay at the old solution (if parent fullscreen, above all, otherwise threat normal)?
I could also go with my next pr and automatically mark such windows as `OnTop` so that you can push them back on demand.

VuiMuich:
How would pushing them back on demand work? Would it be a new command or just with unfocusing?
Generally this sound like a nice idea to me.
The old solution was just done like this for the lack of a better idea, I guess.

Eskaan:
> How would pushing them back on demand work?

I want to implement the `OnTop` Window State that, when enabled and the window is floating, forces the window to the first layer. Then I would give every transient window which has a fullscreen parent that window state on window initialization. You can then choose to use the `ToggleOnTop` Command to remove the Window State so that it is behind the fullscreen window again.

Eskaan:
I implemented it exectly like the old window sorting for now, this should be ready for review.

mautamu (approved):
Been testing this for a while in local and it appears to be working fine. Looks good from a code perspective also; thanks Eskaan!

Eskaan:
Two problems:
1. c9681f49c4d62471e6d8c44011c49aff5d58a3d6 The `WindowState::Above` solution wasn't really working: Only newly created windows would be affected and those would keep the `Above` state even when the parent went out of fullscreen.
2. 1c0700165b9ff3d2eba3fd73e35419efe52e704b The maximized windows were on the wrong layer.

PR #1158, opened 2023-10-22
Rework fullscreen order restoring

Eskaan:
- [x] **Blocked** - waiting for #1156. This is currently based ontop of `Eskaan:refactor-window-sorting`, as it has conflicts with this branch. Will be rebased as soon as it gets merged. Only additional commit is 489155e292d80004949ef96ff63238ebcdaadabf.

# Description

Massively simplify #1052 by removing [line 124](https://github.com/leftwm/leftwm/blob/7628b688be4aafb155b1efdd09125de891e212a7/leftwm-core/src/state.rs#L124) from state.rs

Also a non-noticeable performance improvement by not having to do all the moves to restore the order.

## Type of change

# Checklist:


mautamu (approved):
Thank you so much for this Eskaan! Looks good from my local and the code is so much cleaner now :confetti_ball:

PR #1166, opened 2023-11-04
fixed dialog windows ignoring application-defined sizes

### LINKED ISSUE #1118: Newly opened floating windows with specific dimensions do not get sized properly
### Describe the bug

Some floating windows are sized improperly when they are opened, it seems to be only floating windows that have set sizes in the application that don't always get sized correctly.

Here is a screenshot of an example of the GIMP dds plugin popup that shows up when you open a dds file.
gimp_dds_example

With all other window managers I've used this dds popup window would be minimized and to the smallest size it would scale the window nicely. But with leftwm the window doesn't minimize automatically like I would expect it to and there isn't a window_rules option to automatically set the dimensions of a newly spawned window so I don't see a way around it other than manually resizing the window every time it is opened.

Here is a screenshot of the expected behavior, this is using qtile on the exact same machine.
gimp_expected_behavior

1. Download gimp 2.10
2. Find a dds file and open it with gimp
3. See the difference in behavior between the newly spawned leftwm floating windows and other window manager floating windows.

### OS / Distro

Arch and Garuda

### LeftWM Check

```text
:: LeftWM version: 0.4.2
:: LeftWM git hash: unknown
    -> journald-log OK
    -> lefthk OK
    -> Binaries OK 
    -> Configuration loaded OK 
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK 
:: Checking theme . . .
    -> Theme OK
```

Silicasandwhich:
# Description

Fixed an issue where leftwm would ignore the requested size of dialog-type windows when they are spawned, particularly when they are spawned without a parent. The previous behavior (which appears to be filling in the space left by the xyhw avoided) is still default when there is no desired size given. Additionally, this change has been carried over to splash screen windows and floating normal windows, just in case this issue applies to other windows with those types.

Fixes #1118 

## Type of change

N/A

# Checklist:

### DIFF
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -440,17 +440,27 @@ fn setup_window(
         WindowType::Normal => {
             window.apply_margin_multiplier(ws.margin_multiplier);
             if window.floating() {
-                set_relative_floating(window, ws, ws.xyhw);
+                match window.requested {
+                    Some(xyhw) => set_relative_floating(window, ws, xyhw),
+                    None => set_relative_floating(window, ws, ws.xyhw),
+                };
             }
         }
         WindowType::Dialog => {
             if window.can_resize() {
                 window.set_floating(true);
-                let new_float_exact = ws.center_halfed();
-                window.normal = ws.xyhw;
-                window.set_floating_exact(new_float_exact);
+                if let Some(xyhw) = window.requested {
+                    set_relative_floating(window, ws, xyhw);
+                } else {
+                    let new_float_exact = ws.center_halfed();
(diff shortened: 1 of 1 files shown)


VuiMuich:
An far as I can tell this works well and even rids something I just recently observed in qutebrowser, where dialog windows that are opened repeatedly (like file selection for upload) only got sized correctly on the first instance and every further instance got sized at the minimum defaults. Nice, thanks.
If no one reports anything hindering, I'd like to merge this tomorrow in the morning.

mautamu (approved):
Looks good from a code perspective and from a testing perspective. Thanks for this, Silicasandwhich!

PR #1178, opened 2023-11-15
Fix wrong version number in `leftwm/Cargo.toml`

VuiMuich:


adamcstephens:
FYI this won't be sufficient for the nix pacakge. 

```
error: the lock file /build/source/Cargo.lock needs to be updated but --frozen was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --frozen flag and use --offline instead.
error: builder for '/nix/store/i2m0zf9g93j37r19dlgq83ifyf69i0hr-leftwm-0.5.0.drv' failed with exit code 101;
```

VuiMuich:
> FYI this won't be sufficient for the nix pacakge. 
> 
> 
> 
> ```
> 
> error: the lock file /build/source/Cargo.lock needs to be updated but --frozen was passed to prevent this
> 
> If you want to try to generate the lock file without accessing the network, remove the --frozen flag and use --offline instead.
> 
> error: builder for '/nix/store/i2m0zf9g93j37r19dlgq83ifyf69i0hr-leftwm-0.5.0.drv' failed with exit code 101;
> 
> ```

🤦🏼‍♂️ of course! That was a stupid thing to overlook when pushing this commit. 

Well we gonna have a 0.5.1 anyways 😇

PR #1181, opened 2023-11-18
feat: add AUR publishing capabilities

mautamu:
# Description

Adds the capability to publish crate updates to the AUR. Although I am reasonably confident in this build, I'm not certain it's finished. Additionally, we will need to add the secrets to GitHub to finalize this change.

## Type of change

# Checklist:

  Note: To fully reproduce CI checks, you will need to run `make test-full-nix`. Usually, this is not necessary. [N/A]

### DIFF
--- a/leftwm/release/PKGBUILD
+++ b/leftwm/release/PKGBUILD
@@ -1,0 +1,61 @@
+# Maintainer: Lex Childs <lexchilds@gmail.com>
+# Contributor: éclairevoyant
+
+pkgbase=leftwm
+pkgname=(leftwm leftwm-nonsystemd)
+pkgver=0.5.1
+pkgrel=2
+pkgdesc="A tiling window manager for the adventurer"
+arch=('i686' 'x86_64')
+url="https://github.com/leftwm/leftwm"
+license=('MIT')
+depends=(gcc-libs)
+makedepends=('cargo' 'git')
+optdepends=('bash: themes'
+            'dmenu: default launcher'
+            'eww: flexible status bar'
+            'feh: used to set background images'
+            'lemonbar: light weight bar'
+            'polybar: light weight bar')
+source=("${pkgbase}-${pkgver}.tar.gz::${url}/archive/refs/tags/${pkgver}.tar.gz")
+sha256sums=('3c8ab0fdbfe205b33ad7ae108d3a604bdd22663458bf803e0a3a4a924aad963a')
+install='readme.install'
+
+prepare() {
+	cd $pkgbase-$pkgver
+	cargo fetch --target "$CARCH-unknown-linux-gnu"
+}
+
+build() {
+	cd $pkgbase-$pkgver
+
+	export CARGO_TARGET_DIR=target_non_systemd
+	cargo build --frozen --release --no-default-features --features=lefthk,sys-log
+
+	export CARGO_TARGET_DIR=target_systemd
+	cargo build --frozen --release
+}
+
+_package() {
+	install -Dm755 leftwm{,-worker,-state,-check,-command} lefthk-worker -t "$pkgdir"/usr/bin/
+
+	cd ../../
+	install -Dm644 leftwm/doc/leftwm.1 -t "$pkgdir"/usr/share/man/man1/
+	install -d "$pkgdir"/usr/share/leftwm
+	cp -R themes "$pkgdir"/usr/share/leftwm
+	install -Dm644 leftwm.desktop -t "$pkgdir"/usr/share/xsessions/
+	install -Dm644 LICENSE.md "$pkgdir"/usr/share/licenses/leftwm/LICENSE
+}
+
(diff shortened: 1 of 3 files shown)


leftwm/release/PKGBUILD
```diff
@@ -0,0 +1,61 @@
+# Maintainer: Lex Childs <lexchilds@gmail.com>
+# Contributor: éclairevoyant
+
+pkgbase=leftwm
+pkgname=(leftwm leftwm-nonsystemd)
+pkgver=0.5.1
+pkgrel=2
```
VuiMuich, line 7:
```suggestion
pkgrel=1
```
We want to initialise the pakage bump with value `1` as increments only should happen on changes of the packaging script itself while the version stays unchanged.

VuiMuich:
Plus we need @leftbot added as comaintainer on all of the AUR packages. Afaik only @lex148 has permision to add new maintainers.

Eskaan (commented):
Is there a way to move the systemd/non-systemd part to the build function? I don’t really like to see it build twice every time.

leftwm/release/PKGBUILD
mautamu: Good callout! I think that this is corrected in the process of building/updating sha as well but not 100% certain of that fact.

hertg:
@mautamu If you want to see a reference workflow for automatic AUR releases, I implemented that in another project of mine. It looks like you are coming to a very similar solution.

https://github.com/hertg/lightdm-neon/blob/main/.github/workflows/aur-release.yml

https://github.com/hertg/lightdm-neon/blob/99724786c5ad1e73cac762dc1ce4016e9c263bca/.github/workflows/release.yml#L29-L38

https://github.com/hertg/lightdm-neon/blob/main/.ci/generate-pkgbuild.sh

mautamu:
@hertg Would like your thoughts on this as part of #1258 as well. This may be something that we'd reject depending on the strategy there.

Eskaan:
*Referencing my review above*
> Is there a way to move the systemd/non-systemd part to the build function? I don’t really like to see it build twice every time.

I don't really know all about the PKGBUILDs build functions, but could you create a non-systemd build function instead so it doesn't build twice?

hertg:
> @hertg Would like your thoughts on this as part of #1258 as well. This may be something that we'd reject depending on the strategy there.

Not sure how you mean? Generally, I think it would be a good idea to publish things via CI, because then the exact publishing process is "documented" since it was scripted out, and the publishing itself is quite transparent, since it happens via a workflow that can be audited by users. When it comes to signing, I'm not sure how we should handle that.  I'm not really experienced myself when it comes to that, and we should probably check out how other projects handle it.  (I am also a bit out of the loop when it comes to building leftwm, with all the syslog and systemd feature flags...)

Should some of the core devs sign releases manually? How could we do that, and still keep the release process as simple and transparent as possible? Should *leftbot* be able to sign stuff? If so, how do we protect that private key, and how would we know if it was compromised? I don't have the answers to those questions and it would be great if we could collect some resources / case studies on how other (large) foss projects do this... Maybe it would be possible to trust the keys of a few core devs, let them update and sign the leftbot key every few months or so, and then leftbot can sign releases by itself? I have to look more into PGP signing to figure out if that even makes sense.

---

There's also another discussion to be had, which CI we want to use. You mentioned that you'd like to use codeberg more, and I feel the same way. I already sponsor them [via liberapay](https://liberapay.com/codeberg/), but I'm also thinking about [joining their association](https://join.codeberg.org/). They also [offer CI capabilities](https://docs.codeberg.org/ci/), which I think we'd need to apply for first. But all this (codeberg / github) seems like another meta discussion, that we probably should move to a separate thread.

hertg:
**Minor update**
 I've dug a bit into Codeberg's CI. In addition to the hosted Woodpecker CI mentioned above, it is also possible to host one's own Woodpecker CI, or to use [Forgejo Actions](https://forgejo.org/docs/latest/user/actions). Codeberg [does not yet provide built-in runners for Forgejo Actions](https://docs.codeberg.org/ci/actions/), but it is fairly easy to set up, and connect runners to Codeberg. I tried it out in private and it seems to work alright. I migrated a small personal Rust project and the Github Action workflows only needed minor adjustments (for example, the actions linked in `uses:` must use an absolute domain, this way you can even use actions hosted on Github):

```yaml
- uses: https://github.com/actions-rs/toolchain@v1
  with:
    toolchain: stable
```

When the action is referenced only by relative name, it will use `https://code.forgejo.org/` by default. They offer a few "built-in" actions already: https://code.forgejo.org/actions

It looks like they try to be as compatible with Github Actions as possible, which I find great. I'll try this out a bit more in the coming weeks, as I intend to move some personal projects from Github over to Codeberg. 

---

Btw. I have some free computing capacity on a server cluster, and I'd be happy to provide some Forgejo runners for LeftWM.

PR #1197, opened 2023-12-14
Fix freeze when process tries to aquire stderr

### LINKED ISSUE #1196: Lefthk freezes when keybind program tries to aquire terminal buffer
### Describe the bug

I investigated an issue I had with leftwm freezing (expressed [here](https://github.com/leftwm/leftwm/issues/1168#issuecomment-1793529189)), where leftwm froze when I tried to execute `npc n` using rofi opened by a keybind.

I think I narrowed it down to programs that try to aquire the terninal's second display (or whatever it’s called). `npc n` autocompletes to `ncmpcpp` in rofi, and I was able to reproduce the issue by setting a keybind to `ncmpcpp` as well as `htop`.

I am unsure if this is a lefthk or leftwm issue at the moment, so I am posting it here. If it’s a lefthk problem, we should still try to handle it somewhat gracefully and try to at least keep program key inputs (e.g. for saving you currently edited file before killing leftwm).

leftwm dosen’t freeze to a point where it has to be terminated with SIGKILL.

1. create a keybind for htop (or any other program that uses the second display)
  `(command: Execute, value: "htop", modifier: ["modkey", "alt"], key: "f")`
2. reload to apply keybind
3. press keybind
4. End up with frozen keyboard inputs for both lefthk and programs (mouse works fine, can spawn windows from tty2.
5. leftwm needs to be killed from a tty (killing lefthk does not work).

```text
No output from leftwm regarding the problem, no terminating output due to being killed by SIGKILL
```

### OS / Distro

Arch Linux on 6.6.4-zen kernel

_No response_

### LeftWM Check

```text
:: LeftWM version: 0.5.1
:: LeftWM git hash: 0.5.1-4-eaa13b37
    -> journald-log OK
    -> lefthk OK
    -> Binaries OK
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK
```

Eskaan:
# Description

When a scratchpad tries to aquire stderr (experienced with programs that use a "second display" like htop, man or ncmpcpp)
I don't know the exact cause, but above is assumed as disabling stderr fixes it.

I first experienced it with lefthk (will be fixed with leftwm/lefthk#38), but it also was the case for scratchpads (tested) and is assumed to be for some other cases like the autostart desktop files (code changed for those, didn't explicitly test them).

Fixes #1196 

## Type of change

# Checklist:


Eskaan:
@mautamu there you go

mautamu:
Thanks @Eskaan!

For this, assuming we want to release https://github.com/leftwm/lefthk/pull/38 first, version bump lefthk-core/lefthk to 0.2.1, and then release this one after updating leftwm/Cargo.lock (which should automatically encompass 0.2.1 since currently we use "0.2" in leftwm/Cargo.toml): https://github.com/leftwm/leftwm/blob/5293219827bf0d2a982bf548b5fdbb0c4cc7d235/leftwm/Cargo.toml#L21 

Will plan to do this later this weekend if there aren't any objections.

PR #1198, opened 2023-12-18
Move clippy allowance to xlib-display-server, deprecate `make test-full`

Eskaan:
# Description

Moved the allowance of sign-casting stuff to only xlib-display-server, where we actually to such things (when operating with the c library).
Also deprecate `make test-full` and move the pedantic checks to `make test`.

## Type of change

# Checklist:


Makefile
```diff
@@ -25,24 +25,17 @@ all: build test
...
-		-A clippy::must_use_candidate\
-		-A clippy::cast_precision_loss\
-		-A clippy::cast_possible_truncation\
-		-A clippy::cast_possible_wrap\
-		-A clippy::cast_sign_loss\
-		-A clippy::mut_mut
+	@printf '\n\033[38;5;9mNote: \"make test-full\" is deprecated. You can use \"make test\" instead with the same functionality\n'
 
-test-full-nix: test-full test-nix
+test-full-nix: test test-nix
```
mautamu, line 38: Do we want to rename this to `test-nix` as opposed to `test-full-nix`?

mautamu (commented):
This looks good overall, thank you Eskaan!

Makefile
Eskaan: I already thought about that, but I didn’t rename it because it would mean another breaking change. I guess no one is using that in an automated script anyways so I can deprecate it too.

Eskaan: After looking at the code, I noticed that the nix test itself is already named `test-nix`. So do you want `test-nix` to also run `test` or just deprecate `test-full-nix` with the note to run both as a replacement?

mautamu: Ah yeah; I can see the use case where we would want to keep both `test-nix` and `test-nix-full` separate. I think we can leave it unless there are concerns.

PR #807, opened 2022-07-30
General Refactoring

AethanFoot:
# Description

Just some general refatoring, moving things where they should be and hiding method that are only used locally and try and order the functions in a more searchable way (alphabetically? or any other ideas?).
Many Thanks (finally luafied my neovim so can concentrate on this)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Updated user documentation:

Please insert user documentation that should be updated (as in the wiki).

See [CONTRIBUTING.md User Documentation section](../CONTRIBUTING.md#user-documentation) for further details.

**Note: Manual page changes must be performed in a commit, not in this PR section.**

# Checklist:

- [ ] Ran `make test-full` locally with no errors or warnings reported
- [ ] Manual page has been updated accordingly
- [ ] Wiki pages have been updated accordingly (to perform **after** merge)

PR #1200, opened 2023-12-19
Rework "fixed dialog windows ignoring application-defined sizes (#1166)"

Eskaan:
# Description

I wasn't able to do a full review of #1166 before it was merged, but I had taken a look at the code and already noticed that the solution seemed complicated. However I kept it as a unread notification until when I had time (now).

Basically, the `set_relative_floating` function already respects all size requests, it just wasn't called.

This also corrects the mistake made before that (not respecting size hints from resizable dialog windows). However I wasn't able to trace that further than #807 (e2517de95)

## Type of change

# Checklist:


mautamu (approved):
Thank you for catching this, Eskaan!

PR #1211, opened 2024-01-08
using move_to_top when windows take focus

### LINKED ISSUE #1206: floating vs floating should indeed restack when moving focus.
floating vs floating should indeed restack when moving focus.

When changing focus (say with modifier+k or modifier+j) between floating windows, they do not restack, which seems counter-intuitive to me

_cf post by @VuiMuich in https://github.com/leftwm/leftwm/issues/1094#issuecomment-1596535187_

barkyq:
# Description

using move_to_top when windows take focus

Fixes #1206

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [ ] Ran `make test` locally with no errors or warnings reported


Eskaan (changes requested):
Moving a window on top here is only temporary as it's still in a different place in the focus managers window stack and will be sorted up front as soon as `sort_windows` is called again (for example when a new window is opened).

barkyq:
changed to draft PR, need to figure out bugs with focus changing when switching tags etc.

PR #1215, opened 2024-01-09
Spawn new floating windows above old ones

Eskaan:
# Description

Fixes what presumably is a regression I caused in #1158 where newly spawned and previously tiled floating windows aren't put up front but behind already existing ones. 

> [!NOTE]  
> I assume ignoring the `inset_behavior` config be intended behavior here.

## Type of change

# Checklist:


barkyq:
merged into my personal branch, no issues

mautamu:
> I assume ignoring the inset_behavior config be intended behavior here.

I believe so too; if not, I think we can probably feature gate alternatives.

Thanks again Eskaan and barkyq!

PR #1217, opened 2024-01-10
feat(mouse handler): allow user to swap resize and move using ShiftMask

### LINKED ISSUE #1213: Add a way to resize floating windows with the left mouse button.
### Describe the feature

It would be convenient for touchscreen users to have a way to resize windows in a way that did  not require the right mouse button.

I didn't put much thought into my hard-coded solution [here](https://github.com/leftwm/leftwm/compare/main...WhitingJarod:leftwm:main), but I could submit a PR if desired.

A more involved solution might be to add mouse binding options to the configuration file.

Would be nice to have

I don't know, may be a niche edge-case

barkyq:
# Description

Make it so `shift + mousekey + (left) click + drag` resizes windows and `shift + mousekey + (right) click + drag` moves windows. 

Remove static quantities from state struct.

Fixes #1213 

## Type of change

Holding down `mousekey` while left-clicking is used to move windows, and holding down `mousekey` while right-clicking is used to resize windows. Alternatively, one can hold `mousekey` and Shift while left-clicking to resize windows or `mousekey` and Shift while right-clicking to move windows.

# Checklist:

- [X] Ran `make test` locally with no errors or warnings reported

### DIFF
--- a/leftwm-core/src/utils/modmask_lookup.rs
+++ b/leftwm-core/src/utils/modmask_lookup.rs
@@ -8,7 +8,7 @@ bitflags! {
         /// Used as the zero value
         const Zero = 0;
         const Any = 1;
-        const Shift = 1 << 1;
+        const Shift = 1 << 0;
         const Control = 1 << 2;
         /// Mod1
         const Alt = 1 << 3;
--- a/leftwm-core/src/utils/modmask_lookup.rs
+++ b/leftwm-core/src/utils/modmask_lookup.rs
@@ -9,6 +9,8 @@ bitflags! {
         const Zero = 0;
         const Any = 1;
         const Shift = 1 << 0;
+        /// Caps Lock
+        const Lock = 1 << 1;
         const Control = 1 << 2;
         /// Mod1
         const Alt = 1 << 3;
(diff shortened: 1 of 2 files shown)


barkyq:
Some of my changes are opinionated, and happy to change anything which rubs you the wrong way!

mautamu:
This looks good to me but will wait to merge to see if anyone else has any comments since I'm admittedly not a heavy user of this functionality. Thanks again for all your changes this week :) !

Feel free to rebase and force push [7d86f72](https://github.com/leftwm/leftwm/pull/1217/commits/7d86f727a27de7b08c8569ffa896216623b62973) away if you'd like.

Eskaan (changes requested):
You don't have to move the implementation to the Manager to get direct access to the config, we already load it into the State struct.

barkyq:
I guess I was confused why these configuration settings are loaded into the `state`, since these settings do not change. As far as I can tell, the state should be for storing information which is changed by the program. 

Moreover, most of the other methods in display_event_handler.rs use the `Manager` rather than the state.

barkyq:
related to the discussion in #1219 about what config variables should be stored in `State` struct.

NicTanghe:
is this dead?

mautamu:
@NicTanghe This is ready to test if you'd like. Should be working now.

leftwm-core/src/utils/modmask_lookup.rs
```diff
@@ -8,7 +8,7 @@ bitflags! {
         /// Used as the zero value
         const Zero = 0;
         const Any = 1;
-        const Shift = 1 << 1;
+        const Shift = 1 << 0;
```
mautamu: Saw this in https://tronche.com/gui/x/xlib/events/keyboard-pointer/keyboard-pointer.html

This is the part that probably needs reviewed.

PR #1218, opened 2024-01-10
default_height and default_width

barkyq:
# Description

remove default_width and default_height from State struct, get from Config instead.

Fixes #1212

## Type of change

- [X] Development change (no change visible to user)
- [X] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [X] Ran `make test` locally with no errors or warnings reported.

PR #1219, opened 2024-01-11
Refactor theme config loading

Eskaan:
# Description

Slightly rework the way the theme config is loaded into leftwm and untangle some of the color loading.

Fixes #(1213)

# Changes
> [!Note]
> This pr is best reviewed per-commit, as there are a lot of renames.

- 2eace8188a226ebace36b7ddabb03d4b6604eb8d
  Add `default_width/height` to loaded theme vars (replaces #1218 by a simpler solution)
  https://github.com/leftwm/leftwm/blob/2eace8188a226ebace36b7ddabb03d4b6604eb8d/leftwm-core/src/state.rs#L181-L182

- 5debd321132ce0b18c5c486319be3b662b291837
  Xlib Display Server `new()` now first loads the config into `Xwrap` and then calls `init()` without the config param.
  https://github.com/leftwm/leftwm/blob/5debd321132ce0b18c5c486319be3b662b291837/display-servers/xlib-display-server/src/lib.rs#L38-L39
  Split `load_colors()` into `load_config()` and `update_colors()`. `reload_config()` now first fully loads the new config and then updates the colors.

- 439f9a673f7b4770964f260a4e98db4e532ddfd3
  Rename `ThemeSettings` to `ThemeConfig` for consistency`
  
- bb993b48dd8e0bd77592818e422c407b2ea90fdd
  Rename `load_config` to `reload_config`, as the config struct is already given with the `new()` function. In both cases the config is relayed to Xwrap's `load_config()`.

## Type of change

# Checklist:

### DIFF
--- a/display-servers/xlib-display-server/src/lib.rs
+++ b/display-servers/xlib-display-server/src/lib.rs
@@ -35,7 +35,8 @@ impl DisplayServer for XlibDisplayServer {
     fn new(config: &impl Config) -> Self {
         let mut wrap = XWrap::new();
 
-        wrap.init(config); // setup events masks
+        wrap.load_config(config);
+        wrap.init(); // setup events masks
 
         let root = wrap.get_default_root();
         let instance = Self {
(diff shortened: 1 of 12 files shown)


barkyq:
I agree that your solution to default_height and default_width is simpler, I am still confused why we duplicate these configuration variables into the state, while my PR #1218 simplifies the state struct by removing the duplicated fields.

barkyq:
Yes I agree that handling config variable in `State` should be done in a more principled way (rather than removing some but not all like I had previously done). I have merged this PR into my local branch (even though it undoes #1218).

mautamu (approved):
Thanks Eskaan, this looks good to me as well!

The one thing I'm a bit confused by is why default_height and default_width are `theme` options, but I don't think it's worth it to change that. Will follow up if I have any hiccups today while using this.

barkyq:
Agreed that `default_height` and `default_theme` should be in `config.ron`

PR #1073, opened 2023-05-31
Implement wayland support using Smithay

blackdragon2447:
# Description

This PR's aim is to give LeftWM support for wayland as an alternative to xorg.

## Todo

- [ ] Fix Stuff
- [ ] Finish LeftWM <=> Smithay communication system:
    - [ ] Implement `SmithayHandle::wait_readable`
    - [ ] Add `send_event` method to `SmithayState`
- [ ] Move the running of Smithay's calloop into a separate thread and return a `SmithayHandle` from `SmithayHandle::new`
- [ ] Implement the sending of `DisplayEvent`s and acting on `DisplayAction`s. 

## Type of change

- [ ] Development change (no change visible to user)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation only update (no change to the factual codebase)
- [ ] This change requires a documentation update

## Updated user documentation:

TBD

# Checklist:

- [ ] Ran `make test-full` locally with no errors or warnings reported
  Note: To fully reproduce CI checks, you will need to run `make test-full-nix`. Usually, this is not neccesary.
- [ ] Manual page has been updated accordingly
- [ ] Wiki pages have been updated accordingly (to perform **after** merge)

Issue #596, opened 2021-12-02
Future Plans.


AethanFoot:
This is just a simple list of future ideas for leftwm (In no particular order):
- [ ] Optimise calls to xlib/minimise calls to xlib.
- [ ] #558 
- [ ] Use the lefthk-core library for keybind handling, and make it feature gated (to allow users to disable built in hotkey daemon and use their own).
- [ ] Change config language (#496).
- [ ] Build up greater configuration and theme settings.
- [ ] Stabilise leftwm-core as a library.
- [ ] Document using leftwm-core as a library.
- [ ] Make leftwm-core display server agnostic (before this we would want the x11 side to be as stable as possible).
- [ ] Add wayland support to leftwm-core then to leftwm.

grtcdr:
Are there any plans to fully comply with the EMWH spec so that `wmctrl` can detect the window manager?

AethanFoot:
I have added a fix for _NET_WM_NAME and added WM_CLASS to #587. I am assuming you were only talking about 
```wmctrl -m```.

grtcdr:
Thank you so much :)

And thank you for the wonderful work you're putting into LeftWM, I really hope it pays off amazingly :heart:

VentGrey:
This may be a shot to the air but, how about "shorthand" commands in `leftwm-state`? This could be very useful for scripts as a whole.

The best example I have is the bspwm client where:

- `bspc query -D --names` throws the desktop names
- `bspc query -D --names -d focused` throws the focused desktop name

I know `leftwm-state` can use liquid templates but the example in the themes folders seems a bit verbose, it would be nice to have for example: `leftwm-state -w0 --tagnames`. 

I say this might be "shot to the air" because I don't know how much users use lemonbar / xmobar as their actual bars but using an [external program](https://github.com/VentGrey/Epitaph/blob/master/camel#L6) seems / feels odd :sweat_smile: .

Any thoughts on this? :thinking:

VuiMuich:
> I say this might be "shot to the air" because I don't know how much users use lemonbar / xmobar as their actual bars but using an [external program](https://github.com/VentGrey/Epitaph/blob/master/camel#L6) seems / feels odd sweat_smile .

Well, this is basically the Unix way, ins't it?

But in general I think, this would make a great additional "external" sub-command to `leftwm` similar to `leftwm-theme`, so it can be added on users desire. Does anybody know how `cargo` is adding externally installed subcommands without hardcoding them in the main binary?

[...]

PR #1221, opened 2024-01-15
Add x11rb backend + Improve backends modularity

Syudagye:
# Description

This PR adds a new backend for interacting with the X11 display server based on the pure rust x11 implementation [`x11rb`].
The main goal here was to see what it takes to add a new backend to LeftWM, and discuse the best approche to do so.
As we have seen with #1073, it is likely we will want to add multiple backends in the future, and making it as convenient as
possible for devs to work with as well as users to switch between them seamlessly is crucial.

Some future backends that could make sense for LeftWM (and that i can think of) would be:
- A full wayland sesion (as stated above)
- A wayland protocol like [river's layout protocol](https://github.com/riverwm/river/blob/master/protocol/river-layout-v3.xml)

This PR was also motivated by #596 goal to move away from xlib.

### This PR

As said before, this PR adds a new backend based around the [`x11rb`] crate.
This backend is similar in the design and the structure as the already existing `xlib` version.

Thanks to `x11rb` beeing pure rust, this backend brings better error handling and reporting, and it removes the dependency on `xlib`/`libX11`.

### Downside

This addition also comes significant downsides:
- Higher compile times, since this technically doubles the backend code. Plus `x11rb` beeing a full x11 implementation, it also takes a long time to compile this specific crate (as stated [here][compile-time])
- More maintaining cost, also due to the backend duplication.

### TODO-List

Non-exhaustive list of things not yet done, don't hesitate to suggest things to add:

- [x] Fully implement LeftWM
- - [x] General window management
- - [x] Window Decorations
- - [x] Fullscreen / Maximize
- [x] Backend switching mechanism. (For now it's mostly hacky)
- [x] Fix CI
- [x] Fix Nix 
- [x] add a `log_level` feature (maybe via an option) to control level of logging

### Bugs

- [x] Improper parsing of `PropertyNotifyEvent`
- [x] Fullscreen does not make the window actually fullscreen.
- [X] Transparent window borders
- [x] Cursor becomes an X when hovering the root window
- [x] EWMH compliance (e.g. polybar not receiving workspaces with windows)
- [x] Resizing floating windows does not work
- [ ] Floating confirmation popups are on top but not focused
- [x] Maximizing not working
- [x] `_NET_SUPPORTING_WM_CHECK` has type missmatch

## Type of change

*TODO*

# Checklist:

[`x11rb`]: https://github.com/psychon/x11rb
[compile-time]: https://github.com/psychon/x11rb/issues/883


leftwm-core/src/models/screen.rs
```diff
@@ -100,6 +100,20 @@ impl From<x11_dl::xrandr::XRRCrtcInfo> for Screen {
     }
 }
 
+impl From<x11rb::protocol::randr::GetCrtcInfoReply> for Screen {
```
Syudagye, line 103: Couldn't it be better to implement `Into<T>` from the backends direclty ? I feel like this would be more natural

leftwm/src/bin/leftwm-worker.rs
```diff
@@ -21,13 +38,35 @@ fn main() {
...
-        manager.register_child_hook();
-        rt.block_on(manager.start_event_loop())
+        match args.get(1) {
+            #[cfg(feature = "xlib")]
+            Some(name) if name == "xlib" => {
+                let manager = Manager::<leftwm::Config, XlibDisplayServer>::new(config);
+
+                manager.register_child_hook();
+                //TODO: Error handling
+                rt.block_on(manager.start_event_loop())
```
Eskaan, line 48: I think you can take this out of the match condition by defining `let manager;` before or returning it from the statement.
That isn't really necessary now, but if you want error handling it would certainly be duplicated code.

Eskaan:
I just want to not that this is a **breaking change** as you now need to pass an argument for leftwm to start.

This also is awesome work! Really looking forward to this.

Syudagye:
> I just want to not that this is a **breaking change** as you now need to pass an argument for leftwm to start.

Yup, for now this is the behaviour, but i though of that as a temporary thing while making everything else work properly.
I think it could be nice to have a default fallback (to `xlib` first and maybe `x11rb` later) if nothing is passed, and maybe later depending on environment variables to differentiate the type of session (x11, wayland, etc)

leftwm/src/bin/leftwm-worker.rs
Syudagye: It's not possible because `manager` has a different type on each branch. One is `Manager<leftwm::Config, XlibDisplayServer>`, the other one is `Manager<leftwm::Config, X11rbDisplayServer>`

Syudagye:
I am currently also trying to remove `leftwm-core`'s dependency on backend specific crates (`x11`, `x11rb`). I think doing so with a trait would be the best idea, but this means basically putting generics with trait bound everywere, and I have no idea if it's considered good practice or not.

I'm thinking of changing the current `WindowHandle` enum to something like this
rust```
pub trait WindowHandle<'de>: Serialize + Deserialize<'de> + Clone + Copy + PartialEq + Eq {}
```

File: leftwm-watchdog/src/main.rs [hunk 3]
```diff
@@ -66,6 +72,24 @@ fn execute_subcommand(subcommand: Subcommand, subcommand_args: SubcommandArgs) -
...
+        .help("Specify the backend to use (`leftwm help backend` for details).");
+
+    #[cfg(feature = "xlib")]
+    {
+        arg = arg.default_value("xlib");
+    }
+    #[cfg(not(feature = "xlib"))]
+    {
+        arg = arg.default_value("x11rb");
+    }
```
Syudagye, line 88: I forgot about the `DEFAULT_BACKEND` i set above

Syudagye:
Well, i think this is mostly complete in the implementation side, there are just a few things left to do:
- Fixing tests (Currently broken, i didn't checked... (I think this is also causing the nix build to fail))
- Documentation for the new `--backend` flag
- All the final steps in the "Checklist" section

Syudagye:
This PR is pretty much ready, it just needs some external review in the case i missed something.

I have still one issue but it is not introduced by this PR: starting without backgroud gives you a black background, which was gray before. I believe this was introduced by #1219, i'll make an issue and investigate later (probably after this is dealt with).

Eskaan:
I just reverted and built #1219 and still have a black background after a reload.
My debug logging tells me `set_background_color()` is called with the result of `get_color("#333333")`.

Eskaan:
@Syudagye Have you disabled picom (or similar) in your up script?
Picom seems to overwrite the default background, after an hour of debugging I realized that `pkill picom` fixes all my issues. 🤦‍♂️

Syudagye:
> @Syudagye Have you disabled picom (or similar) in your up script?

yes, i was testing it in an empty archlinux VM without compositing

Eskaan:
I checked out your branch and did some testing. It could be that at least some of the issues mentioned stem from the failed parse of `PropertyNotifyEvent`s that are printed to the system console:
```
An error occured when processing the event PropertyNotify(PropertyNotifyEvent { .. }): IntConversion[Ok(())] Unable to parse int from a value: out of range integral type conversion attempted
```
Which env var do I need to set for a backtrace (if you want it).

Here a few issues I have atm:
- Fullscreen / Maximize: _NET_WM_STATE_FULLSCREEN is set, but window is not actually fullscreen (same for maximize), probably desync between the wm hints and leftwm-core (missing WindowChange event?).
- If the cursor hovers over the root window (background), it switches to a X-like form. 
- Window borders are transparent (see the pr I just opened to simplify this operation: [#1229](https://github.com/leftwm/leftwm/pull/1229) 65c22e46bc077e5b8f511a1010792dbac9644ae4)
- If a (floating) dialog window pops up for confirmation (in my case for confirming access to a password from keepassxc), it is in front but not focused (the window behind it gets the key inputs).
- The workspaces that have windows on them aren't communicated to polybar (they should be highlighted). I assume it's the same with the urgent tag. They do seem to be communicated correctly with `leftwm-state`, so I think it is something with ewmh compliance.
- Resizing doesn't seem to work (doesn't seem to enter resizing mode?)

Eskaan:
After some more researching it seems like the server can have custom atoms besides the protocol specified ones.

You might want to take a look at [xlib-display-server/src/xatom.rs](https://github.com/leftwm/leftwm/blob/main/display-servers/xlib-display-server/src/xatom.rs#L235-L243) and https://tronche.com/gui/x/xlib/window-information/XInternAtom.html

Syudagye:
> Which env var do I need to set for a backtrace (if you want it).

You can have it with `RUST_BACKTRACE=1`, i'll put it in the wiki afterward.
Maybe it could also be nice to have a little message say how to enable it when it's disabled too in the logs

Syudagye:
Those commits fix the error you reported for `PropertyNotifyEvent` parsing, and it seems to somehow also fix fullscreen (at least on my VM).

I also mirrored b964e9e, but couldn't manage to find a case where there was transparency in my VM, to if this is useless i'll just force remove it later.

Eskaan:
Yes, it also fixed fullscreen for me. Maximized doesn't yet work, but I investigated that and created a commit-patch for you: 
[0001-Fix-ToggleMaximized-and-wm-states.patch.txt](https://github.com/leftwm/leftwm/files/14376788/0001-Fix-ToggleMaximized-and-wm-states.patch.txt)

And here another one correcting my clippy mistake: 
[0002-Fix-my-clippy-mistakes-apply-aa221757-on-x11rb.patch.txt](https://github.com/leftwm/leftwm/files/14376837/0002-Fix-my-clippy-mistakes-apply-aa221757-on-x11rb.patch.txt)

The commit you made is indeed fixing my transparency issue with the borders. I suppose it's some picom setting that makes the borders transparent, idk. I would keep it like this for now for feature parity.

Syudagye:
Thank you for the patches !
I'll apply them when I soon as I can

Syudagye:
>  If a (floating) dialog window pops up for confirmation (in my case for confirming access to a password from keepassxc), it is in front but not focused (the window behind it gets the key inputs).

@Eskaan I am not able to replicate this issue, i have tried with dialogs from multiple application, including keepassxc, and they all seems to work fine for me. Is this still happening on the latest commit ?

Eskaan:
Yes, the focus issue still exists for me. I've made a recording to show the issue.

https://github.com/leftwm/leftwm/assets/76821666/baff827c-b92c-497b-b9be-02975392de1f

I usually just press "Enter" to select the default choice (marked green). The problem is, that the window is not focused and thus the key inputs go to the focused window behind it. If I go out of the window and back in, it gets focused and I can press Enter to allow the key usage.
Here is my current [config.ron.txt](https://github.com/leftwm/leftwm/files/14480944/config.ron.txt), I think it has something to do with `focus_behavior: Sloppy`.

I updated my initial issue comment, the only thing left is this issue. :+1:

Syudagye:
thanks for the details !
i'll patch this when i will be able to

Eskaan:
I am currently investigating my focus issue, and looking at the trace logging, the only difference seems to be with the `_NET_ACTIVE_WINDOW` part.

Faulty log dump (trying to access keypassxc > password prompt):
```
DisplayAction: AddedWindow(WindowHandle(X11rbWindowHandle(16777538)), true, true)
DisplayAction: SetWindowTag(WindowHandle(X11rbWindowHandle(16777538)), Some(2))
DisplayAction: SetWindowOrder([WindowHandle(X11rbWindowHandle(16777538)), WindowHandle(X11rbWindowHandle(14680067)), WindowHandle(X11rbWindowHandle(10485762)), WindowHandle(X11rbWindowHandle(16777222)), WindowHandle(X11rbWindowHandle(27262979)), WindowHandle(X11rbWindowHandle(23068675)), WindowHandle(X11rbWindowHandle(44040195)), WindowHandle(X11rbWindowHandle(20971564))])
DisplayAction: WindowTakeFocus { window: Window { handle: WindowHandle(X11rbWindowHandle(16777538)), transient: Some(WindowHandle(X11rbWindowHandle(16777224))), visible: false, can_resize: true, is_floating: true, must_float: false, floating: Some(Xyhw { x: 614, y: 413, h: -808, w: -1229, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }), never_focus: true, urgent: false, debugging: false, name: Some("KeePassXC -  Access Request"), legacy_name: Some("KeePassXC -  Access Request"), pid: Some(0), type: Dialog, tag: Some(2), border: 0, margin: Margins { top: 0, right: 0, bottom: 0, left: 0 }, margin_multiplier: 1.0, states: [Above, Modal, Modal], requested: Some(Xyhw { x: 1980, y: 403, h: 272, w: 691, minw: 556, maxw: 999999999, minh: 144, maxh: 999999999 }), normal: Xyhw { x: 1366, y: 0, h: 1080, w: 1920, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }, start_loc: None, container_size: None, strut: None, res_name: Some("keepassxc"), res_class: Some("KeePassXC") }, previous_window: Some(Window { handle: WindowHandle(X11rbWindowHandle(23068675)), transient: None, visible: true, can_resize: true, is_floating: false, must_float: false, floating: None, never_focus: false, urgent: false, debugging: false, name: Some("Terminal"), legacy_name: Some("Terminal"), pid: Some(0), type: Normal, tag: Some(2), border: 0, margin: Margins { top: 0, right: 0, bottom: 0, left: 0 }, margin_multiplier: 1.0, states: [], requested: Some(Xyhw { x: 0, y: 0, h: 600, w: 800, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }), normal: Xyhw { x: 1366, y: 20, h: 1060, w: 1920, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }, start_loc: None, container_size: Some(Xyhw { x: 1366, y: 0, h: 1080, w: 1920, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }), strut: None, res_name: Some("Alacritty"), res_class: Some("Alacritty") }) }
PropertyNotify: _NET_WM_STATE : PropertyNotifyEvent { .. }
PropertyNotify: WM_STATE : PropertyNotifyEvent { .. }
PropertyNotify: _NET_WM_USER_TIME : PropertyNotifyEvent { .. }
ClientMessage: 16777538 : "_NET_ACTIVE_WINDOW"
PropertyNotify: _NET_WM_DESKTOP : PropertyNotifyEvent { .. }
PropertyNotify: WM_HINTS : PropertyNotifyEvent { .. }
```
Working log dump (opening a terminal per keybind):
```
PropertyNotify: WM_HINTS : PropertyNotifyEvent { .. }
DisplayAction: AddedWindow(WindowHandle(X11rbWindowHandle(46137347)), false, true)
DisplayAction: SetWindowTag(WindowHandle(X11rbWindowHandle(46137347)), Some(1))
DisplayAction: SetWindowOrder([WindowHandle(X11rbWindowHandle(14680067)), WindowHandle(X11rbWindowHandle(10485762)), WindowHandle(X11rbWindowHandle(16777222)), WindowHandle(X11rbWindowHandle(27262979)), WindowHandle(X11rbWindowHandle(23068675)), WindowHandle(X11rbWindowHandle(44040195)), WindowHandle(X11rbWindowHandle(20971564)), WindowHandle(X11rbWindowHandle(46137347))])
DisplayAction: WindowTakeFocus { window: Window { handle: WindowHandle(X11rbWindowHandle(46137347)), transient: None, visible: false, can_resize: true, is_floating: false, must_float: false, floating: None, never_focus: false, urgent: false, debugging: false, name: Some("Terminal"), legacy_name: Some("Terminal"), pid: Some(0), type: Normal, tag: Some(1), border: 1, margin: Margins { top: 0, right: 0, bottom: 0, left: 0 }, margin_multiplier: 1.0, states: [], requested: Some(Xyhw { x: 0, y: 0, h: 600, w: 800, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }), normal: Xyhw { x: 0, y: 0, h: 0, w: 0, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }, start_loc: None, container_size: None, strut: None, res_name: Some("Alacritty"), res_class: Some("Alacritty") }, previous_window: Some(Window { handle: WindowHandle(X11rbWindowHandle(27262979)), transient: None, visible: true, can_resize: true, is_floating: false, must_float: false, floating: Some(Xyhw { x: 280, y: 173, h: -370, w: -1169, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }), never_focus: false, urgent: false, debugging: false, name: Some(""), legacy_name: Some(""), pid: Some(0), type: Normal, tag: Some(1), border: 1, margin: Margins { top: 0, right: 0, bottom: 0, left: 0 }, margin_multiplier: 1.0, states: [], requested: Some(Xyhw { x: 0, y: 0, h: 600, w: 800, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }), normal: Xyhw { x: 0, y: 176, h: 748, w: 1366, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }, start_loc: None, container_size: Some(Xyhw { x: 0, y: 156, h: 768, w: 1366, minw: -999999999, maxw: 999999999, minh: -999999999, maxh: 999999999 }), strut: None, res_name: Some("Alacritty"), res_class: Some("Alacritty") }) }
PropertyNotify: _NET_WM_STATE : PropertyNotifyEvent { .. }
PropertyNotify: WM_STATE : PropertyNotifyEvent { .. }
PropertyNotify: _NET_WM_DESKTOP : PropertyNotifyEvent { .. }
PropertyNotify: _NET_ACTIVE_WINDOW : PropertyNotifyEvent { .. }
PropertyNotify: WM_HINTS : PropertyNotifyEvent { .. }
```

Eskaan:
Eskaan: *screams in `never_focus: true` set on the window* (I have spent at least an hour on this alone and only got this far)

display-servers/x11rb-display-server/src/xwrap/window.rs
```diff
@@ -0,0 +1,465 @@
...
+            // hint.w = std::cmp::max(xyhw.w, hint.w);
+            // hint.h = std::cmp::max(xyhw.h, hint.h);
+            hint.update_window_floating(&mut w);
+            hint.update(&mut requested);
+        }
+
+        w.requested = Some(requested);
+        w.can_resize = can_resize;
+        if let Some(hint) = wm_hint {
+            w.never_focus = hint.input.unwrap_or(false);
```
Eskaan, line 88: @Syudagye does `hint.input == Some(true)` mean the window actually **wants** to be focused? i.e. did you forget to invert this flag?

Eskaan:
> > I just want to not that this is a **breaking change** as you now need to pass an argument for leftwm to start.
> 
> Yup, for now this is the behaviour, but i though of that as a temporary thing while making everything else work properly. I think it could be nice to have a default fallback (to `xlib` first and maybe `x11rb` later) if nothing is passed, and maybe later depending on environment variables to differentiate the type of session (x11, wayland, etc)

How about a config option instead? (https://github.com/Syudagye/leftwm/pull/3)

Syudagye:
> @Syudagye can you explain (because I can't to myself) why we explicitly leak memory after passing a reference to it into `replace_property_long()` calls on the **xlib**-ds? And especially, why we only do so for some of these calls. (And, why your implementation doesn't need to do that).

It's because it's passing it to `libX11` as a raw pointer in `replace_property_long()`, and it probably (as in i don't know the exact details of this) is better to free it in a C-like style instead of using standard rust dropping. It's just freeing, it's not running the destructor, but it's not leaking. [*doc for `mem::forget`*](https://doc.rust-lang.org/std/mem/fn.forget.html)

Why i don't do this in my implementation is simply because it's not needed, since x11rb does not link to any external libraries. Everything is handled with rust's data model, so freeing memory is done by dropping.


> How about a config option instead? ([Syudagye#3](https://github.com/Syudagye/leftwm/pull/3))

This probably is a better option than messing with the cli like i did :laughing: , I'll look at this, thanks !

Eskaan:
> It's because it's passing it to libX11 as a raw pointer in replace_property_long(), and it probably (as in i don't know the exact details of this) is better to free it in a C-like style instead of using standard rust dropping. It's just freeing, it's not running the destructor, but it's not leaking. [doc for mem::forget](https://doc.rust-lang.org/std/mem/fn.forget.html)

Thanks for your explanation. I didn't notice the difference between freeing and leaking noted there. Could be worth standardizing how we threat the pointer we pass to the function (as I said `mem::forget` is only sometimes used after).

Syudagye:
As mentioned [here](https://github.com/Syudagye/leftwm/pull/3) by @Eskaan, since this PR adds some level of error reporting it could also have further logging controlled by an exterior config, like an option, an env var or even a `leftwm-command` toggle, which could control this at runtime.

Eskaan:
It's not exactly related to this pr, but I don't want to cause conflicts and have time to do it now. There you go: https://github.com/Syudagye/leftwm/pull/4

Syudagye:
Sorry for my recent lack of activity here, I was busy irl.

I think this is ready for merge, I don't see anything left to be done other than updating the wiki.

display-servers/x11rb-display-server/src/xwrap/window.rs
Eskaan: @Syudagye did you overlook this one?

I think this flag needs to be inverted i.e. you're currently trying to set if the window is **allowed to be focused** instead of it being **denied focus** (never_focus)

This table (and code) shows how I think this flag should be used:
| hint.input | never_focus |
|-|-|
| Some(true) | false |
| Some(false)| true |
| None | false |
```suggestion
            w.never_focus = !hint.input.unwrap_or(true);
```

Reference: https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.7

mautamu:
Awesome work on this @Syudagye and @Eskaan! Thank you for implementing this.

mautamu:
Looks like the AUR build is failing at the moment; will need to specify a backend for it to work

Eskaan:
@Syudagye https://github.com/leftwm/leftwm/pull/1221#discussion_r1513520014 So what about this?

Syudagye:
> @Syudagye [#1221 (comment)](https://github.com/leftwm/leftwm/pull/1221#discussion_r1513520014) So what about this?

Oh, I completely missed that :sweat_smile: ...
I'll take a look a bit later tho, i can't right now

0323pin:
https://github.com/leftwm/leftwm/issues/1245

PR #1228, opened 2024-02-08
Fix build without lefthk

pastalian:
# Description

Fix builds without lefthk feature.

revert eaa13b3 and partially revert f78ed16 with minor edit.

## Type of change

# Checklist:

### DIFF
--- a/leftwm/src/bin/lefthk-worker.rs
+++ b/leftwm/src/bin/lefthk-worker.rs
@@ -1,25 +1,28 @@
-#![cfg(feature = "lefthk")]
-
+#[cfg(feature = "lefthk")]
 use lefthk_core::{config::Config, worker::Worker};
+#[cfg(feature = "lefthk")]
 use xdg::BaseDirectories;
 
 fn main() {
-    leftwm::utils::log::setup_logging();
+    #[cfg(feature = "lefthk")]
+    {
+        leftwm::utils::log::setup_logging();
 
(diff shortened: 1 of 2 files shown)


Eskaan:
Oh, no idea why I forgot to check the build without lefthk when doing this.

This solution is fine with me, an alternative would be adding a feature condition in Cargo.toml and removing the feature checks in lefthk-worker entirely. (Idk which solution is better, I'm fine with both).
[fix-lefthk-build-patch.txt](https://github.com/leftwm/leftwm/files/14213318/fix-lefthk-build-patch.txt)

Eskaan (approved):
LGTM (see comment above)

mautamu:
Thank you for this, @pastalian! 

I've been waffling between this and the [fix-lefthk-build-patch.txt](https://github.com/leftwm/leftwm/files/14213318/fix-lefthk-build-patch.txt). I merged this for now, but I think the patch's design pattern would be more likely to prevent similar issues in the future. Let me know your thoughts on if we want to add that as a follow-on @Eskaan.

pastalian:
:o Didn't know about `required-features`. Preventing lefthk-worker from being built seems like the _right_ move. Mine was a quick dirty fix.

mautamu:
Sounds good! In that case will leave https://github.com/leftwm/leftwm/pull/1230 open for a bit and merge.

PR #1229, opened 2024-02-08
Simplify setting the fourth byte of a pixel to 0xff

Eskaan:
# Description

Replace the complicated solution for setting the fourth byte by a simple OR operation.

## Type of change

# Checklist:


mautamu:
Thanks as always, Eskaan!

PR #1239, opened 2024-03-22
implement window urgency update / switch-to on activation

willfindlay:
# Description

This patch introduces a new configuration option, `focus_on_activation` which controls leftwm's behaviour when a window activates (e.g. you click a link and the browser opens a new tab). It has three possible settings:

- `MarkUrgent` (default) marks the window as urgent
- `SwitchTo` switches to the window's tag
- `DoNothing` does nothing, which was the old behaviour prior to this change

This resolves a painful issue I ran into where I wasn't sure which browser was opening a link I clicked when I had browsers open in more than one workspace. The i3 window manager implements a similar configuration setting [1], so I figured it made sense to port it to leftwm.

To make this work properly, I also had to make a few small changes to how we handle window urgency elsewhere, but I think these changes make intuitive sense. In particular, we now clear urgency when a window takes focus, and we only update urgency when a window is not already visibile.

[1]: https://github.com/i3/i3/blob/47cab33aa8b3c8b47e34de6148440a1bc30a2297/src/handlers.c#L432

## Type of change

# Checklist:


willfindlay:
Please review carefully, in particular the `SwitchTo` implementation. I only started ramping up on the codebase yesterday, so forgive me if some obvious way already existed to do what I wanted :)

FWIW I'm running this locally and it works a treat.

willfindlay:
Marking this as draft for now as I found a small bug with clearing window urgency.

leftwm-core/src/handlers/display_event_handler.rs
```diff
@@ -20,6 +20,7 @@ impl<C: Config, SERVER: DisplayServer> Manager<C, SERVER> {
                 .mouse_combo_handler(mod_mask, button, handle, x, y),
 
             DisplayEvent::WindowTakeFocus(handle) => from_window_take_focus(state, handle),
+            DisplayEvent::WindowSwitchToTag(handle) => from_window_switch_to_tag(state, handle),
             DisplayEvent::HandleWindowFocus(handle) => from_handle_window_focus(state, handle),
```
Eskaan, line 24: You can just focus the window using `WindowTakeFocus`, there's no need to add a new event type.

Also, your event technically does not focus the window but only the tag it is on (if there's another window on that tag and the mouse hovers over it, it gets focused instead of the browser).

display-servers/xlib-display-server/src/event_translate.rs
```diff
@@ -85,7 +86,12 @@ fn from_focus_in(x_event: XEvent) -> Option<DisplayEvent> {
         };
         xw.focus(xw.focused_window, never_focus);
     }
-    None
+
+    // Clear window urgency
+    let handle = event.window.into();
+    let mut change = WindowChange::new(handle);
+    change.urgent = Some(false);
+    Some(DisplayEvent::WindowChange(change))
```
Eskaan, line 94: I know there is a bug with unsetting the urgency tag somewhere, this looks like a good workaround for now (haven't been able to test it yet).

I'll probably try to find a better place to put this (leftwm-core instead of the display server) whilst testing, but here should work too.

leftwm-core/src/handlers/focus_handler.rs
```diff
@@ -151,6 +151,17 @@ impl State {
...
+    /// Finds the managed window for a [`leftwm_core::models::window::WindowHandle`].
+    pub fn find_window(&self, handle: &WindowHandle) -> Option<Window> {
+        // Find the handle in our managed windows.
+        let found: &Window = self.windows.iter().find(|w| &w.handle == handle)?;
+        // Docks don't want to get focus. If they do weird things happen. They don't get events...
+        if !found.is_managed() {
+            return None;
+        }
+        Some(found.clone())
+    }
```
Eskaan, line 163: I'm sure we can add lifetimes to this so it returns `Option<&Window>`.

Edit: Or just re-integrate it below when removing the new event.

leftwm-core/src/models/focus_manager.rs
```diff
@@ -33,6 +33,23 @@ impl FocusBehaviour {
...
+    MarkUrgent,
+    /// Switch to the window.
+    SwitchTo,
+}
+
+impl Default for FocusOnActivationBehaviour {
+    fn default() -> Self {
+        Self::MarkUrgent
+    }
+}
```
Eskaan, line 51: How about using the `Default` macro with `#[default]` over the type instead? (You can do it for `FocusBehaviour` above too if you want)

leftwm-core/src/models/window_change.rs
```diff
@@ -59,7 +59,8 @@ impl WindowChange {
         if let Some(urgent) = self.urgent {
             let changed_urgent = window.urgent != urgent;
             changed = changed || changed_urgent;
-            window.urgent = urgent;
+            // If the window is already visible, there is no need to mark it as urgent
+            window.urgent = urgent && !window.visible();
```
Eskaan, line 63: Good catch! That should be at least one bug removed from the urgent-hint implementation. I'll look into it if the workaround mentioned above is still needed after this one.

Eskaan (changes requested):
Thank you for your work! I know the urgent-hint implementation hat some bugs at the moment (you seem to be fixing at least one of them), and this is a config option that I plan on using too.

I think you did some good work for your first contribution here, especially the config part LGTM.
You seem to have missed the `WindowTakeFocus` event type that you could've used. I also have some more minor formatting suggestions in the comments below (everything is open for debate, if you don't agree just explain why).

Edit: If you're willing to wait, I usually submit the changes I want as patches when I am done testing (will finish the review this evening european time).

willfindlay:
@Eskaan thanks for the feedback. I wasn't sure if I was missing something but WindowTakeFocus didn't actually seem to be working properly for me. It was actually the first thing I tried. So maybe there's some bug there or it's related to my other config options.

leftwm-core/src/handlers/focus_handler.rs
willfindlay: Well the original code was also doing a clone, so we will need to rework regardless if we want to use a lifetime instead and avoid the clone.

leftwm-core/src/models/focus_manager.rs
willfindlay: Sure, that probably should have been done in the first place so I'll fix up both.

leftwm-core/src/handlers/display_event_handler.rs
willfindlay: See my other comment about this not working. I'm going to try again to see if I can figure out what's going wrong with `WindowTakeFocus`.

willfindlay: I think I found the bug. I'll push the fix as a separate patch with an explanation.

willfindlay: Found the bug, I've got a partial fix but working on resolving one more issue. I'll push as a separate commit.

willfindlay:
@Eskaan I found and fixed the bug with `WindowTakeFocus` and removed the new event type I added. Thanks for pointing me in the right direction.

leftwm-core/src/handlers/focus_handler.rs
willfindlay: In the mean time I just reverted the change to use the old event as suggested.

Eskaan:
I just started testing, and there seems something wrong with your tag switching logic. After a reload, the wrong tags get focused.

I'd assume the failing CI/tests (`cargo test`) are caused due to the same issue.

willfindlay:
I'll take a look in the morning

Eskaan:
I looked at this for long enough now. I can't figure out a good way to focus a window together with it's tag that covers all cases etc. Maybe a new Event type wasn't so bad after all. I will try this again tomorrow when I had some sleep.

willfindlay:
@Eskaan I've arrived at something that kind of works but imo it's ugly and probably hiding the real problem. Anyway tests seem to pass at least.

willfindlay:
Hmm actually I managed to reproduce some of the weirdness you encountered around tag switching. I'll see if I can add some test cases to capture this and hopefully fix the bug.

housedhorse:
Aha, so at least in my case, I've found an explanation for the strange tag-switching behaviour I saw. When you close a window, the previously focused window becomes focused, even if that's on another tag. The old code wasn't switching tags properly on window focus, but with my fixes here that is no longer the case. So you automatically switch to another tag, which can be quite jarring. I suppose the right thing to do here is probably to switch focus to a sibling in the same tag or root if there is no sibling.

willfindlay:
Aha, so at least in my case, I've found an explanation for the strange tag-switching behaviour I saw. When you close a window, the previously focused window becomes focused, even if that's on another tag. The old code wasn't switching tags properly on window focus, but with my fixes here that is no longer the case. So you automatically switch to another tag, which can be quite jarring. I suppose the right thing to do here is probably to switch focus to a sibling in the same tag or root if there is no sibling.

NicTanghe:
is this dead ?

AnthonyMichaelTDM:
what's blocking this from being implemented? I would find the functionality very usefull

mautamu:
@willfindlay / @Eskaan do we want to take on the fix described below in another PR?:
> I suppose the right thing to do here is probably to switch focus to a sibling in the same tag or root if there is no sibling.

If so, this looks good to me and willing to merge at the end of the week.

Will correct the fmt momentarily.

mautamu:
Going to go ahead and merge this for now; we can tackle focusing for sibling-less scenarios in another PR.

mautamu (approved):
Looks good, thank you @willfindlay for this!

PR #1240, opened 2024-03-24
Add directional focus between windows

### LINKED ISSUE #771: Directional Focus and Window movement
This topic is meant as reference for the regularly upcoming wish for directional movement of focus and windows.

As this seems to be mostly coming from new users who are not used to the simpler workflow yet I believe the priority for this is pretty low. But the effort to implement this is quite a bunch of areas of code to touch.
But we might have a poll eventually to see how many users would like to have this.

Now on some ideas how to approach this:
- Each window gets a `Neighbors` with directions `North`, `South`, `East`, `West` and a `Target: bool` fields 
- a method `FindNeighbors` runs on window creation and every command that changes window positions. Should it ignore floating windows? Should it find neighbors only in the current tag or across workspaces as well?
- if a window has multiple neighbors in one direction the following commands apply to the leftmost or topmost window
- we add commands `FocusNorth`, `FocusSouth`, `FocusEast`, `FocusWest` and `MoveTargetNorth`, `MoveTargetSouth`, `MoveTargetEast`, `MoveTargetWest`
- we add a new `border` type and color `target`
- target moving works as follows: target-mode is entered with the first `MoveTarget{}` command and stays active as long as the modifier(s) are kept pressed, we move the target border around and when all keys are released, we swap the current window and the target window
- `Sloppy` and `ClickTo` targeting modes might be nice too
- we might eventually add a target-mode to unfloat windows

I am a bit concerned that `FindNeighbors` might be somewhat expensive to calculate and therefore might make the affected actions sluggish, but that might just be my lack of knowledge of efficient algorithms for such tasks.

So far my thoughts, might extend that later.

marianomarciello:
# Description

This patch introduces directional focus between windows.
This patch introduces one new command: `FocusWindowAt`.
The `FocusWindowAt` can be called with 4 parameters:
1. `North`: focus window on north
2. `South`: focus window on south
3. `West`: focus window on west
4. `East`: focus window on east

Those commands allow users to use vim key bindings to move the focus on different windows. For example my keybindings for the directional focus are:
`(command: FocusWindowAt, value: "North", modifier: ["modkey"], key: "k"),`
`(command: FocusWindowAt, value: "South", modifier: ["modkey"], key: "j"),`
 `(command: FocusWindowAt, value: "West", modifier: ["modkey"], key: "h"),`
 `(command: FocusWindowAt, value: "East", modifier: ["modkey"], key: "l"),`

Fixes #771 

## Type of change

Should be added a new command (`FocusWindowAt`) in the External Commands section.

# Checklist:

### DIFF
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -113,7 +113,7 @@ fn process_internal<C: Config, SERVER: DisplayServer>(
         Command::FocusWindowUp => move_focus_common_vars!(focus_window_change(state, -1)),
         Command::FocusWindowDown => move_focus_common_vars!(focus_window_change(state, 1)),
         Command::FocusWindowTop { swap } => focus_window_top(state, *swap),
-        Command::FocusWindowAt(param) => focus_window_direction(state, param),
+        Command::FocusWindowAt(param) => focus_window_direction(state, *param),
         Command::FocusWorkspaceNext => focus_workspace_change(state, 1),
         Command::FocusWorkspacePrevious => focus_workspace_change(state, -1),
 
@@ -378,15 +378,14 @@ fn focus_window(state: &mut State, param: &str) -> Option<bool> {
 }
 
 // TODO: add comment
-fn focus_window_direction(state: &mut State, dir: &FocusDirection) -> Option<bool> {
+fn focus_window_direction(state: &mut State, dir: FocusDirection) -> Option<bool> {
     let workspace = state.focus_manager.workspace(&state.workspaces)?.rect();
     let mut rects: Vec<Rect> = vec![];
     let cur_window = state.focus_manager.window(&state.windows)?;
 
-    let mut i = 0;
-    let mut cur = i;
+    let mut cur = 0;
 
-    for x in state.windows.iter().filter(|w| w.visible()) {
+    for (i, x) in state.windows.iter().filter(|w| w.visible()).enumerate() {
         if cur_window.handle.eq(&x.handle) {
             cur = i;
         }
@@ -396,10 +395,9 @@ fn focus_window_direction(state: &mut State, dir: &FocusDirection) -> Option<boo
             x.width() as u32,
             x.height() as u32,
         ));
-        i += 1;
     }
 
-    let next_window = FocusDirection::find_neighbor(&rects, cur, *dir, &workspace);
+    let next_window = FocusDirection::find_neighbor(&rects, cur, dir, &workspace);
 
     match next_window {
         Some(next) => {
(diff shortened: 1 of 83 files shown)


leftwm-core/src/handlers/command_handler.rs
```diff
@@ -375,6 +377,44 @@ fn focus_window(state: &mut State, param: &str) -> Option<bool> {
     }
 }
 
+// TODO: add comment
+fn focus_window_direction(state: &mut State, dir: FocusDirection) -> Option<bool> {
+    let workspace = state.focus_manager.workspace(&state.workspaces)?.rect();
+    let mut rects: Vec<Rect> = vec![];
+    let cur_window = state.focus_manager.window(&state.windows)?;
+
+    let mut cur = 0;
```
Eskaan, line 386: I would prefer an Option here, as you're otherwise asserting the focused window is always on the focused workspace, which leftwm does not guarantee (edge-cases with Sloppy focus and DisplayEvent/DisplayAction buffers may apply).

Eskaan (changes requested):
LGTM (besides the one edge-case note)

leftwm-core/src/handlers/command_handler.rs
marianomarciello: Hi @Eskaan  I was not aware of that. I will make this change right now. Thanks for your time.

Eskaan: I did take a closer look at `handle_window_focus` in focus_handler.rs and the edge-case I thought of does not exist, as we are only calling the display server to move the mouse over the window, which then causes xorg to emit a EnterNotify event, which works it's way back to `focus_window`. My mistake was that I assumed `MoveMouseOver` also focused the window before the workspace.

Eskaan: Also I was refering to line 388 `let mut cur = 0;`, were I would prefer `let mut cur = None` and replacing line 392 with `cur = Some(i);`. You can then unwrap or return in line 402 with a question mark and return if no window was found: `find_neighbor(&rects, cur?, dir, &workspace)`

marianomarciello: Hi @Eskaan thanks for your time and for the clarification.
I have added your suggestions with another commit.

Eskaan:
You can revert [2e4802e](https://github.com/leftwm/leftwm/pull/1240/commits/2e4802e5bee0618e9cd4096c71e31c6c26b5a08b) if you want, it's not making any difference.

Otherwise LGTM.

marianomarciello:
> You can revert [2e4802e](https://github.com/leftwm/leftwm/pull/1240/commits/2e4802e5bee0618e9cd4096c71e31c6c26b5a08b) if you want, it's not making any difference.
> 
> Otherwise LGTM.

HI @Eskaan I prefer it without reverting the commit, it seems more compliant with the rest of the project.

Eskaan:
@mautamu

mautamu:
Thank you for adding this functionality, @marianomarciello! And thanks for reviewing, Eskaan!

matarina:
wiki hadn't updated for usage of the directional focus funciton

PR #1269, opened 2024-06-13
Add distinction between visible and busy workspace

### LINKED ISSUE #787: Add the ability to dinstinct between visible workspace and busy workspace with leftwm-state
I'm trying display the workspace on polybar on multiple screens. I have a bar on screen A and screen B. On screen B polybar I don't want to display what workspace is selected on screen A, I only want to display if there in a window or not on every workspace.
 To do so, I though that I could use the `busy` property from leftwm-state, but it also returns `busy == true` when a workspace is visible on any screen. Is it possible to have `busy == false` when a workspace is visible but there is no window in it?
We already have the `visible` property to check if a workspace is visible. 

```
        {
          "name": "6",
          "index": 5,
          "mine": false,
          "visible": true,
          "focused": false,
          "busy": true
        },
```

fcolona:
# Description

This patch updates the behavior of the ```busy``` attribute. Previously, it would be set to ```true``` even if there were no windows in a workspace. With this patch, a workspace tag is considered ```busy``` only if the tag is both visible in that workspace and at least one window (excluding dock and desktop windows) has that tag.

Closes #787 

## Type of change

# Checklist:

### DIFF
--- a/leftwm-core/src/models/dto.rs
+++ b/leftwm-core/src/models/dto.rs
@@ -1,5 +1,5 @@
-use crate::state::State;
 use crate::models::TagId;
+use crate::state::State;
 use serde::{Deserialize, Serialize};
 
 use super::Handle;
@@ -135,7 +135,12 @@ impl<H: Handle> From<&State<H>> for ManagerState {
             .tags
             .all()
             .iter()
-            .filter(|tag| state.windows.iter().any(|w| w.has_tag(&tag.id) && w.is_managed()))
+            .filter(|tag| {
+                state
+                    .windows
+                    .iter()
+                    .any(|w| w.has_tag(&tag.id) && w.is_managed())
+            })
             .map(|t| t.id)
             .collect();
 
--- a/leftwm-core/src/models/dto.rs
+++ b/leftwm-core/src/models/dto.rs
@@ -21,7 +21,7 @@ pub struct ManagerState {
     pub desktop_names: Vec<String>,
     pub viewports: Vec<Viewport>,
     pub active_desktop: Vec<String>,
-    pub working_tags: Vec<(String, usize)>,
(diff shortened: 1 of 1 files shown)


FelipeColona:
While writing this patch, a few questions arose: 
I'm unsure about whether or not a tag that has a window, but is not visible in any workspace, should be flagged as ```busy```. If so, should it be ```true``` only in the primary workspace or should it flagged in all of them? Besides that, should the ```working_tags``` be a struct or is it enough for it to be a tuple?

grndplane:
This may address a problem I have been having. I have four tags, and when I close a program on the last(4th) tag it still shows as active in polybar ewmh module.

mautamu:
Apologies about the delay, @fcolona!

> whether or not a tag that has a window, but is not visible in any workspace, should be flagged as `busy`

I think we should flag these as `busy`, but am neither here nor there on this.

>  should the working_tags field be a struct or is it enough for it to be a tuple?
I think that might be useful for the future.

fcolona:
Hi @mautamu, no problems with the delay!

> I think we should flag these as `busy`, but am neither here nor there on this.
Okay, I'll start working on that as soon as I have some free time.

> I think that might be useful for the future.
Just to clarify, are you saying that it should be a struct?

mautamu:
> Just to clarify, are you saying that it should be a struct?

Yeah, that works; thank you!

fcolona:
Hey @mautamu, I forgot to tell you that the PR was ready for review, sorry about that. Could you take a look at it when you have some time? Thank you!

mautamu (approved):
Hey @fcolona! Thanks for the bump and for doing this!!

This looks good to me, going to go ahead and merge if it fails due to the same error that caused #1288 to initially fail, should be fixed when we merge #1288 later today.

PR #1274, opened 2024-08-14
Fix hiding window behaviour

### LINKED ISSUE #1100: Black screen for XComposite window capture on OBS
### Describe the bug

When capturing a window with OBS using the XComposite capture source, if the window is not on the same tag as OBS, is will result in a black output.

Proper window capture, even if the window is not on the currently focused tag.

1) Spawn a window that you want to capture and OBS
2) Add the XComposite source to your source list in obs, and select the desired window.
3) See the result when OBS is on the same tag as your captured window, and when it's not

_No response_

### OS / Distro

Linux archbtw 6.3.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 10 Jun 2023 00:35:35 +0000 x86_64 GNU/Linux

_No response_

### LeftWM Check

```text
:: LeftWM version: 0.4.2
:: LeftWM git hash: 8ad3204-modified
    -> journald-log OK
    -> lefthk OK
    -> Binaries OK
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK
```

Syudagye:
# Description

Fixes #1100 by moving the window out of the screen rather than using `Map`/`Unmap` requests, which causes the window to appear black when captured when hidden

## Type of change

# Checklist:

<sub>(clippy seems mad at some previous change....)</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - The currently supported MSRV is 1.74.0
+- Added `window_hiding_strategy` config option (via #1274 by @Syudagye)
 
 ### Fixed
+
+- Hidden window are not able to be captured (fixes #1100 via #1274 by @Syudagye)
+- Numlock prevents moving and resizing for floating windows (via #1276 by @Syudagye)
 
 ## [0.5.0]
 
--- a/display-servers/x11rb-display-server/src/xwrap.rs
+++ b/display-servers/x11rb-display-server/src/xwrap.rs
@@ -1,6 +1,7 @@
 use std::{io::IoSlice, os::fd::AsRawFd, sync::Arc, time::Duration};
 
 use leftwm_core::{
+    config::WindowHidingStrategy,
     models::{FocusBehaviour, WindowHandle},
     utils::{self, modmask_lookup::ModMask},
     Config, Mode, Window,
@@ -74,6 +75,7 @@ pub(crate) struct XWrap {
     pub focus_behaviour: FocusBehaviour,
     pub mouse_key_mask: ModMask,
     pub mode_origin: (i32, i32),
+    pub window_hiding_strategy: WindowHidingStrategy,
 
     #[allow(unused)]
     task_guard: oneshot::Receiver<()>,
@@ -164,6 +166,7 @@ impl XWrap {
             focus_behaviour: FocusBehaviour::Sloppy,
             mouse_key_mask: ModMask::Zero,
             mode_origin: (0, 0),
+            window_hiding_strategy: Default::default(),
 
             task_guard,
             task_notify,
@@ -194,6 +197,7 @@ impl XWrap {
             active: self.get_color(&config.focused_border_color())?,
             background: self.get_color(&config.background_color())?,
         };
+        self.window_hiding_strategy = config.window_hiding_strategy();
         Ok(())
     }
 
--- a/display-servers/x11rb-display-server/src/xwrap/window.rs
+++ b/display-servers/x11rb-display-server/src/xwrap/window.rs
@@ -209,12 +209,12 @@ impl XWrap {
         Ok(())
     }
 
-    /// "hides" a window my moving it out of view.
+    /// "hides" a window by moving it out of view.
     /// see https://github.com/leftwm/leftwm/issues/1100
     pub fn toggle_window_visibility(&self, window: xproto::Window, visible: bool) -> Result<()> {
         if visible {
             // NOTE: The window does not need to be moved here, if it's beeing made visible it's
-            // going to be naturally tiled of placed floating where it should
+            // going to be naturally tiled or placed floating where it should
 
             // Set WM_STATE to normal state.
             self.set_wm_state(window, WMStateWindowState::Normal)?;
--- a/display-servers/x11rb-display-server/src/xwrap/window.rs
+++ b/display-servers/x11rb-display-server/src/xwrap/window.rs
@@ -9,7 +9,7 @@ use x11rb::{protocol::xproto, x11_utils::Serialize};
 use crate::xatom::WMStateWindowState;
 use crate::{error::Result, X11rbWindowHandle};
 
-use super::{root_event_mask, XWrap};
+use super::XWrap;
 
 impl XWrap {
     /// Sets up a window before we manage it.
@@ -209,20 +209,15 @@ impl XWrap {
         Ok(())
     }
 
-    /// Maps and unmaps a window depending on it is visible.
+    /// "hides" a window my moving it out of view.
+    /// see https://github.com/leftwm/leftwm/issues/1100
     pub fn toggle_window_visibility(&self, window: xproto::Window, visible: bool) -> Result<()> {
-        // We don't want to receive this map or unmap event.
-        let mask_off = root_event_mask().remove(xproto::EventMask::SUBSTRUCTURE_NOTIFY);
-        let mut attrs = xproto::ChangeWindowAttributesAux {
-            event_mask: Some(mask_off),
-            ..Default::default()
-        };
-        xproto::change_window_attributes(&self.conn, self.root, &attrs)?;
         if visible {
+            // NOTE: The window does not need to be moved here, if it's beeing made visible it's
+            // going to be naturally tiled of placed floating where it should
+
             // Set WM_STATE to normal state.
             self.set_wm_state(window, WMStateWindowState::Normal)?;
-            // Make sure the window is mapped.
-            xproto::map_window(&self.conn, window)?;
             // Regrab the mouse clicks but ignore `dock` windows as some don't handle click events put on them
             if self.focus_behaviour.is_clickto()
                 && self.get_window_type(window)? != WindowType::Dock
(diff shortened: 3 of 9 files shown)


mautamu (approved):
This looks good from a code perspective; test driving and will plan to merge tomorrow if no issues. Thanks again as always @Syudagye!

mautamu:
Hmm, looks like depending on how you enter the state it might not correctly remap the windows. E.g., if I use d9c7cbc and then enter [638ab3e](https://github.com/leftwm/leftwm/pull/1274/commits/638ab3e3e80074fd3cbc7067282ff1e5a23d7db9) with features `xlib,lefthk,journald-log`, it leaves multiple windows laid overtop one another for me :thinking:, might just be an issue on my machine though?

Syudagye:
Ah, didn't catch this one apparently !
It seems to happening on the `xlib` backend, `x11rb` seems fine tho

DrWarpMan:
It works for OBS, thanks for that.

But, you still can not screenshare inactive windows (windows in different tags) in applications like Discord, Telegram or any browser and probably many others. They are not even listed as an option unless they are in the same tag as the application that is doing the screenshare. And if they are listed as an option, as soon as you put them in an inactive tag, it stops rendering them.

Syudagye:
Discord seems to not use the same mechanism as obs, it is dependent on the [WM_STATE_WINDOW_STATE atom](https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#changing_window_state).
Either it does not want to capture windows that are iconic (minimized), or it is unable to.

By removing [this line](https://github.com/Syudagye/leftwm/blob/4008615050d944ae16165b87ccf732e1d37aabb6/display-servers/x11rb-display-server/src/xwrap/window.rs#L240) (and [this one](https://github.com/Syudagye/leftwm/blob/4008615050d944ae16165b87ccf732e1d37aabb6/display-servers/xlib-display-server/src/xwrap/window.rs#L250) for parity) it works, but then windows are not notified they are not shown anymore.
In my opinion this is not a big deal, since for example obs is able to capture hidden windows, than means they are rendered anyway, so removing this shouldn't cause higher resources usage.

But, should this have a config option or window rule option available  to fallback to correclty setting the iconic state, or even the `map`/`unmap` behaviour ?

hertg:
> But, should this have a config option or window rule option available to fallback to correclty setting the iconic state, or even the `map`/`unmap` behaviour ?

Just reading through the Issue + PR, I also wanted to suggest that. I think it would be a good idea to have a flag in the config to be able to revert to the current behavior. If you think the change in resource usage is negligible, I don't mind if we change the default so that there's less friction for screen sharing.

I occasionally have a similar issue at work, when I try to share a window in Google Meet, I have to temporarily move my browser window to the same tag as my "to be shared" window so that I can select it. I assume this would also fix that?

Syudagye:
> I occasionally have a similar issue at work, when I try to share a window in Google Meet, I have to temporarily move my browser window to the same tag as my "to be shared" window so that I can select it. I assume this would also fix that?

yes, this is why, it should be fixed with this pr

DrWarpMan:
> I occasionally have a similar issue at work, when I try to share a window in Google Meet, I have to temporarily move my browser window to the same tag as my "to be shared" window so that I can select it. I assume this would also fix that?

Well, that's the thing, it won't completely fix that.

I tested right now, and it seems I am able to screenshare a specific window in Brave, but first I need to put it in the same tag. Which is really annoying. But afterwards, when I move it to another tag, it seems to work as expected and it doesn't stop rendering.

Tested on both Telegram and Brave, on Discord, this is still buggy.

In KDE, I didn't have these issues at all.

DrWarpMan:
Did a quick test in Discord and browser, both seem to work, good job 😎

hertg (commented):
Everything looks great, only things I saw were some typo's in comments :)
I am fine with changing the default as you did here. But maybe it would be good to actually measure the performance impact this has, especially for applications that might use more resources. I know that some games have settings like `mute sound when minimized`, or `pause when minimized`, etc., I assume this change could result in unexpected behavior there? (Unfortunately, I can't name the games with those settings, I just know that I've seen those settings in the past).

I'm not too cautious about changing this default in the master branch, I think we'll get some valuable feedback before the next release about what to do best.

hertg:
Oh, one small thing. Could you already update the `CHANGELOG.md` and include this change there. While you're at it, you could also add a mention of the fix you did with #1276. It will be easier to do the next release if we keep the changelog up-to-date in the main branch.

Syudagye:
> But maybe it would be good to actually measure the performance impact this has, especially for applications that might use more resources. I know that some games have settings like `mute sound when minimized`, or `pause when minimized`, etc., I assume this change could result in unexpected behavior there? (Unfortunately, I can't name the games with those settings, I just know that I've seen those settings in the past).

I have some, but i don't really know how i can properly monitor that. I'll ask on the discord if some people would be able to help for that.

hertg (approved):

Syudagye:
I did a little investigation on my side on this to measure GPU usage.

## Setup

I used `radeontop` for this and too the `graphics pipe` metric to have a quick idea of the impact.

My setup (if relevant): 
- cpu: Ryzen 5 2600
- gpu: Radeon RX 6600
- ram 16Go

I ran tests on the three settings of `window_hiding_strategy` by measuring gpu load on leftwm startup at idle, launching Apex Legends and going into the firing range (to have things to render), and then osu! (the open source "lazer" project) with uncapped fps on the title screen.
For both Apex and osu, i tested the load in 3 situations: when focused, shown but not focused and hidden.

## Results

||Unmap|MoveMinimize|MoveOnly|
|:-|-:|-:|-:|
|IDLE|0%|0%|0%|
|Apex Legends - in focus|65% to 75%|65% to 75%|65% to 75%|
|Apex Legends - not in focus|20% to 25%|20% to 25%|20% to 25%|
|Apex Legends - hidden|20% to 25%|20% to 25%|20% to 25%|
|osu! - in focus|55% - 60%|40% - 70%|40% - 60%|
|osu! - not in focus|5%|5%|5%|
|osu! - hidden|0%|3%|3% - 5%|

## Observations

There could in fact be a little performance hit with not using `Map`/`Unmap`, and also not minimizing, since the window will still render, and in the case of not minimizing, xorg is probably also doing things with the buffer (see osu! when hidden).

But, we can see that these application are in fact reducing their GPU usage when unfocused, and not necessarely when unmapped (see Apex).

hertg:
Amazing, thanks so much for taking the time to test this. I see... the results are - as expected - highly dependent on how the application in question handles being unfocused and minimized. I'd imagine that whether something is running through Wine or not will also influence things. 

Taking everything into consideration, what would be your suggestion to do with the default? I am torn, because on one hand having less friction for screen sharing sure seems nice, and it is something that I was personally affected by. If I could make it work by changing a setting, this would be fine for me, but people having the issue need first to figure out that such a setting exists. Setting the default to `MoveOnly` would resolve this friction and the requirement to know about this setting. But we also have a lot of users that chose LeftWM for its small resource footprint (people usually choose window managers, especially a Rusty one for the purpose of minimalism and/or resource footprint reduction), changing the default to the more resource intensive option would not be great in that case. Also thinking about laptop users, I don't think many would sacrifice some battery life for a QoL improvement when it comes to screen sharing.

Just thinking out loud here, but after those thoughts, I tend to lean more towards leaving the default at the current behavior. Any other opinions? @mautamu @DrWarpMan @kyleguarco

DrWarpMan:
I'm leaning towards keeping the `MoveOnly` default. 

In either case, I suggest placing information about this setting in a prominent section of the documentation to ensure that users are aware of this.

Syudagye:
> Taking everything into consideration, what would be your suggestion to do with the default?

I would suggest to use `MoveOnly` by default, since it is expected to be able to share a window without issues. Plus, application that will have a big effect on battery life (game, 3d software, etc) are probably going to be used (in focus) most of the time they're opened. Moreover, dwm also has this hiding behaviour, and a lot of users use it anyway for its low footprint.

Plus, if the usecase really is there, we could add a window rule for this setting, letting people set the default to `Unmap`, but still being able to share specific windows

mautamu:
> Any other opinions?

On one hand, I only screenshare once in a blue moon, so for me specifically, it would make sense to use `Unmap`. On the other hand, I agree that LeftWM should just work "out of the box" on things like this and would be fine with `MoveOnly` as the default.

> Plus, if the usecase really is there, we could add a window rule for this setting, letting people set the default to Unmap, but still being able to share specific windows

I like this idea also, but totally understand if that's not something we want to add in this PR!

hertg:
Thanks for your feedback, I'm generally fine with changing the default to MoveOnly. I'd say we change it in the main branch now, and we can discuss it again if we get any resistance from the community.  To me, the performance aspect isn't that big of a deal in itself, but the potential increase in power consumption isn't great. We should prominently note that in the wiki description for this new setting. But without concrete measurement data for battery drain, we can only guess the impact. I assume it won't be too big, but might be noticeable to people who often work "unplugged" with more graphic intensive applications.

Btw. also really like the idea of setting this in window rules, @Syudagye can you open a new issue to track this? We can add that later IMO.

I'll merge this evening (Europe+02:00) if no one else objects until then.

Syudagye:
> Btw. also really like the idea of setting this in window ***@***.*** can you open a new issue to track this? We can add that later IMO.

Yes no problems, i'll tackle that in the next few days.

> We should prominently note that in the wiki description for this new setting.

Definitly yes, but could it be nice to have a build time warning, like for lefthk migration ? Or is it not needed since it's not a breaking change ?

PR #1282, opened 2024-09-06
Lazily call randr on windows hiding

Lieunoir:
# Description

This makes some calls for the screens context when windows are hidden lazy. In particular this avoids some randr calls, which on some setups (when using nvidia prime for example) requires the dGPU to be awake: when it isn't, there is a noticeable latency. This avoids these lags as well as the unnecessary gpu wake up calls.

There may be a more in depth way to avoid these calls, but this removed the (quite frequently recurring) issue for me.

## Type of change

# Checklist:


mautamu:
Thank you again Lieunoir!!

Issue #1283, opened 2024-09-14
Window hiding strategy issues (MoveOnly, MoveMinimize)


DrWarpMan:
### Describe the bug

I have encountered first issues with the two new window hiding strategy options (https://github.com/leftwm/leftwm/pull/1274).
So far, I have identified only two applications — games that are affected, but I am sure there are more.

When using the `MoveOnly` or `MoveMinimize` window hiding strategies, I cannot switch tabs properly.
In the provided videos, you can see **when I try to switch to a different tag while the game is running**, the **window manager** goes crazy and **starts rapidly switching between the two tags**.

**Additional notes:**
- The game also often temporarily freezes when you try to switch back to it, sometimes even permanently, necessitating a process kill. _This happens mainly in Valheim._
- _Only in Counter-Strike 1.6_, using windowed mode seems to prevent the bug from happening.

https://github.com/user-attachments/assets/2ec5ab2a-47c6-45aa-99c3-eb862b440cdf

https://github.com/user-attachments/assets/bc8b21b8-fa3c-4848-8438-dfdf7753f2f0

### Steps to reproduce (optional)

1. Start Counter-Strike 1.6 **OR**  Valheim
2. Join game (server or a local map/world)
3. Switch tag to a different one than the game itself

### OS / Distro

Linux 6.10.9-arch1-2

### LeftWM Check

```text
:: LeftWM version: 0.5.1
:: LeftWM git hash: 0.5.1-40-g36609e0
:: Enabled features:  journald-log lefthk x11rb xlib
:: Checking feature dependencies . . .
    -> journald-log OK
    -> lefthk OK
:: Checking for leftwm binaries . . .
    -> Binaries OK 
:: Loading configuration . . .
    -> Configuration loaded OK 
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK 
:: Checking theme . . .
    -> Theme OK
```

Syudagye:
Just tested with both games you provided, using the same version of leftwm (36609e0fcdfd131834b2ec9f1913211da6138ba1) and `window_hiding_strategy: MoveOnly`, but i didn't observe these behaviors:

https://github.com/user-attachments/assets/88f2fda3-c60f-4cb8-8404-690c805abceb


https://github.com/user-attachments/assets/d590216a-8e32-4ce0-b0bd-08b70e1640fa

[...]

PR #1284, opened 2024-09-15
Implement window rule for `hiding_strategy`

Syudagye:
# Description

In addition of #1274, this PR adds a new window rule parameter in case some window do not behave correctly under `MoveMinimize` or `MoveOnly` hiding behaviours, or if you want to be able to share specific windows but still have `Unmap` for other windows.

## Type of change

- `hiding_strategy` When specified, will force the window to hide in a certain way. See `window_hiding_strategy` for values.

# Checklist:

- [ ] Ran `make test` locally with no errors or warnings reported


Syudagye:
an example situation where this could be useful: #1283

mautamu (approved):
Thank you very much for this PR @Syudagye!

This looks good to me, will go ahead and merge it since not seeing any objections.

PR #1297, opened 2024-11-24
Commented template.liquid

thealio:
I' ve added comments to the template.liquid file.
Now is better usable
I've also made some additions , with showing the leftwm layout and the focused window title.


thealio:
Of course , this can be used also for other bars , such as polybar (I've also tested on it)

mautamu (approved):
Hello, fellow adventurer!

Thanks for submitting this documentation addition, Thealio! This looks good to me.

PR #1306, opened 2025-02-03
Update Github Action for Linting, Fix clippy issues

hertg:
# Description

- Update Github Action to use same clippy arguments as when running `make test` locally (#1233)
- Fix clippy warnings

## Type of change

- [x] Development change (no change visible to user)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation only update (no change to the factual codebase)
- [ ] This change requires a documentation update

## Updated user documentation:

Please insert user documentation that should be updated (as in the wiki).

See [CONTRIBUTING.md User Documentation section](../CONTRIBUTING.md#user-documentation) for further details.

**Note: Manual page changes must be performed in a commit, not in this PR section.**

# Checklist:

- [x] Ran `make test` locally with no errors or warnings reported
  Note: To fully reproduce CI checks, you will need to run `make test-full-nix`. Usually, this is not necessary.
- [ ] Manual page has been updated accordingly
- [ ] Wiki pages have been updated accordingly (to perform **after** merge)

PR #1305, opened 2025-02-02
fix sloppy focus after window closes

fransklaver:
After a window closes, the focus is updated to another window. For sloppy focus this happens by queueing FocusWindowUnderCursor. However, when this action is being executed, there is a check for change in focus which happens _after_ the window being destroyed has been removed from the window history. This in turn causes us to think there is nothing to do and we exit early.

Push a None focus onto the window history to force an update. After all, we don't actually have anything in focus at the moment.

fixes #1295

## Type of change

- [ ] Development change (no change visible to user)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation only update (no change to the factual codebase)
- [ ] This change requires a documentation update

# Checklist:

Note: make test already fails for me on main. ~~I'll make a separate PR for that.~~ No actions required for the other two items.

- [ ] Ran `make test` locally with no errors or warnings reported
  Note: To fully reproduce CI checks, you will need to run `make test-full-nix`. Usually, this is not necessary.
- [ ] Manual page has been updated accordingly
- [ ] Wiki pages have been updated accordingly (to perform **after** merge)

PR #1308, opened 2025-02-04
chore: version to 0.5.4, MSRV to 1.83.0

mautamu:
# Description
This update bumps leftwm to 0.5.4 to push #1305 and #1306 to packaging as requested by @AnthonyMichaelTDM in #1305.

The liquid library has bumped their latest version with a corresponding increase in MSRV to 1.83.0. While ideally we would not bump the MSRV by 9 minor increments at once (we previously have followed a cadence of around 5 minor increments), there have been a number of [major changes](https://github.com/cobalt-org/liquid-rust/commits/master/) to liquid since then which would be enabled by bumping the MSRV. 

## Type  of change

CHANGELOG.md has been updated

# Checklist:

- [ ] Manual page has been updated accordingly (N/A)
- [ ] Wiki pages have been updated accordingly (to perform **after** merge) (N/A)


VuiMuich:
I think bumping the MSRV by so much is ok, since previously the bumps were closer in time as well, if memory serves correctly.
And it allows for some very nice lints to improve code quality.

On thing though: I would suggest to also bump the rust version in all our sub-crates.

mautamu:
Good catch @VuiMuich!

Went ahead and updated the MSRV to be consistent.

VuiMuich:
Lgtm 

Thanks for the chores @mautamu!

PR #1312, opened 2025-02-20
xlib, x11rb: fix ClickTo focus after switching tags

fransklaver:
When switching tags back and forth in `ClickTo` focus, we can't click to focus any windows. Only after the windows have come into focus through other means, are we able to click on them.

So when switching tags, windows are hidden using the `OnlyMove` strategy. In this case these windows are left in the `Normal` window state, even if they aren't visible. When we switch back to the original tag, we then don't call `toggle_window_visibility`, because we think there's nothing to do. This in turn means we don't call `grab_mouse_events` on these windows, so we're never receiving any click events that enable us to click to focus.

So when marking windows invisible, always make sure we take them out of the 'normal' state.

## Type of change

# Checklist:

### DIFF
--- a/display-servers/x11rb-display-server/src/xwrap/window.rs
+++ b/display-servers/x11rb-display-server/src/xwrap/window.rs
@@ -279,12 +279,16 @@ impl XWrap {
                 }
             }
 
-            // Set WM_STATE to iconic state.
-            if hiding_strategy == WindowHidingStrategy::Unmap
+            // make sure to always change the window state, so we don't forget
+            // to grab mouse button events when we show the window again.
+            let new_state = if hiding_strategy == WindowHidingStrategy::Unmap
                 || hiding_strategy == WindowHidingStrategy::MoveMinimize
             {
-                self.set_wm_state(window, WMStateWindowState::Iconic)?;
-            }
+                WMStateWindowState::Iconic
+            } else {
+                WMStateWindowState::Withdrawn
+            };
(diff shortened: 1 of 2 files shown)


mautamu (approved):
Using the below configuration for testing, this appears to resolve the ClickTo issue. Code changes look superb also.

Thank you very much for another great contribution @fransklaver!

```
#![enable(implicit_some)]
(
    modkey: "Mod4",
    mousekey: "Mod4",
    workspaces: [],
    tags: [], // omitted for brevity
    max_window_width: None,
    layouts: [], // omitted for brevity
    layout_mode: Tag,
    insert_behavior: Bottom,
    scratchpad: [(name: "Alacritty", value: "alacritty", x: 0.1, y: 0.1, height: 0.8, width: 0.8)],
    //window_rules: [(window_class : "Alacritty", spawn_floating : true),],
    window_rules: [( window_class: "Alert", spawn_floating: true ), ( window_class: "gcr-prompter", spawn_floating: true ), ( window_class: "Toolkit", spawn_floating: true )],
    disable_current_tag_swap: false,
    disable_tile_drag: false,
    create_follows_cursor: false,
    disable_window_snap: true,
    focus_behaviour: ClickTo,
    focus_new_windows: false,
    disable_cursor_reposition_on_resize: true,
    sloppy_mouse_follows_focus: true,
    state_path: None,
    keybind: [] // omitted for brevity
)
```

PR #1319, opened 2025-03-02
x11rb: getters: fix getting the output name

fransklaver:
Getting the output name always fails, as indicated by the `FIX` note, leading to leftwm-state to always say the current output is `"output_name"`.

This happens because we're assuming we're still doing ffi here and as such are trying to read a null-terminated string from some bytes. In fact the data isn't null-terminated, at least not here. And since there is no null-termination, the conversion always fails.

To get this to work, we can simply convert the data to `str` and drop the whole `CStr` use.

## Type of change

# Checklist:

### DIFF
--- a/display-servers/x11rb-display-server/src/xwrap/getters.rs
+++ b/display-servers/x11rb-display-server/src/xwrap/getters.rs
@@ -1,4 +1,4 @@
-use std::{backtrace::Backtrace, ffi::CStr};
+use std::backtrace::Backtrace;
 
 use leftwm_core::models::{
     BBox, DockArea, Screen, WindowHandle, WindowState, WindowType, XyhwChange,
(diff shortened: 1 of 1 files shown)


fransklaver:
Alright, seems like there's no impact there. This can be merged as far as I'm concerned.

mautamu (approved):
Thank you again for all of these wonderful fixes @fransklaver!

:warning: This may require a hard reload rather than a soft reload :warning: 

Code appears to be fine, test path:
1. Build with 902973d, `cargo install --features journald-log,x11rb,lefthk --no-default-features --path leftwm/`
2. Run `leftwm-state | grep output_name` and observe highlighted `output_name`
3. Rebuild with 5725a9d, `cargo install --features journald-log,x11rb,lefthk --no-default-features --path leftwm/`
4. No longer observe issue.

PR #1330, opened 2025-05-18
fix: block focus change when current window is fullscreen

### LINKED ISSUE #1329: fix: block focus changes when a window is fullscreen
### Describe the enhancement

When a window in the current tag is currently fullscreen, we should refuse to change focus as the fullscreened window should always retain focus

mentioned in the discord, thought I'd open an issue for it

Would be nice to have

VuiMuich:
# Description

When the currently focused window is fullscreen it is possible to move focus away from it. This can lead to a chaotic situation, especially, when one tries to un-fullscreen the window as this results in fullscreening the next focused window in the background.

Marking as draft because:
- [ ] Can't personally test, right now
- [ ] Unsure if this little fix is sufficient, or if we need to check in other places as well
- [ ] Want to write tests for this szenario (yes I know that this is kinda backwards :joy:)

Fixes #1329

## Type of change

# Checklist:

### DIFF
--- a/leftwm-core/src/handlers/command_handler.rs
+++ b/leftwm-core/src/handlers/command_handler.rs
@@ -805,10 +805,15 @@ fn focus_window_change<H: Handle>(
     val: i32,
 ) -> Option<bool> {
     let is_handle = |x: &Window<H>| -> bool { x.handle == handle };
+    let is_fullscreen = helpers::relative_find(&to_reorder, is_handle, 0, true)?;
+    if is_fullscreen.is_fullscreen() {
+        // state.windows.append(&mut to_reorder);
+        state.handle_window_focus(&is_fullscreen.handle);
+        return Some(!is_fullscreen.is_fullscreen());
+    }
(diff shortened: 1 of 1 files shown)


moqodow:
I think blocking changing focus on the current tag if one window is fullscreen is a good change.

I have a question, would that also block switching to another tag. For me that would be an undesired outcome.

If that is the indent please add a setting to change the behavior. Thanks

VuiMuich:
If memory serves well, the mechanic is, that `FocusUp` and `FocusDown` only affect windows of the same tag, moving focus to a different tag should be an explicit action by a different command. But to be fair I basically never used LeftWM in a multimonitor setup (with some extremely rare and very short exceptions), so I am not perfectly sure I remember this correctly.

Under this assumption I chose the `focus_change_window` function to not potentially affect any movements that might lead to a different workspace and/or tag.
I am not sure either if the issue also could occur with directional focus, but as I understand that mechanics, it checks for visible neighbor windows, which should return zero results for a fullscreen window.

NicTanghe:
what about Dialogue windows though ?

PR #1337, opened 2025-07-01
fix: include path in warning when path does not exists

hosfeld:
# Description

This small change enhances the warning message when a submitted path does not exist by including the actual path value.

## Type of change


mautamu:
Thanks for the patch @hosfeld! Change looks good to me.

Issue #1307, opened 2025-02-03
fix ignored ref_option clippy issue


hertg:
I ignored a clippy issue in #1306 as I wasn't able to quickly fix it. The line here should be removed:

https://github.com/leftwm/leftwm/blob/fa91750c00bcb5b51630a6850ce74f545bdc5444/leftwm/src/config.rs#L735

---

This will require changes in the serde serialization here
https://github.com/leftwm/leftwm/blob/fa91750c00bcb5b51630a6850ce74f545bdc5444/leftwm/src/config.rs#L79

and here
https://github.com/leftwm/leftwm/blob/fa91750c00bcb5b51630a6850ce74f545bdc5444/leftwm/src/config.rs#L86

PR #1341, opened 2025-07-29
config: Add SerializableRegex to serialize Regex (#1307)

marianomarciello:
# Description

 SerializableRegex allows to serialize Regex without using a reference to an option. This will avoid the clippy warning about ref_option.

Fixes #(1307)

## Type of change

# Checklist:

- [x ] Ran `make test` locally with no errors or warnings reported

### DIFF
--- a/leftwm/src/config.rs
+++ b/leftwm/src/config.rs
@@ -92,9 +92,14 @@ pub struct SerializableRegex {
 }
 
 impl SerializableRegex {
-    pub fn new(str: String) -> Self {
+    /// # Panics
+    ///
+    /// Function can only panic if the str used is not a valid regex
+    #[must_use]
+    pub fn new(str: &str) -> Self {
         SerializableRegex {
-            regex: Regex::new(&str).unwrap(),
+            regex: Regex::new(str)
+                .unwrap_or_else(|e| panic!("Can't create regex from {str} error {e}")),
         }
     }
 }
@@ -116,7 +121,7 @@ impl<'de> Deserialize<'de> for SerializableRegex {
             None => Err(<D::Error as serde::de::Error>::custom(
                 "Error during MyRegex deserialization",
             )),
-            Some(re) => Ok(SerializableRegex::new(re)),
+            Some(re) => Ok(SerializableRegex::new(re.as_str())),
         }
     }
 }
--- a/leftwm/src/config.rs
+++ b/leftwm/src/config.rs
@@ -92,14 +92,14 @@ pub struct SerializableRegex {
 }
 
 impl SerializableRegex {
-    /// # Panics
+    /// # Errors
     ///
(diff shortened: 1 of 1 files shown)


leftwm/src/config.rs
```diff
@@ -96,6 +86,46 @@ pub struct WindowHook {
...
 
+#[derive(Debug, Clone)]
+pub struct SerializableRegex {
+    regex: Regex,
+}
+
+impl SerializableRegex {
+    /// # Panics
+    ///
+    /// Function can only panic if the str used is not a valid regex
```
mautamu, line 97: Probably should call out for reference which syntax to use; think it's still https://github.com/google/re2/wiki/Syntax, no?

leftwm/src/config.rs
```diff
@@ -96,6 +86,46 @@ pub struct WindowHook {
...
+    #[must_use]
+    pub fn new(str: &str) -> Self {
+        SerializableRegex {
+            regex: Regex::new(str)
```
mautamu, line 101: Since we can return an Error from deserialize(), should we also do so here and use a chained let in deserialize() to surface the error upwards without a panic?

mautamu (commented):
These changes seem reasonable to me, though yet to test. Left a couple comments.

leftwm/src/config.rs
marianomarciello: I think that maybe instead of using a github page with the definition, it might be easier to link the docs from the create that is used ? https://docs.rs/regex/latest/regex/#syntax

leftwm/src/config.rs
marianomarciello: Added in the below commit. :)

leftwm/src/config.rs
mautamu: Works even better! Great idea.

leftwm/src/config.rs
mautamu: Awesome, thank you very much! :)

marianomarciello:
Hey @VuiMuich  I have added two tests in my latest commit. Please if you have time take a look :). Thanks

VuiMuich:
Thanks @marianomarciello, looking good. 
Just a tiny nitpick and one thing I am unsure about myself:
- Should the first be actually better be two separate tests for serialization and deserialization, so each test is as minimal as possible? To be fair I am no expert on this and I am completely fine to leave this as is, too.
- it occurred to me, right after sending my previous remark, but hadn't time to edit and then forgot about it until now reading the ready test: the second test should be named `attempt_deserialize_...`

marianomarciello:
Hey @VuiMuich  I added two more tests and fixed the typo.
Thanks for spotting that :D 

Let me know if any other change for this PL are needed. Thanks.

VuiMuich:
Good thinking with the two new tests. And thanks for your contribution!

PR #1364, opened 2026-07-12
fix: keep dialogs above  windows

NicTanghe:
Keep dialog-like windows visible and stacked above fullscreen windows without relying on focus. Restack windows when transient hints change.

# Description

Fixes #(issue)

## Type of change

I'm unfamiliar

# Checklist:

### DIFF
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -341,13 +341,8 @@ fn insert_window<H: Handle>(state: &mut State<H>, window: &mut Window<H>, layout
         }
     }
 
-    // If a window is a dialog, splash, utility, floating or scractchpad we want it to be at the top.
-    if window.r#type == WindowType::Dialog
-        || window.r#type == WindowType::Splash
-        || window.r#type == WindowType::Utility
-        || window.floating()
-        || is_scratchpad(state, window)
-    {
+    // Keep dialog-like, floating, and scratchpad windows at the top.
+    if window.r#type.is_dialog_like() || window.floating() || is_scratchpad(state, window) {
         state.windows.insert(0, window.clone());
         return;
     }
--- a/leftwm-core/src/handlers/window_handler.rs
+++ b/leftwm-core/src/handlers/window_handler.rs
@@ -155,6 +155,7 @@ impl<H: Handle, C: Config, SERVER: DisplayServer<H>> Manager<H, C, SERVER> {
         let mut changed = false;
         let mut fullscreen_changed = false;
         let mut above_changed = false;
+        let mut transient_changed = false;
         let strut_changed = change.strut.is_some();
         let windows = self.state.windows.clone();
         if let Some(window) = self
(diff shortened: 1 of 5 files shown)


leftwm-core/src/models/window_type.rs
```diff
@@ -17,3 +17,22 @@ pub enum WindowType {
...
 }
+
+impl WindowType {
+    #[must_use]
+    pub fn is_dialog_like(&self) -> bool {
+        matches!(
+            self,
+            Self::Dialog
+                | Self::Splash
+                | Self::Utility
```
mautamu, line 28: Definitely outside the scope of this PR: do any bars use Utility as a window type? Hopefully they are all using Dock or Toolbar.

leftwm-core/src/handlers/window_handler.rs
```diff
@@ -518,6 +523,89 @@ mod tests {
     use crate::layouts::MONOCLE;
     use crate::models::{MockHandle, Screen};
 
+    fn last_window_order(state: &State<MockHandle>) -> Vec<WindowHandle<MockHandle>> {
```
mautamu, line 521: Is this function just for tests?

leftwm-core/src/handlers/window_handler.rs
mautamu: Do we want to update this list as well?

mautamu:
Thank you, @NicTanghe ! Apologies for the delay in responding. Just the two questions really -- the question around bars is probably best served elsewhere.

leftwm-core/src/models/window_type.rs
NicTanghe: Hello,

**TLDR**: a lot of fulschreen apps spawn tiny utility windows wich should be on top in fact its even more necessary for these then for normal dialogue windows.you can dl it and polyextrude smth. [houdini](https://www.sidefx.com/products/houdini-apprentice/)


This isn`t so much about bars and more about child windows spawned by fullschreen applications.
Utility windows espectially where windows I was having issues with in particular and need this behaviour on the most. I dont know how they relate to bars as i haven't ran xprop on those.

[Houdini](https://www.sidefx.com/products/houdini-apprentice/) is a 3d vfx application and when you edit geometry in the vieuwport its parameters spawn on the vieuwport as seperate windows (wich is neat because picom blurs their bg aswell) but it causes allot of problems with leftwm. you can spawn a utility window by poly extruding something with this patch a tiny white window should appear and grab your mouse.

mautamu: Makes sense; sounds like, even if a bar does use Utility, we should probably handle those as desired behaviour to show up atop everything.

leftwm-core/src/handlers/window_handler.rs
NicTanghe: yes

leftwm-core/src/handlers/window_handler.rs
NicTanghe: seems best ,code is updated.

mautamu:
Looks good, thanks for the answers and fixes!

Issue #880, opened 2022-10-06
Set sane default size when window requests 0x0


hertg:

VuiMuich:
Interesting, so do they let the window do whatever it wants in Floating mode?

VuiMuich:
I think this might be the relevant code in [dwm](https://git.suckless.org/dwm/file/dwm.c.html)
```c

void
   1944 updatesizehints(Client *c)
   1945 {
   1946 	long msize;
   1947 	XSizeHints size;
   1948 
   1949 	if (!XGetWMNormalHints(dpy, c->win, &size, &msize))
   1950 		/* size is uninitialized, ensure that size.flags aren't used */
   1951 		size.flags = PSize;
   1952 	if (size.flags & PBaseSize) {
   1953 		c->basew = size.base_width;
   1954 		c->baseh = size.base_height;
   1955 	} else if (size.flags & PMinSize) {
   1956 		c->basew = size.min_width;
   1957 		c->baseh = size.min_height;
   1958 	} else
   1959 		c->basew = c->baseh = 0;
   1960 	if (size.flags & PResizeInc) {
   1961 		c->incw = size.width_inc;
   1962 		c->inch = size.height_inc;
   1963 	} else
   1964 		c->incw = c->inch = 0;
   1965 	if (size.flags & PMaxSize) {
   1966 		c->maxw = size.max_width;
   1967 		c->maxh = size.max_height;
   1968 	} else
   1969 		c->maxw = c->maxh = 0;
   1970 	if (size.flags & PMinSize) {
   1971 		c->minw = size.min_width;
   1972 		c->minh = size.min_height;
   1973 	} else if (size.flags & PBaseSize) {
   1974 		c->minw = size.base_width;
   1975 		c->minh = size.base_height;
   1976 	} else
   1977 		c->minw = c->minh = 0;
   1978 	if (size.flags & PAspect) {
   1979 		c->mina = (float)size.min_aspect.y / size.min_aspect.x;
   1980 		c->maxa = (float)size.max_aspect.x / size.max_aspect.y;
   1981 	} else
   1982 		c->maxa = c->mina = 0.0;
   1983 	c->isfixed = (c->maxw && c->maxh && c->maxw == c->minw && c->maxh == c->minh);
   1984 	c->hintsvalid = 1;
   1985 }

```

MaKaNu:
[...]

Issue #574, opened 2021-11-11
Floating dialogues smaller than expected


tramhao:
**Describe the bug**
Floating dialogues popped up by programs are smaller than normal.

**To Reproduce**
1. start gimp and open any image.
2. Click on 'colors'->'desaturate'->'desaturate'
3. the floating dialogue is smaller and some options are hidden. It can be enlarged by mod+mouse right drag, but need to do it every time.

**Expected behavior**
The size of floating windows should have a sane default to display everything in the diagloue.

**Screenshots**

**Environment:**
 - OS/Distro:  arch
 - Output of `leftwm-check`
x leftwm-check
:: LeftWM version: 0.2.10
:: LeftWM git hash: 26cc711
:: Loading configuration . . .
    -> Configuration loaded OK
:: Checking keybinds . . .
    -> All keybinds OK
:: Checking environment . . .
    -> Environment OK
:: Checking theme . . .
    -> Theme OK

**Additional context**
wrong dialogue
![pic-window-211111-1301-27](https://user-images.githubusercontent.com/7751505/141240372-51a0b955-ff3d-466c-8089-7c736f362e2f.jpg)

correct dialogue
![pic-window-211111-1301-50](https://user-images.githubusercontent.com/7751505/141240391-4c636d67-54d0-4a3e-9042-170962179545.jpg)

VuiMuich:
Is it possible for you to checkout if #571 fixes it for you?
If not, could you post the `xprop` info, please?

tramhao:
Thanks so much. I think #571 solved the problem I have. Now the dialogue is big enough to show all items inside it. Have a nice day.

hertg:
I still experience this issue in [DBeaver](https://dbeaver.io/) on all popup windows.
I guess because they don't specify a min-size?

**xprop of preferences window**
```
_NET_WM_DESKTOP(CARDINAL) = 7
WM_STATE(WM_STATE):
		window state: Normal
		icon window: <field not available>
_NET_WM_STATE(ATOM) = _NET_WM_STATE_MODAL, _NET_WM_STATE_SKIP_TASKBAR
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: True
		Initial state is Normal State.
		bitmap id # to use for icon: 0x5602a60
		bitmap id # of mask for icon: 0x5602a66
		window id # of group leader: 0x5600001
_GTK_THEME_VARIANT(UTF8_STRING) = "dark"
[...]

PR #1371, opened 2026-08-29
fix: preserve geometry for dialog-like windows

NicTanghe:
Use client-requested position and dimensions instead of unconditionally centering dialog-like windows.

# Description

Fixes #(issue)

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - feat(MSRV): update edition to 2024, MSRV to 1.85.0 (via #1338 by @mautamu and @VuiMuich)
 - feat(config.toml): remove TOML support entirely from leftwm, leftwm-check (via #1339 by @mautamu).
 
+### Added
+
+- Add `disable_mouse_grab` to window rules and suppress creation-time mouse warping for utility windows by default (via #1366).
+
 ### Fixes
 
 - Sloppy focus when switching tags with mouse over margins is now fixed (via #1311 by @fransklaver)
--- a/display-servers/x11rb-display-server/src/event_translate.rs
+++ b/display-servers/x11rb-display-server/src/event_translate.rs
@@ -97,6 +97,18 @@ fn from_focus_in(
     Ok(None)
 }
 
+fn requested_geometry(event: &xproto::ConfigureRequestEvent) -> XyhwChange {
+    let value_mask = u16::from(event.value_mask);
+    XyhwChange {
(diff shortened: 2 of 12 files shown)


VuiMuich:
Similar to your other PR, I gave this one a brief read and nothing stuck out badly to me, but again it's atm difficult for me to properly review and test.

NicTanghe:
OK thanks for having a look !
If someone ends up merging, I advise merging the other one first.

mautamu (changes requested):
I've been daily driving this for a couple of days and I'm not a fan of this behavior. I don't mind if it's added to leftwm, but the position portion should be gated behind a default-off configuration option in my opinion. No issues with the code itself; thanks for revisiting this, I recognize that this has been a tricky area of the code in the past to get right.

NicTanghe:
> I've been daily driving this for a couple of days and I'm not a fan of this behavior. I don't mind if it's added to leftwm, but the position portion should be gated behind a default-off configuration option in my opinion. No issues with the code itself; thanks for revisiting this, I recognize that this has been a tricky area of the code in the past to get right.

Hmm what was bothering you with this ?
Perhaps it doesn`t functionally do what i want it to do ?
in my tests this just makes windows appear where i think the application spawning them thinks they should logically go ? And it has transformed working with houdini from being extremely frustating to the way it should function.

I will gate this behind a toggle default off,

2 remarks though.


1. Since you're not a fan i'm not sure it`s doing what its intended to do on your system. And there might be q mistake in the code.
2. i think child floating windows going where they are supposed to go should be default on.

mautamu:
The latest changes look great; thanks @NicTanghe!

The two reasons I'm not a particularly big fan is because: 1) this came up before and this was the solution and 2) there are scenarios where `leftwm` will paint a window behind `eww` or other bars when positioning them at 0,0 which is frequently the requested position (e.g., for Thunderbird's new meeting window).

Related:
https://github.com/leftwm/leftwm/issues/668
https://github.com/leftwm/leftwm/issues/574
https://github.com/leftwm/leftwm/issues/880

mautamu (approved):

NicTanghe:
> The latest changes look great; thanks @NicTanghe!
> 
> The two reasons I'm not a particularly big fan is because: 1) this came up before and this was the solution and 2) there are scenarios where `leftwm` will paint a window behind `eww` or other bars when positioning them at 0,0 which is frequently the requested position (e.g., for Thunderbird's new meeting window).
> 
> Related: #668 #574 #880

Yes i have noticed that the file browser will pup up behind my eww bar consistantly when it spawns,
Il have a another look for some sort of workaround maybe idc.

mautamu:
Awesome, thank you @NicTanghe!

